PR #16, opened 2015-01-25
Configure PWM ClockFrequency (fix towards V2.7.7 release)

brentpicasso:
turns out that the V2 firmware had no way of changing the PWM clock frequency. reported by customer. Default PWM clock frequency is 20KHz.

Fixed so PWM clock frequency can be updated via Lua Script. Later, clock frequency setting will be configurable via PWM configuration screen in app. 

Some code cleanup in areas that were touched.

Bench tested on MK1 and MK2

### DIFF
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -260,9 +260,9 @@ typedef struct _PWMConfig{
 } PWMConfig;
 
 /// PWM frequency in Hz.
-#define MAX_PWM_CLOCK_FREQUENCY             200000
-#define MIN_PWM_CLOCK_FREQUENCY	 	10
-#define DEFAULT_PWM_CLOCK_FREQUENCY			10000
+#define MAX_PWM_CLOCK_FREQUENCY             65000
+#define MIN_PWM_CLOCK_FREQUENCY	 			10
+#define DEFAULT_PWM_CLOCK_FREQUENCY			20000
 
 /// Maximum duty cycle value.
 #define MAX_PWM_DUTY_CYCLE              	100
@@ -527,7 +527,7 @@ char filterPwmLoggingMode(int config);
 unsigned short filterPwmDutyCycle(int dutyCycle);
 unsigned short filterPwmPeriod(int period);
 int filterImuRawValue(int accelRawValue);
-int filterPwmClockFrequency(int frequency);
+uint16_t filterPwmClockFrequency(uint16_t frequency);
 char filterTimerMode(int config);
 unsigned char filterPulsePerRevolution(unsigned char pulsePerRev);
 unsigned short filterTimerDivider(unsigned short divider);
--- a/src/logger/luaLoggerBinding.c
+++ b/src/logger/luaLoggerBinding.c
@@ -128,7 +128,11 @@ int Lua_GetAtSplit(lua_State *L){
 
 int Lua_SetPWMClockFrequency(lua_State *L){
 	if (lua_gettop(L) >= 1){
-		getWorkingLoggerConfig()->PWMClockFrequency = filterPwmClockFrequency(lua_tointeger(L,1));
+		LoggerConfig *loggerConfig = getWorkingLoggerConfig();
+
+		uint16_t clock_frequency = filterPwmClockFrequency(lua_tointeger(L,1));
+		loggerConfig->PWMClockFrequency = clock_frequency;
+		PWM_set_clock_frequency(clock_frequency);
 	}
 	return 0;
 }
--- a/stm32_base/hal/PWM_stm32/PWM_device_stm32.c
+++ b/stm32_base/hal/PWM_stm32/PWM_device_stm32.c
@@ -155,7 +155,8 @@ unsigned short PWM_device_channel_get_period(unsigned int channel){
 }
 
 void PWM_device_set_duty_cycle(unsigned int channel, unsigned short duty){
-	uint32_t CCR_period = (g_currentClockPeriod * 100) / (1000/duty) / 10;
+	duty = duty > MAX_DUTY_CYCLE ? MAX_DUTY_CYCLE : duty;
+	uint32_t CCR_period = (g_currentClockPeriod * 100) / (1000 / duty) / 10;
 	switch(channel){
 		case 0:
 			TIM4->CCR4 = CCR_period;
--- a/stm32_base/hal/PWM_stm32/PWM_device_stm32.c
+++ b/stm32_base/hal/PWM_stm32/PWM_device_stm32.c
@@ -83,8 +83,8 @@ int PWM_device_init(){
 
 void PWM_device_set_clock_frequency(uint16_t clockFrequency){
 
-	uint32_t period = (1000 * ((CLOCK_FREQUENCY_PERIOD_SCALING * 10000)/clockFrequency)) / 10000;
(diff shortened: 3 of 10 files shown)


include/logger/loggerConfig.h
```diff
@@ -260,9 +260,9 @@ typedef struct _PWMConfig{
 } PWMConfig;
 
 /// PWM frequency in Hz.
-#define MAX_PWM_CLOCK_FREQUENCY             200000
-#define MIN_PWM_CLOCK_FREQUENCY	 	10
-#define DEFAULT_PWM_CLOCK_FREQUENCY			10000
+#define MAX_PWM_CLOCK_FREQUENCY             65000
```
stieg: https://www.kernel.org/doc/Documentation/CodingStyle : Chapter 12

Please make it one space between the end of the definition and the value

```#define MAX_PWM_CLOCK_FREQUENCY 65000```

File: include/logger/loggerConfig.h [hunk 2]
```diff
@@ -527,7 +527,7 @@ char filterPwmLoggingMode(int config);
 unsigned short filterPwmDutyCycle(int dutyCycle);
 unsigned short filterPwmPeriod(int period);
 int filterImuRawValue(int accelRawValue);
-int filterPwmClockFrequency(int frequency);
+uint16_t filterPwmClockFrequency(uint16_t frequency);
```
stieg: :+1:

src/logger/luaLoggerBinding.c
```diff
@@ -128,7 +128,11 @@ int Lua_GetAtSplit(lua_State *L){
 
 int Lua_SetPWMClockFrequency(lua_State *L){
 	if (lua_gettop(L) >= 1){
-		getWorkingLoggerConfig()->PWMClockFrequency = filterPwmClockFrequency(lua_tointeger(L,1));
+		LoggerConfig *loggerConfig = getWorkingLoggerConfig();
+
+		uint16_t clock_frequency = filterPwmClockFrequency(lua_tointeger(L,1));
+		loggerConfig->PWMClockFrequency = clock_frequency;
+		PWM_set_clock_frequency(clock_frequency);
```
stieg: Should the `PWM_set_clock_frequency` method also handle updating the LoggerConfig too?  Seems like it would be better bundled that way unless there are cases where you would want to change one but not the other.

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
```diff
@@ -10,7 +10,11 @@
 #define MAX_DUTY_CYCLE 100
 #define PWM_CHANNEL_COUNT 4
 
-#define DEFAULT_PERIOD 1000
+#define CLOCK_FREQUENCY_PRESCALER_SCALING 42000000
+
+//magic number 20970 = trimmmed value.
+//adjust to calibrate further (if using a more accurate scope)
+#define CLOCK_FREQUENCY_PERIOD_SCALING 20970
```
stieg: :100: That is the correct way per our standard.

include/logger/loggerConfig.h
Jeff-Ciesielski: I noticed a lot of non-kernel style violations as well, but ignored them for the moment.  Is better to go through and address style issues at once rather than ad-hoc to avoid inconsistency IMO.

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
```diff
@@ -37,6 +41,8 @@ static gpio analogCtrlGpios[] = {
 	{ RCC_AHB1Periph_GPIOE, GPIOE, GPIO_Pin_1}
 };
 
+static uint16_t g_currentClockPeriod = 0;
```
stieg: the `= 0` is not needed here as that is the default value of static variables as defined per BSS rules on ELF files.  The = 0 makes the firmware larger for no good reason (albeit a very small size increase).

src/logger/luaLoggerBinding.c
stieg: Hrmm... I'm on the fence about this one.  Your call here.

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
```diff
@@ -69,14 +75,16 @@ static void initAnalogControlGpios(){
...
 
-	uint16_t prescaler = (uint16_t) ((SystemCoreClock) / 210000) - 1;
-	prescaler = 4 - 1;
+	return 1;
+}
+
+void PWM_device_set_clock_frequency(unsigned short clockFrequency){
+
+	uint32_t period = (1000 * ((CLOCK_FREQUENCY_PERIOD_SCALING * 10000)/clockFrequency)) / 10000;
+	uint16_t prescaler = (uint16_t) ((SystemCoreClock) / CLOCK_FREQUENCY_PRESCALER_SCALING) - 1;
```
stieg: Use `const` for code safety.

src/logger/luaLoggerBinding.c
brentpicasso: No, it shouldn't based on the pattern and style of the rest of the HAL layer.

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
brentpicasso: so is that a thumbs up?

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
brentpicasso: for which lines / variables?

include/logger/loggerConfig.h
brentpicasso: this is no different style than what it was before. Although, I am touching other things elsewhere as I go (stdint 's  for example).

stieg: My rule has always been that if you touch it, you are responsible for make it fit to style.  It promotes moving towards the standard.  I like @Jeff-Ciesielski 's approach, but until that is done I think my policy is the way to move forward.

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
stieg: yes.

stm32_base/hal/PWM_stm32/PWM_device_stm32.c
stieg: Period and prescaler.

include/logger/loggerConfig.h
Jeff-Ciesielski: I disagree, if we do that, then we have multiple non-related commits polluting the pull request.  It's a lot of noise.

brentpicasso: I don't see where in chapter 12 of that document it calls for one space between end of definition and value. What am I missing?

stieg: I don't see it mentioned explicitly either... but I remember that being the standard.

@Jeff-Ciesielski,
   I don't follow your logic.  Perhaps let me rephrase my point:

If the person changing the code is _touching_ the lines anyways (as @brentpicasso was doing in this case), then any lines he touches should be brought up to code standard.  I'm not implying doing that for the whole block per the reason you just mentioned.  Just the lines that are touched.

brentpicasso: A couple of things:
 
1) I don't see in the kernel coding guide a recommendation on this topic
2) if there is little risk to bugs, I would like to make this visual formatting an exception to the style guide. I really like seeing the visual formatting.

PR #18, opened 2015-01-27
make max virtual channel count separately configurable for MK1 and MK2

brentpicasso:
make max virtual channels configurable based on hardware version (MK1 / MK2), using capabilities.h.


Jeff-Ciesielski:
:+1:

stieg:
:shipit:

PR #19, opened 2015-01-30
V2.7.8 work in progress

stieg:
Merge of Firmware release 2.7.8 back into master.


src/logging/printk.c
```diff
@@ -70,6 +70,12 @@ int writek_int(int value){
 	return write_to_log_buff(buf);
 }
 
+int writek_float(float value){
+	char buf[20];
```
brentpicasso: is this enough for a floating point number?

stieg: Yes.  Floating point has up to 7 digits, plut sign and period and exp values in rare cases.

src/tracks/tracks.c
```diff
@@ -107,7 +107,8 @@ int isStartPointValid(const Track *t) {
    return isValidPoint(&p);
 }
 
-GeoPoint getSectorGeoPointAtIndex(const Track *t, const int index) {
+GeoPoint getSectorGeoPointAtIndex(const Track *t, int index) {
+   if (index < 0) index = 0;
    const int max = isStage(t) ? STAGE_SECTOR_COUNT : CIRCUIT_SECTOR_COUNT;
```
brentpicasso: this looks like it might've fixed a crashing bug (checking if < 0) . good catch

stieg: actually... no.  But it may have caused the sector stuff to break in a stage scenario.

stieg: If you look where the sector array starts on both stage and circuit, you can see the finish point is always at the -1 position to the sector array.

stieg: So on a circuit track there would be no problem, but on a stage track the sector logic might possibly not work.  Anyway... its good to have.  I'm talking too much and will shut up startiiiiiinnnnngggg now.

PR #20, opened 2015-01-31
(OBD2 PID querying optimizations for MK2; bumping fast telemetry link back to 50Hz) V2.7.9 work in progress

brentpicasso:
MK1's OBD2 PID querying was limited to 10Hz so the OBD2 task wouldn't swamp the shared SPI connection with the IMU and SD card. (I can't believe so many things are hung off of the same bus).

This change allows MK2 to query OBD-II pids at basically full speed (1000Hz max) while still allowing MK1 to be throttled.

I'd closely scrutinize my logic around the OBD-II task to make sure it works correctly when OBD-II is disabled or no channels are configured. If there are no OBD-II channels to query, it should fall into the long 2 second delay at the bottom of the loop.

Also bumped telemetry speed back to 50Hz for bluetooth; we had it at 10Hz b/c of app. App now should be able to handle 50Hz since the UI is updating asynchronously from the rate of the data being fed. (UI updates itself at 10Hz currently regardless of data rate coming from bluetooth)

This was all to allow Harry's lap timer to see full rate for PIDs.


Jeff-Ciesielski:
Change looks good except I think you have some typos: MAX_SAMPE_RATE -> MAX_SAMPLE_RATE

brentpicasso:
what?! you don't know what a SAMPE is? ;)

Jeff-Ciesielski:
Well, it IS engineering related...http://www.sampe.org/

brentpicasso:
ok fixed

Jeff-Ciesielski:
:+1:

PR #21, opened 2015-02-01
update setObd2Config API to prevent API buffer overflow (V2.7.9 work in progress)

brentpicasso:
This change adds a 'starting from index' for setting an array of OBD-II PIDs.  Reason: setting 20 PIDs all at once was overflowing the serial Rx buffer, and I felt it was wasteful to expand the buffer to accommodate the full message.  

This requires the app to set all 20 OBD-II pids in two steps, instead of setting all at once. up to 10 PIDs can be sent at a time. (20 separate single PID calls or 2 calls with 10 PIDs each can be made to set all 20 PIDs)

Added unit tests to check for success and failure modes.


src/logger/loggerApi.c
```diff
@@ -1085,25 +1085,41 @@ static const jsmntok_t * setPidExtendedField(const jsmntok_t *valueTok, const ch
 	return valueTok + 1;
 }
 
+/* Max number of PIDs that can be written in the setOBD2Config message */
+#define MAX_OBD2_MESSAGE_PIDS 10
```
brentpicasso: OK to put the #define close to where it would be used, or is it better form to put it at the top? was conflicted about this.

Jeff-Ciesielski: IMO I'd say better form to put it at the top.

Jeff-Ciesielski:
Aside from my comment about the #define, looks good to me.  Kudos on writing tests for it as well :). :+1:

brentpicasso:
feels real good writing tests, although it still feels like drudgery :0

stieg:
Looks pretty good.  FYI your indentation is off in the Unit tests.

src/logger/loggerApi.c
```diff
@@ -1088,22 +1091,35 @@ static const jsmntok_t * setPidExtendedField(const jsmntok_t *valueTok, const ch
...
+	}
 
-      for (pidsTok++; pidIndex < arrSize; pidIndex++) {
-         PidConfig *pidCfg = obd2Cfg->pids + pidIndex;
-         ChannelConfig *chCfg = &(pidCfg->cfg);
-         pidsTok = setChannelConfig(serial, pidsTok, chCfg, setPidExtendedField, pidCfg);
-      }
-   }
+	const jsmntok_t *pidsTok = findNode(json, "pids");
+	if (pidsTok != NULL && (++pidsTok)->type == JSMN_ARRAY) {
```
stieg: [Feedback]: this is borderline too complex (with the increment inline in the second statement).  Normally I would use consts values and let the compiler handle the optimizations.

PR #22, opened 2015-02-22
Get rcpsim compiling on OSX

paulbutcher:


brentpicasso:
This is great, thanks. What version of GCC did you use to compile it?

PR #23, opened 2015-03-20
V2.7.9 imu calibration fix

brentpicasso:
This fixes a number of things for IMU. This will feed into the 2.7.9 release.  Bench tested on MK1 and MK2.

1) calibration was broken when the channels were re-mapped (for alternate install orientation)

2) restored the natural orientations the IMU outputted (for some reason we tried to change it around a long time ago. I forgot why. Probably trying to be too clever. So, we'll let the IMU output what it thinks is positive and negative G's and Gyros, and say that's industry standard, dammit)

3) Everytime we wrote IMU configuration, it called imu_init, which created a new IMU task. Not good.  Created a imu_soft_init function to bypass this boot-up init.


src/imu/imu.c
```diff
@@ -29,19 +29,6 @@ float imu_read_value(unsigned char imuChannel, ImuConfig *ac){
...
-		case IMU_CHANNEL_YAW:
-		case IMU_CHANNEL_PITCH:
-		case IMU_CHANNEL_ROLL:
-			scaledValue = -scaledValue;
-			break;
-		default:
-			break;
-	}
-
 	//now invert based on configuration
```
brentpicasso: I forget why I thought we needed to do this. I must've been drunk. killing it.

src/logger/loggerApi.c
```diff
@@ -591,7 +591,7 @@ static void getImuConfigs(size_t channelId, void ** baseCfg, ChannelConfig ** ch
 }
 
 int api_setImuConfig(Serial *serial, const jsmntok_t *json){
-	int res = setMultiChannelConfigGeneric(serial, json, getImuConfigs, setImuExtendedField, imu_init);
+	int res = setMultiChannelConfigGeneric(serial, json, getImuConfigs, setImuExtendedField, imu_soft_init);
 	return res;
```
brentpicasso: call soft init so we don't create a new IMU task every time we write new IMU config. >.>

src/imu/imu.c
```diff
@@ -71,17 +58,24 @@ void imu_calibrate_zero(){
 		int zeroValue = g_imu_filter[physicalChannel].current_value;
 		//adjust for gravity
 		float countsPerUnit = imu_device_counts_per_unit(physicalChannel);
-		if (c->physicalChannel == IMU_CHANNEL_Z) zeroValue-= (countsPerUnit * (c->mode != MODE_IMU_INVERTED ? 1 : -1));
+		if (i == IMU_CHANNEL_Z) zeroValue -= (countsPerUnit * (c->mode != MODE_IMU_INVERTED ? 1 : -1));
 		c->zeroValue = zeroValue;
```
brentpicasso: adjust for gravity on the actual mapped channel.

src/imu/imu.c
```diff
@@ -71,17 +58,24 @@ void imu_calibrate_zero(){
 		int zeroValue = g_imu_filter[physicalChannel].current_value;
 		//adjust for gravity
 		float countsPerUnit = imu_device_counts_per_unit(physicalChannel);
-		if (c->physicalChannel == IMU_CHANNEL_Z) zeroValue-= (countsPerUnit * (c->mode != MODE_IMU_INVERTED ? 1 : -1));
+		if (i == IMU_CHANNEL_Z) zeroValue -= (countsPerUnit * (c->mode != MODE_IMU_INVERTED ? 1 : -1));
```
Jeff-Ciesielski: This statement is entirely too tricky.

src/imu/imu.c
```diff
@@ -71,17 +58,24 @@ void imu_calibrate_zero(){
...
 		c->zeroValue = zeroValue;
 	}
 }
 
 int imu_init(LoggerConfig *loggerConfig){
 	imu_device_init();
-	init_filters(loggerConfig);
+    init_filters(loggerConfig);
 	return 1;
```
Jeff-Ciesielski: Broken indentation

src/imu/imu.c
```diff
@@ -71,17 +58,24 @@ void imu_calibrate_zero(){
...
 }
 
+int imu_soft_init(LoggerConfig *loggerConfig){
+    init_filters(loggerConfig);
+    return 1;
+}
```
Jeff-Ciesielski: Any reason we are returning a value here?

src/imu/imu.c
brentpicasso: I think we can put the true part of the if statement it it's own block, per our convention.

But, are ternary operators in general "too tricky"?

src/imu/imu.c
brentpicasso: want to run everything through a code formatter after this is done.

src/imu/imu.c
brentpicasso: It's a standard convention for HAL / device init methods, even if it always succeeds.

src/imu/imu.c
Jeff-Ciesielski: Case dependant methinks.  I really dislike them as a matter of preference (Kernel standard states "Don't put multiple statements on a single line unless you have something to hide"), but as with most things in C, they have their place if used sparingly.

src/imu/imu.c
Jeff-Ciesielski: I see now why you are doing this.  Looks icky out of context, but this function is effectively adhering to an interface.

src/imu/imu.c
```diff
@@ -64,24 +51,35 @@ static void imu_flush_filter(size_t physicalChannel){
...
 		size_t physicalChannel = c->physicalChannel;
 		imu_flush_filter(physicalChannel);
 		int zeroValue = g_imu_filter[physicalChannel].current_value;
-		//adjust for gravity
 		float countsPerUnit = imu_device_counts_per_unit(physicalChannel);
-		if (c->physicalChannel == IMU_CHANNEL_Z) zeroValue-= (countsPerUnit * (c->mode != MODE_IMU_INVERTED ? 1 : -1));
+		if (logicalChannel == IMU_CHANNEL_Z){ //adjust for gravity
+		    if (c->mode == MODE_IMU_INVERTED){
+		        countsPerUnit = -countsPerUnit;
+		    }
```
brentpicasso: Ok, made less tricky.

src/imu/imu.c
brentpicasso: pushed an update to make less tricky.

PR #24, opened 2015-03-20
2.7.9 float overflow fix

brentpicasso:
This was discovered when the android app was sending GPS coordinates with a fractional number with many digits - this was overflowing our atof / atod routines, resulting in corrupted conversion values.

Fixed by guarding against overflows and added unit tests.  Hopefully in the future we can just add standard library conversion routines.


src/util/modp_atonum.c
```diff
@@ -1,5 +1,8 @@
 #include "modp_atonum.h"
 
+#define MAX_DOUBLE_DECIMAL_PLACES 10
```
stieg: This is too low for a double whose mantissa is ~= 16 decimal characters.

stieg:
Read the patch.  Its ok and should work but is not ideal because it doesn't handle super large values prior to the decimal.  I wish we could move over to uLlibC for these methods.  Also is it not possible to save on code by having the atof method use the same code as atod and simply cast the double to a float?

brentpicasso:
Yeah, we could wrap the atod in a call to atof(). Think it's worth doing?

On Fri, Mar 20, 2015 at 3:30 PM, Andrew Stiegmann <notifications@github.com>
wrote:

> Read the patch. Its ok and should work but is not ideal because it doesn't
> handle super large values prior to the decimal. I wish we could move over
> to uLlibC for these methods. Also is it not possible to save on code by
> having the atof method use the same code as atod and simply cast the double
> to a float?
>
> —
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/24#issuecomment-84170136>
> .
>



-- 
Brent Picasso

src/util/modp_atonum.c
```diff
@@ -64,24 +67,25 @@ int modp_atoi(const char *str){
 double modp_atod(const char *str){
 	
 	int negative = 0;
-	int fractionMode = 0;
+	int decimalPlaces = 0;
```
Jeff-Ciesielski: Camel case.

src/util/modp_atonum.c
```diff
@@ -96,24 +100,25 @@ double modp_atod(const char *str){
 float modp_atof(const char *str){
 	
 	int negative = 0;
-	int fractionMode = 0;
+	int decimalPlaces = 0;
 	float floatValue = 0.0;
```
Jeff-Ciesielski: ditto

test/util/atonum_test.cpp
```diff
@@ -18,11 +18,41 @@ AtoNumTest::tearDown()
...
+		double expected = 1.123456;
+		CPPUNIT_ASSERT_EQUAL(expected, result);
+	}
+	{
+		const char *test = "1.12345678912345612345672322";
+		double result = modp_atod(test);
+		double expected = 1.123456789;
+		CPPUNIT_ASSERT_EQUAL(expected, result);
+	}
+}
```
Jeff-Ciesielski: yay tests!

src/util/modp_atonum.c
brentpicasso: I forgot what we agreed on for variable name format. I was assuming the general convention of camel case.

PR #25, opened 2015-03-21
V2.7.9 work in progress

brentpicasso:
This is the mother of all 2.7.9 updates. It has the binary mode GPS MK2, and lap statistics broken out into it's own section. 

Just finished adding in the unix time to DateTime struct mapping.


src/gps/dateTime.c
```diff
@@ -133,3 +133,118 @@ tiny_millis_t getUptime() {
 int getUptimeAsInt() {
    return (int) getUptime();
 }
+
+void getDateTimeFromEpochMillis(DateTime *dateTime, millis_t millis) {
```
stieg: Why `void`?  Why not DateTime?  What is the benefit of making the caller allocate the space?

src/gps/dateTime.c
```diff
@@ -133,3 +133,118 @@ tiny_millis_t getUptime() {
...
+
+   uint64_t SecondsSinceEpoch = millis / 1000;
```
stieg: Camel case: `secondsSinceEpoch`

src/gps/dateTime.c
brentpicasso: more efficient than to pass by value.

src/gps/dateTime.c
brentpicasso: Yup, will fix

src/gps/dateTime.c
```diff
@@ -133,3 +133,118 @@ tiny_millis_t getUptime() {
...
+   annuals = (uint32_t)(sec / 31536000); // 365*24*3600
+   if (annuals > 3)
+   {
+     annuals = 3;
+   }
+   sec -= annuals * 31536000ULL;
+
+   // Calculate the year and find out if it's leap
+   year = 1601 + quadricentennials * 400 + centennials * 100 + quadrennials * 4 + annuals;
+   leap = !(year % 4) && (year % 100 || !(year % 400));
```
stieg: There is a method you can use to calculate this as opposed to duplicating the logic.  See `isLeapYear`

stm32_base/hal/gps_skytraq/gps_device_skytraq.c
```diff
@@ -32,16 +39,21 @@ typedef struct _BaudRateCodes{
...
+#define MSG_ID_QUERY_POSITION_UPDATE_RATE					0x10
+#define MSG_ID_CONFIGURE_POSITION_UPDATE_RATE				0x0E
+#define MSG_ID_POSITION_UPDATE_RATE 						0x86
+#define MSG_ID_CONFIGURE_SERIAL_PORT						0x05
+#define MSG_ID_CONFIGURE_GNSS_NAVIGATION_MODE               0x64
+#define MSG_SUBID_CONFIGURE_GNSS_NAVIGATION_MODE            0x17
+#define MSG_ID_CONFIGURE_NMEA_MESSAGE						0x08
+#define MSG_ID_CONFIGURE_MESSAGE_TYPE						0x09
+#define MSG_ID_CONFIGURE_NAVIGATION_DATA_MESSAGE_INTERVAL	0x11
+#define MSG_ID_NAVIGATION_DATA_MESSAGE						0xA8
```
Jeff-Ciesielski: Crazy alignment

src/gps/dateTime.c
stieg: I don't necessarily agree with the readability tradeoff vs the minor performance gain here.  Either way space needs to be made on the stack and values copied.  GCC does a very good job of optimizing this and considering the speed of the processor even on the MK1 device the savings are a no-op.  Also its easier for someone to mess up programming this method by accidentally passing a NULL pointer to the first argument.  You don't have that issue when you return the struct.  But that said this is a fundamental difference in programming style and it doesn't matter too much.

brentpicasso: On a limited stack environment (freeRTOS thread) we have to be careful what we allocated. By passing it by value, we first allocate it once, then it's allocated again within the function; then it's copied back to the caller.   So we have 2x stack usage as well as copying the values back to the caller.

I don't know if GCC will optimize this out; I'm going to assume it doesn't  unless there's definitive proof.

stm32_base/hal/gps_skytraq/gps_device_skytraq.c
brentpicasso: I proposed running everything through a code reformatter once this is committed.

src/gps/dateTime.c
brentpicasso: Added call to this function. thanks!

src/gps/dateTime.c
stieg: TIL that [Return Value Optimization](http://en.wikipedia.org/wiki/Return_value_optimization) apparently only applies to C++ and not C.  So you are correct in noting that there is a 2x usage in stack space.  I was of the opinion that our optimizer would apply this and thus make the space argument moot.

In general I favor simple expressions over complex ones.  Primarily I do this not for myself, but for the other programmers that need to read and use the code that I am creating.  Having a method say its going to do something and return the result is much easier to grasp than needing to understand that you need to pass your return value by reference into the method so it may be populated.  I also follow this strategy to isolate memory access between functions.  Passing pointers is both wonderful and the work of satan when it breaks.  Following a memory isolationist strategy tends to lead to fewer memory issues because you isolate memory usage at the functional level.  This often buys you better reliability at the expense of more memory usage.  This in my book is a worthwhile tradeoff in our case.

  Sure the struct we are passing in this particular example is large when compared to other structures in our code base, but in the overall scheme of things it is still quite small.  At what point do we start getting this anal retentive with stack space savings?  Do we not pass a unit64 type back on the stack because it would cause us to use 16 bytes instead of 8?  What about a structure that is 16 bytes in size?  Where do you draw the tradeoff between code that is more simple and less prone to screw ups vs speed gains and stack space savings?

I don't think your arguments are wrong.  In fact they are spot on accurate in terms of space usage.  We absolutely need to be careful about how much stack space we use.  But I feel that what you are doing is a premature optimization when one is not really needed.  We have more than enough space to spare in our stack sizes and we are very good at keeping our structures small and to the point.  I would trade a bit of memory waste for cleaner more isolated code any day.  All within reason of course.

Jeff-Ciesielski: I have to agree with Brent on this one.  Passing in struct pointers to a function for use as a return is an INCREDIBLY common C programming paradigm.  It's everywhere.  This isn't complicated or tricky, and arguably if other people looking at this code have no idea that this is a thing, they probably aren't the people you want contributing to your codebase.  

To your memory isolationist comment: Upping stack usage for us has the exact opposite effect of reliability.  Dealing with stack overflow problems is incredibly frustrating on embedded platforms, and many of them will come up and bite you in the ass without ever knowing it.  This is in no way 'more reliable'.

When do we start getting anal retentive with stack space?  Always.  I think you're overstating the cost/benefit here.

This hardly feels like premature optimization, it is a coding style preference.  When I saw your original comment, my first thought was "Who the fuck passes a structure by value? That's terrible for stack useage".  What if this function could fail?  How would we return an error condition?  We'd be in the same boat as we are now, but what would your argument be?

stieg: > This isn't complicated or tricky, and arguably if other people looking at this code have no idea that this is a thing, they probably aren't the people you want contributing to your codebase.

Such a comment on par with Linus.  Love it.

> This hardly feels like premature optimization, it is a coding style preference.

Fair point.

> What if this function could fail? How would we return an error condition?

This method doesn't do that so its not a concern of this particular discussion.  It would be a different case should that functionality be needed.  That in itself is a reason to follow the design in question.  But that is not the case here.


Ok.  This horse is dead enough.  This is obviously a difference in design mentalities that will not be resolved in a code review (if it need be).  Lets get something from all this discussion.  Do we want to pass all structures by reference to keep stack space small as a general practice?  This will at least keep things consistent going forward.

brentpicasso: Thanks all for the feedback. I feel less dumb after this conversation :) By and large we pass things by reference in the codebase, except for maybe a few areas where we send tiny structs around (GPSPoint, maybe).  I think for the time being we can stay with this convention.

Going to close this PR and move on. Much thanks and respect for your opinions and wisdom.

PR #26, opened 2015-03-28
V2.8.0 trackmap

brentpicasso:
Store a hard track map ID so it can be positively correlated with the app, so we can disambiguate between different venues.

Also refactored the logic for versionInfo - trackmaps store version info similar to loggerConfig instead of using the different "magic number" approach.

Still to-do - figure out the new max number of trackmaps we can store for MK1 and MK2


src/gps/dateTime.c
```diff
@@ -189,8 +189,6 @@ void getDateTimeFromEpochMillis(DateTime *dateTime, millis_t millis) {
    // (3*100*(365+24/100) + 17*4*(365+1/4) + 1*365)*24*3600 seconds
    sec = secondsSinceEpoch + 11644473600;
 
-   wday = (uint32_t)((sec / 86400 + 1) % 7); // day of week
-
```
brentpicasso: unrelated change, to deal with compiler warning of unused variable.

src/auto_config/auto_track.c
```diff
@@ -56,7 +56,7 @@ const Track* auto_configure_track(Track *defaultCfg, const GeoPoint *gp) {
     const Tracks *tracks = get_tracks();
     if (!tracks || tracks->count <= 0) {
         // Well shit!
-        pr_info("error!\r\n");
+        pr_info("No tracks\r\n");
         return defaultCfg;
     }
```
Jeff-Ciesielski: Was the discrepancy in CR/LF intentional here?

src/tracks/tracks.c
```diff
@@ -16,9 +16,9 @@ static const Tracks g_defaultTracks = DEFAULT_TRACKS;
...
 void initialize_tracks(){
-	if (g_tracks.magicInit != MAGIC_NUMBER_TRACKS_INIT){
-		flash_default_tracks();
-	}
+    if (versionChanged(&g_tracks.versionInfo)){
+        flash_default_tracks();
+    }
 }
 
 int flash_default_tracks(void){
```
Jeff-Ciesielski: Whitespace clobbering.

test/loggerApi_test.cpp
```diff
@@ -871,6 +871,7 @@ void LoggerApiTest::testSetTrackCfgCircuit(){
 	assertGenericResponse(txBuffer, "setTrackCfg", API_SUCCESS);
 
 	CPPUNIT_ASSERT_EQUAL(0, (int)cfg->track.track_type);
+    CPPUNIT_ASSERT_EQUAL(6674, cfg->track.trackId);
```
Jeff-Ciesielski: Screwy indentation

Jeff-Ciesielski:
Minor nits, but looks good to me :+1:

src/tracks/tracks.c
brentpicasso: Eh. I have 4 spaces for tabs right now. We really need to do a once-sweep reformatting through the code. Are there tools for this?

src/auto_config/auto_track.c
brentpicasso: IDK, just trying to get the message to be more helpful

PR #29, opened 2015-04-04
Stage Improvements

stieg:
Adds much improved logic for stage racing including:

* Multiple stage run support without need to restart race capture
* Predictive timing support for stage racing (same restrictions apply).

Also improves our unit tests a bit to help with debugging the sector logic bits.

### DIFF
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -161,6 +161,7 @@ static void lapFinishedEvent(const GpsSnapshot *gpsSnapshot) {
         pr_debug_int(g_lapCount);
         pr_debug("\r\n");
 
+        ++g_lapCount;
         endLapTiming(gpsSnapshot);
         finishLap(gpsSnapshot);
 
@@ -180,12 +181,6 @@ static void _lapStartedEvent(const tiny_millis_t time,
                              const GeoPoint *sp,
                              const float distance,
                              const GpsSnapshot *gpsSnapshot) {
-        // Increment and log that we started a lap.
-        ++g_lapCount;
-        pr_debug("Started Lap ");
-        pr_debug_int(g_lapCount);
-        pr_debug(" with launch control.\r\n");
-
         // Timing and predictive timing
         startLapTiming(time);
         startLap(sp, time);
@@ -214,6 +209,10 @@ static void lapStartedNormalEvent(const GpsSnapshot *gpsSnapshot) {
         const tiny_millis_t time = gpsSnapshot->deltaFirstFix;
         const GeoPoint gp = gpsSnapshot->sample.point;
 
+        pr_debug("Starting lap ");
+        pr_debug_int(g_lapCount);
+        pr_debug("\r\n");
+
         _lapStartedEvent(time, &gp, 0, gpsSnapshot);
 }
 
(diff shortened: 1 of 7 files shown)


brentpicasso:
Looking at it now, looks good so far. Initial question: a customer complained that laps would repeatedly trigger when they were sitting exactly in the start finish line for several minutes. Does this update possibly account for that?

brentpicasso:
Was thinking maybe, due to the cooldown flag?

src/lap_stats/lap_stats.c
```diff
@@ -247,54 +357,39 @@ static void onLocationUpdated(const GpsSnapshot *gpsSnapshot) {
...
    static int startFinishEnabled = 0;
 
-   g_distance += calcDistancesSinceLastSample(gpsSnapshot);
-
-   LoggerConfig *config = getWorkingLoggerConfig();
+   const LoggerConfig *config = getWorkingLoggerConfig();
    const GeoPoint *gp = &gpsSnapshot->sample.point;
    const float targetRadius = degreesToMeters(config->TrackConfigs.radius);
 
+   // FIXME.  This active track configuration here is le no good.
```
brentpicasso: elaborate?

brentpicasso:
Related, if it's in stage mode - can the car sit in the start zone indefinitely without false triggering a lap or lap start?

stieg:
> Initial question: a customer complained that laps would repeatedly trigger when they were sitting exactly in the start finish line for several minutes. Does this update possibly account for that?

Accounted for in stage mode as we always use Launch Control there now.  This should address customer bug.

src/lap_stats/lap_stats.c
stieg: Plainly put the code should not be here.  This code should be invoked right after a configuration has been changed on the track settings.

stieg:
@brentpicasso can you give me more information about the customer issue.  Tell me about their track configuration.  What were their expectations?  I can answer the scenario questions for what will happen going forward (or what should happen anyways).

brentpicasso:
Excellent. So this would apply in circuit mode as well?

Brent Picasso
Autosport Labs

brentpicasso:
The scenario was: Circuit mode, and the car happened to be just sitting right in the start/finish circle for a long time (several mins).  According to report, it periodically clicked off laps (timeout?) with zero lap times.

stieg:
No it does not apply to circuit mode.  I made that choice intentionally because using launch control for every start would cause hiccups in reporting (I can explain this better in person).  The logic says to skip resetting and using launch control when we are in circuit mode.

To address the aforementioned customer issue is something entirely different.  That behavior is RCP working the way its was programmed to work because the assumption was that the aforementioned scenario would never happen.  To fix that we would need a Schmitt trigger like software trigger on Start/Finish detection so that they would have to have to be far enough away from start finish for X time before it would be allowed to trigger again.

brentpicasso:
You done some fine work here. We'll do a final test this evening and then merge this into master.

stieg:
Thanks @brentpicasso.  LMK if you find any further issues.  Filed https://trello.com/c/PV6wd0db/238-implement-schmitt-style-trigger-for-gps-start-finish-zone-to-prevent-inaccurate-triggering-of-start-finish-if-driver-gets-stuck- to track the trigger

PR #30, opened 2015-04-10
GeoTrigger support to replace the cool off timer

stieg:
Since we have had actual customer complaints about the cool off period being insufficient on a real race scenario, we needed a better solution.  This patch introduces the geoTrigger, a simple device that  triggers once you leave its geoCircle.  Using these objects we can replace the concept of the timer and have better results.

This patch adds in the geoTrigger use and removes the timer.  All tests pass.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,9 @@
 ===2.8.0===
-* Refactor stage logic to support multiple stage runs without restarting Race Capture
-* Add support for predictive timing in stage runs (same restrictions & rules apply as in circuit racing).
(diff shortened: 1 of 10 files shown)


brentpicasso:
This looks pretty good. Has it been road tested for circuit and stage use cases? If so, please merge.

stieg:
No road testing.  Tested with unit tests only thus far.  When there were configuration problems the unit tests failed so I am very confident the change will work.

brentpicasso:
Hokay.

PR #31, opened 2015-04-13
P/stieg/elapsed and bug fixes

stieg:
Adds support for the elapsed time field along with new unit tests to make our lives easier.

Also found a minor bug in our distance calculation and fixed it.


brentpicasso:
update new functions to underscore names, per conversations.

PR #35, opened 2015-04-18
P/stieg/osx support

stieg:
Adds the ability for us to compile the unit test from OSX.  Achieves this by moving our unit tests to a 64bit platform and by adding support to work with the clang compiler on OSX.

Next I want to move our unit tests over to clang and begin to use their static analyzer on our code so that we are able to find bugs in the firmware.  Plus moving to clang will make it easier not to break the mac environment.

### DIFF
--- a/src/logger/loggerApi.c
+++ b/src/logger/loggerApi.c
@@ -643,7 +643,7 @@ int api_getAnalogConfig(Serial *serial, const jsmntok_t * json){
 			startIndex = endIndex = modp_atoi(json->data);
 		}
 	}
-	if (startIndex >= 0 && startIndex <= CONFIG_ADC_CHANNELS){
+	if (startIndex <= CONFIG_ADC_CHANNELS){
 		sendAnalogConfig(serial, startIndex, endIndex);
 		return API_SUCCESS_NO_RETURN;
 	}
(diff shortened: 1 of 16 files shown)


src/logger/loggerApi.c
```diff
@@ -643,7 +643,7 @@ int api_getAnalogConfig(Serial *serial, const jsmntok_t * json){
 			startIndex = endIndex = modp_atoi(json->data);
 		}
 	}
-	if (startIndex >= 0 && startIndex <= CONFIG_ADC_CHANNELS){
+	if (startIndex <= CONFIG_ADC_CHANNELS){
 		sendAnalogConfig(serial, startIndex, endIndex);
```
brentpicasso: This is unrelated optimization to the OSX thing, right?

brentpicasso:
some of this looks like general clean up and optimizations, which is good by itself. how much of it was to get OSX support working?

src/logger/loggerApi.c
stieg: Its an unsigned variable where we are checking if its value is less than 0.  So more of cleanup.  CLANG was complaining about it.  Its not needed.

stieg:
All of it was needed to get OSX working with our unit tests.

src/logger/loggerApi.c
brentpicasso: I saw that, looks like a sane change :+1:

Issue #39, opened 2015-04-29
Lua script issues


Pilt2013:
addChannel function: Using over 11 characters for channel name causes unwanted results:

From lua script: ```ABS_Bpressure_Id = addChannel("ABS_Br_Press", 10, 0, -42.5, 425, 'bar')```

Results in logfile output of  "ABS_Br_Pressbar"|"bar"|0|0|10

Also Racecapture v2.7.9 requires a power cycle after uploading a new lua script, otherwise still runs old script.

stieg:
Thanks for the info @Pilt2013.  Yes the firmware currently only supports up to 11 characters (12 with the null termination) for a channel name.  Looks like we need to better enforce that on the method.  However the best way I can think of is simply drop all characters after the 11th.  Perhaps consider using "Brake_Press" or "ABS_Press" as a work around.

stieg:
Ok.  Feedback given.  Sorry for spreading it out between patches and the pull  req.

brentpicasso:
Fixed in https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/40   will show up in 2.8.0

PR #40, opened 2015-04-29
V2.8.0 lua fixes

brentpicasso:
from https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/39

* Fixes for virtual channel label overflow (off by 1 error)
* Auto re-load lua script when lua script flashes

still needs testing on hardware.

### DIFF
--- a/include/memory/memory.h
+++ b/include/memory/memory.h
@@ -6,9 +6,6 @@ typedef enum {
     MEMORY_FLASH_WRITE_ERROR = -1
 } memory_flash_result_t;
 
-#define FLASH_SUCCESS		0
-#define FLASH_WRITE_ERROR	-1
-
 memory_flash_result_t memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
 
 #endif /* MEMORY_H_ */
--- a/include/memory/memory.h
+++ b/include/memory/memory.h
@@ -1,6 +1,9 @@
 #ifndef MEMORY_H_
 #define MEMORY_H_
 
+#define FLASH_SUCCESS		0
+#define FLASH_WRITE_ERROR	-1
+
 int memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
 
 #endif /* MEMORY_H_ */
--- a/include/memory/memory.h
+++ b/include/memory/memory.h
@@ -1,9 +1,14 @@
 #ifndef MEMORY_H_
 #define MEMORY_H_
 
+typedef enum {
+    MEMORY_FLASH_SUCCESS = 0,
+    MEMORY_FLASH_WRITE_ERROR = -1
+} memory_flash_result_t;
+
 #define FLASH_SUCCESS		0
 #define FLASH_WRITE_ERROR	-1
 
-int memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
+memory_flash_result_t memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
 
 #endif /* MEMORY_H_ */
--- a/include/memory/memory.h
+++ b/include/memory/memory.h
@@ -1,11 +1,11 @@
 #ifndef MEMORY_H_
 #define MEMORY_H_
 
-typedef enum {
+enum memory_flash_result_t{
     MEMORY_FLASH_SUCCESS = 0,
     MEMORY_FLASH_WRITE_ERROR = -1
-} memory_flash_result_t;
+};
 
-memory_flash_result_t memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
+enum memory_flash_result_t memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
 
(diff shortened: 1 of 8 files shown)


src/logger/luaLoggerBinding.c
```diff
@@ -550,12 +553,12 @@ int Lua_AddVirtualChannel(lua_State *L){
 	size_t args = lua_gettop(L);
 	if (args >= 2){
 		ChannelConfig chCfg;
-		strncpy(chCfg.label, lua_tostring(L, 1), DEFAULT_LABEL_LENGTH);
+		strncpy(chCfg.label, lua_tostring(L, 1), DEFAULT_LABEL_LENGTH - 1);
```
stieg: No.  This will only copy up to 11 characters and possibly not the null terminated character.  Need to use strnlcpy.  See my patch: https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/41

stieg: For Reference: http://man7.org/linux/man-pages/man3/strcpy.3.html

brentpicasso: :+1:

include/memory/memory.h
```diff
@@ -1,6 +1,14 @@
 #ifndef MEMORY_H_
 #define MEMORY_H_
 
-int memory_flash_region(const void *vAddress, const void *vData, unsigned int length);
+typedef enum {
```
stieg: This is not inline with the linux standard.  Specifically see Chapter 5

```
		Chapter 5: Typedefs

Please don't use things like "vps_t".
It's a _mistake_ to use typedef for structures and pointers. When you see a

	vps_t a;

in the source, what does it mean?
In contrast, if it says

	struct virtual_container *a;

you can actually tell what "a" is.

Lots of people think that typedefs "help readability". Not so. They are
useful only for:

 (a) totally opaque objects (where the typedef is actively used to _hide_
     what the object is).

     Example: "pte_t" etc. opaque objects that you can only access using
     the proper accessor functions.

     NOTE! Opaqueness and "accessor functions" are not good in themselves.
     The reason we have them for things like pte_t etc. is that there
     really is absolutely _zero_ portably accessible information there.

 (b) Clear integer types, where the abstraction _helps_ avoid confusion
     whether it is "int" or "long".

     u8/u16/u32 are perfectly fine typedefs, although they fit into
     category (d) better than here.

     NOTE! Again - there needs to be a _reason_ for this. If something is
     "unsigned long", then there's no reason to do

	typedef unsigned long myflags_t;

     but if there is a clear reason for why it under certain circumstances
     might be an "unsigned int" and under other configurations might be
     "unsigned long", then by all means go ahead and use a typedef.

 (c) when you use sparse to literally create a _new_ type for
     type-checking.

 (d) New types which are identical to standard C99 types, in certain
     exceptional circumstances.

     Although it would only take a short amount of time for the eyes and
     brain to become accustomed to the standard types like 'uint32_t',
     some people object to their use anyway.

     Therefore, the Linux-specific 'u8/u16/u32/u64' types and their
     signed equivalents which are identical to standard types are
     permitted -- although they are not mandatory in new code of your
     own.

     When editing existing code which already uses one or the other set
     of types, you should conform to the existing choices in that code.

 (e) Types safe for use in userspace.

     In certain structures which are visible to userspace, we cannot
     require C99 types and cannot use the 'u32' form above. Thus, we
     use __u32 and similar types in all structures which are shared
     with userspace.

Maybe there are other cases too, but the rule should basically be to NEVER
EVER use a typedef unless you can clearly match one of those rules.

In general, a pointer, or a struct that has elements that can reasonably
be directly accessed should _never_ be a typedef.
```

Instead this should be 
```c
enum memory_flash_result {
        MEMORY_FLASH_SUCCESS = 0,
        MEMORY_FLASH_WRITE_ERROR = -1
};
```

I would just use a boolean here instead and make it easy.

src/logger/loggerApi.c
```diff
@@ -1470,7 +1471,7 @@ int api_getScript(Serial *serial, const jsmntok_t *json){
 int api_setScript(Serial *serial, const jsmntok_t *json){
 
 	int returnStatus = API_ERROR_UNSPECIFIED;
-
+	bool should_reload_script = false;
```
stieg: [Pedantic] Could shorten this variable name.  See Chapter 4 in https://www.kernel.org/doc/Documentation/CodingStyle

include/memory/memory.h
brentpicasso: I did the analysis here and there are many areas in the code that use 0==success semantics, so converting to bool is not trivial at this stage of the game for 2.8.0 release. recommend shelving the bool conversion to later.

stieg: Does this mean we are keeping the typedef?  Or can you fix that?  If fixable then just go that way and be done with it.

brentpicasso:
switched enum to kernel style.

PR #41, opened 2015-04-29
Replace strncpy with strlcpy in our Lua Bindings

stieg:
User reported error that our Lua_Bindings will copy and not
null terminate a string.  This was because we were using strncpy,
which is unsafe around user input.

https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/39


brentpicasso:
is good. thanks.

PR #42, opened 2015-04-30
V2.8.0 heartbeat

brentpicasso:
Heartbeat capabilities. RCP will disconnect after 10 seconds if it doesn't receive a valid message from the server.

Added heartbeat API call, unit tests. Re-ordered some more frequently used APIs towards top of list.


include/logger/loggerApi.h
```diff
@@ -25,8 +25,12 @@
 
 #define LOGGER_API \
```
stieg: A separate but related thought comes to mind... if this define is in the header and the header is included in more than one place, does the code size grow by the number of times this is defined?

src/devices/sim900.c
```diff
@@ -31,15 +35,24 @@ static int writeAuthJSON(Serial *serial, const char *deviceId){
 		serial->put_s(" ");
 		delayMs(250);
 	}
-	serial->put_s("{\"cmd\":{\"schemaVer\":2,\"auth\":{\"deviceId\":\"");
```
stieg: So no more SchemaVer.  Just API ver now?

src/logger/connectivityTask.c
```diff
@@ -273,6 +270,13 @@ void connectivityTask(void *params) {
 				}
 				rxCount = 0;
 			}
+
+			//disconnect if we haven't heard from the other side for a while
+			size_t timeout = getUptimeAsInt() - last_message_time;
```
stieg: [Pedantic] `const size_t timeout ...`

src/logger/loggerApi.c
```diff
@@ -307,6 +307,13 @@ int api_sampleData(Serial *serial, const jsmntok_t *json){
 	return API_SUCCESS_NO_RETURN;
 }
 
+int api_heartBeat(Serial *serial, const jsmntok_t *json){
```
stieg: Formatting is not Linux style.  Fix throughout.

test/loggerApi_test.cpp
```diff
@@ -706,6 +707,19 @@ void LoggerApiTest::testSampleData2() {
                         getSampleResponse(requestJson2));
 }
 
+void LoggerApiTest::testHeartBeat(){
+    resetTicks();
+    incrementTick();
```
stieg: Works but seems sketchy and prone to breaking.  Any way you can expose the ability to set the uptime value as part of the tests.  Then it makes the test more sane.

src/devices/sim900.c
brentpicasso: correct, according to requirements doc.

version.mk
```diff
@@ -1,4 +1,5 @@
 MAJOR=2
 MINOR=8
 BUGFIX=0
+API=3
```
stieg: API version is 3?  Wasn't it 1?  Or does that not matter?

include/logger/loggerApi.h
brentpicasso: AFAIK #defines are symbols for the compiler to use, they don't go into the data section.

src/logger/loggerApi.c
brentpicasso: this is following the style in the rest of the file; I'd prefer to sweep through and change everything at once in a separate style commit.

version.mk
brentpicasso: @ryandoherty suggested since it replaced the old schemaver of 2, it would start with 3.

src/logger/loggerApi.c
stieg: Dude... the longer we put off making our style consistent the harder it will be and the more likely it will be that it won't happen.  Changes don't happen overnight.  We all gotta start doing it the right way and it will encourage us to complete the migration.  Please reconsider.

include/logger/loggerApi.h
stieg: You are correct.

version.mk
stieg: Ok.

src/logger/loggerApi.c
brentpicasso: Ok to do a refactoring across the whole file right after this one? deal?

stieg: Deal.

brentpicasso: too late! I fulfilled on your original request anyway. :)

stieg: Fine.. do what I ask.  Yeesh :P

brentpicasso: :)

test/loggerApi_test.cpp
brentpicasso: fixed. thought about this too; but was lazy. thanks for keeping me honest ;)

brentpicasso:
I think this is well reviewed. merging.

PR #44, opened 2015-05-04
make connectivity task have programmable timeouts based on type of telem...

brentpicasso:
...etry task

Wireless type telemetry tasks don't have a timeout... until we sunset the android specific app

### DIFF
--- a/src/logger/connectivityTask.c
+++ b/src/logger/connectivityTask.c
@@ -38,7 +38,7 @@
 #define IDLE_TIMEOUT							configTICK_RATE_HZ / 10
 #define INIT_DELAY	 							600
 #define BUFFER_SIZE 							1025
-#define DISCONNECT_TIMEOUT                      10000
+#define TELEMETRY_DISCONNECT_TIMEOUT            10000
 
 #define TELEMETRY_STACK_SIZE  					1000
 #define SAMPLE_RECORD_QUEUE_SIZE				10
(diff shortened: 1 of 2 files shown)


src/logger/connectivityTask.c
```diff
@@ -38,7 +38,7 @@
 #define IDLE_TIMEOUT							configTICK_RATE_HZ / 10
 #define INIT_DELAY	 							600
 #define BUFFER_SIZE 							1025
-#define DISCONNECT_TIMEOUT                      10000
+#define TELEMETRY_DISCONNECT_TIMEOUT            10000
```
stieg: Milliseconds I presume?

src/logger/connectivityTask.c
```diff
@@ -107,6 +107,7 @@ static void createCombinedTelemetryTask(int16_t priority, xQueueHandle sampleQue
 		params->init_connection = &null_device_init_connection;
 		params->serial = SERIAL_TELEMETRY;
 		params->sampleQueue = sampleQueue;
+		params->connection_timeout = 0;
```
stieg: Why 0?  What is the significance of 0?

stieg: NVM... read value below.  Nice usage of 0 here.

stieg:
Use of a uint prevents the negative timeout case.  Why do we want to be able to disable it (just curious)?

stieg:
Otherwise looks pretty good to me.  I would say ship it pending answers to those questions.

src/logger/connectivityTask.c
brentpicasso: correctamundo

brentpicasso:
The legacy (current production) app does not send any kind of heartbeat, so the bluetooth connectivity task will time out every 10 seconds to 're establish connection'.  This is to prevent this from happening. In the future, we may re-enable timeout.

PR #45, opened 2015-05-05
V2.8.0 lap detection retry

brentpicasso:
* keep searching for an auto-detected track if none are nearby
* re-detect tracks if new track configuration is written
* refactoring of location update and track configuration routines into separate functions
* some linux code style updates

What's not done yet:

Right now it keeps trying to search on every GPS sample. If there are many many tracks configured for auto-detection, it may cause a CPU hit.  We could have it only check every so often. This would require managing a timestamp and so on. Is it worth doing?

### DIFF
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -27,6 +27,9 @@
 #define TIME_NULL -1
 
 static const Track * g_activeTrack;
+track_status_t g_track_status = TRACK_STATUS_WAITING_TO_CONFIG;
+static int g_sector_enabled = 0;
+static int g_start_finish_enabled = 0;
 
 static int g_configured;
 static int g_atStartFinish;
@@ -48,11 +51,11 @@ static int g_lapCount;
 static float g_distance;
 
 // Our GeoTriggers so we don't unintentionally trigger on start/finish
-static struct GeoTrigger startGeoTrigger;
-static struct GeoTrigger finishGeoTrigger;
+static struct GeoTrigger g_start_geo_trigger;
+static struct GeoTrigger g_finish_geo_trigger;
 
-TESTABLE_STATIC void set_active_track(const Track* defaultTrack) {
-	g_activeTrack = defaultTrack;
+TESTABLE_STATIC void set_active_track(const Track* track) {
+	g_activeTrack = track;
 }
 
 static float degrees_to_meters(float degrees) {
@@ -61,15 +64,7 @@ static float degrees_to_meters(float degrees) {
 }
 
 track_status_t lapstats_get_track_status( void ) {
-	if (g_activeTrack == NULL){
-		return TRACK_STATUS_WAITING_TO_CONFIG;
-	}
-	else if (g_activeTrack->trackId > 0) {
-		return TRACK_STATUS_AUTO_DETECTED;
-	}
-	else {
-		return TRACK_STATUS_FIXED_CONFIG;
-	}
+    return g_track_status;
 }
 
 int32_t lapstats_get_selected_track_id( void ) {
@@ -210,7 +205,7 @@ TESTABLE_STATIC void lap_finished_event(const GpsSnapshot *gpsSnapshot) {
          * finish before we re-arm the system.  This is mainly used
          * in stage situations.
          */
-        resetGeoTrigger(&startGeoTrigger);
+        resetGeoTrigger(&g_start_geo_trigger);
 }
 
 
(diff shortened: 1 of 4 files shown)


brentpicasso:
This PR should be called "v2.8.0 TRACK detection retry"...

src/lap_stats/lap_stats.c
```diff
@@ -391,67 +387,75 @@ static void setupGeoTriggers(const TrackConfig *tc, const Track *track) {
...
+static void lapstats_setup(const GpsSnapshot *gps_snapshot){
+    const LoggerConfig *config = getWorkingLoggerConfig();
+    const float target_radius = degrees_to_meters(config->TrackConfigs.radius);
+    const GeoPoint *gp = &gps_snapshot->sample.point;
+
+    const Track *track = NULL;
+    const TrackConfig *trackConfig = &(config->TrackConfigs);
+    if (trackConfig->auto_detect){
+        track = auto_configure_track(NULL, gp);
+    }
```
stieg: Fix formatting

src/lap_stats/lap_stats.c
```diff
@@ -391,67 +387,75 @@ static void setupGeoTriggers(const TrackConfig *tc, const Track *track) {
...
+    else{
+        track = &trackConfig->track;
+        g_track_status = TRACK_STATUS_FIXED_CONFIG;
+        pr_info("track: using fixed config");
+    }
+    set_active_track(track);
```
stieg: So its ok to potentially set a null active track?

src/lap_stats/lap_stats.c
```diff
@@ -391,67 +387,75 @@ static void setupGeoTriggers(const TrackConfig *tc, const Track *track) {
...
+    if (track){
```
stieg: Unnecessary indentation.  Instead do `if (!track) return;`.

stieg: So then what is your case for when user sets auto track detection but has no tracks that match.  Seems like things will never work no?

src/logger/loggerApi.c
```diff
@@ -1435,8 +1435,16 @@ int api_addTrackDb(Serial *serial, const jsmntok_t *json){
 		Track track;
 		const jsmntok_t *trackNode = findNode(json, "track");
 		if (trackNode != NULL) setTrack(trackNode + 1, &track);
-		add_track(&track, index, mode);
-		return API_SUCCESS;
+		int result = add_track(&track, index, mode);
+		if (result == TRACK_ADD_RESULT_OK){
```
stieg: Formatting looks broken here.  Fix please.

stieg: Also I'm not sure what is supposed to be happening here.  Looks very weird to me.

brentpicasso: What happens here is if we are successfully finished writing tracks then we tell lapstats that the track configuration changed.  In lapstats this triggers a search of the nearest configured track.

brentpicasso: which formatting? the curly brace after the parens?

src/lap_stats/lap_stats.c
brentpicasso: If the config is set for auto detection and there are no tracks nearby, there will be no lap stats, by design. it will _not_ fall through to fixed configuration. It will continue searching for a track until one is found.

brentpicasso: Gosh, I'm really conflicted sometimes with multiple exit points. I get having guard statements at the top, but when returns are sprinkled through the body I think it makes the code harder to maintain. This article illustrates it well. Thoughts?

http://www.tomdalling.com/blog/coding-tips/coding-tip-have-a-single-exit-point/

src/lap_stats/lap_stats.c
brentpicasso: lapstats_processUpdate() controls whether the active track is ever used based on g_configured.  All of the functions that access this variable directly is protected with static... so i think it's fairly safe.

src/lap_stats/lap_stats.c
stieg: IMHO if it gets too confusing then its because your method is doing too much and you need to break it up into something smaller.  Multiple exits are a wonderful feature and make the flow of code much easier to read and save on tons of indentation and potentially complex if/if else/else statements.

src/logger/loggerApi.c
stieg: Looks like spaces were used when spaces where tabs were supposed to be there.  If/else is on different line

src/lap_stats/lap_stats.c
brentpicasso: Sounds like a style preference mostly, then.

stieg: Not just style, but maintainability and ease of understanding and contributing.  Much of this is seein the code I did for LapStats when I broke out the logic for separate start/finish and geoTriggers.  The code is almost read like literature with short functions and arguably correce method names to go with them.

brentpicasso:
merging and addressing style issue in 2.8.1

PR #46, opened 2015-05-12
V2.8.1 style linux formatting

brentpicasso:
because we're so stylish


SAM7s_base/ADC_at91/ADC_device_at91.c
```diff
@@ -28,119 +28,124 @@
...
-        *a5 = AT91F_ADC_GetConvertedDataCH5 (AT91C_BASE_ADC);
-        *a6 = AT91F_ADC_GetConvertedDataCH6 (AT91C_BASE_ADC);
-        *a7 = AT91F_ADC_GetConvertedDataCH7 (AT91C_BASE_ADC);
+                            unsigned int *a1,
+                            unsigned int *a2,
+                            unsigned int *a3,
+                            unsigned int *a4,
+                            unsigned int *a5,
+                            unsigned int *a6,
+                            unsigned int *a7 )
```
stieg: I didn't know that Linux did one parameter per line.  Doesn't seem to inline with kernel code I just reviewed... but meh.

stieg:
It was fun to watch my browser choke on all of these diffs.  Test and merge it.  Thanks for doing this.

brentpicasso:
\o/  I think as a practice we should run this script every time before committing.

PR #47, opened 2015-05-14
V2.8.1 bt provision fix

brentpicasso:
* insert delay before attempting Bluetooth provisioning, allowing BT module to initialize

* set the desired baud rate regardless if we can talk to the BT module. (this was the real culprit)
(when we can't send AT commands, it could be b/c the bt is connected to app. if that's the case, BT module is in transparent mode and won't respond  to AT commands. We have to assume it might already be correctly provisioned in these cases)

* reduce the logging noise when we haven't found a trackmap yet

* astyle formatting


stieg:
@brentpicasso I commented in the patches as the pull itself is noisy with the other two patches in it.

PR #49, opened 2015-05-19
V2.8.2 gps device robusto

brentpicasso:
If the GPS message pump starts timing out, it may not recover, halting GPS message processing. This is likely due to out of sync DMA processing of serial data.

Solution: if this happens, trigger a re-initialization until the GPS message frames remain in sync with DMA.

Also: clean up log messages in the initialization routines.


src/gps/gpsTask.c
```diff
@@ -29,18 +29,21 @@ void GPSTask(void *pvParameters)
     Serial *serial = get_serial(SERIAL_GPS);
     uint8_t targetSampleRate = decodeSampleRate(getWorkingLoggerConfig()->GPSConfigs.speed.sampleRate);
     lapStats_init();
-    gps_status_t gps_status = GPS_init(targetSampleRate, serial);
-    if (!gps_status) {
-        pr_error("GPS: Error provisioning\r\n");
-    }
+    while(1) {
+        gps_status_t gps_status = GPS_init(targetSampleRate, serial);
```
stieg: Either make that variable a `const` or combine the statement into one if statement.

src/gps/gpsTask.c
```diff
@@ -29,18 +29,21 @@ void GPSTask(void *pvParameters)
...
+        if (!gps_status) {
+            pr_error("GPS: Error provisioning\r\n");
```
stieg: So if we fail to provision do we need to backoff?  Why would it work the second time and not the first?

src/gps/gpsTask.c
stieg: Maybe you want to have a log statement here saying you are provisioning?  Maybe its useless.  Thoughts?

src/gps/gpsTask.c
brentpicasso: It seems that GPS message mis alignment causes a problem with how we
process it with dma. This is like a watchdog for the GPS module if
something hiccups in their firmware, too.

Brent Picasso
Autosport Labs

src/gps/gpsTask.c
brentpicasso: Made it const. When you look at the log it's pretty clear it's doing the provisioning work.

-Brent

PR #51, opened 2015-06-19
V2.8.3 connection reset fixes

brentpicasso:
1) read IMEI, cell phone #, and signal strength before we check if we're on the network

2) extend heartbeat timeout to 30 seconds. 10 seconds was too aggressive

3) fix bug where we never reset badMsgCount to zero if we've received a good message


brentpicasso:
Looking at the problem #3) again, I don't think I fixed what I thought I fixed --- I fixed a different problem.  

Looks like the real problem for #3 was that bad messages would never accumulate and therefore never reset the connection.  My fix correctly allows it them to accumulate, but that doesn't address the pre-mature disconnect problem.

I think #2 is the real problem; if we get one bad message, we may miss the timeout, then the connection will reset.

brentpicasso:
:+1:  applied do_style.sh script.

stieg:
The code itself looks fine.  That if/else statement is not Linux style.  This is the style per the Kernel itself: https://www.kernel.org/doc/Documentation/CodingStyle

```c
Note that the closing brace is empty on a line of its own, _except_ in
the cases where it is followed by a continuation of the same statement,
ie a "while" in a do-statement or an "else" in an if-statement, like
this:

	do {
		body of do-loop
	} while (condition);

and

	if (x == y) {
		..
	} else if (x > y) {
		...
	} else {
		....
	}
```

brentpicasso:
Yup, I think the do_style.sh fixed that up for us to conform to the style guide.

PR #52, opened 2015-06-19
Rewrite Filewriter to avoid empty logfiles

stieg:
The old Filewriter code had a bug where it would open a new log file
and immediately close.  This was due to a logic error in a very messy
piece of code.  Rather than just simply fix it (and leave room for other
errors down the line) I redid the method and broke its sub methods out
into smaller and much more readable methods.  Now its much easier to follow
the state and behavior.

Need to add unit tests to ensure this is all chipper.  Also need to consider
how we will handle cases where we will need to roll over to a new file if
the old one hits the 2G file limit on FAT32 systems.

### DIFF
--- a/src/logger/fileWriter.c
+++ b/src/logger/fileWriter.c
@@ -231,12 +231,8 @@ static void flushLogfile(FIL *file)
     }
 }
 
-static void file_led_set(bool on) {
-        if (on) {
-                LED_enable(3);
-        } else {
-                LED_disable(3);
-        }
+static void error_led(bool on) {
+        on ? LED_enable(3) : LED_disable(3);
 }
 
 static void logging_led_toggle() {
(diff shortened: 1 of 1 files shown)


brentpicasso:
Update changelog, please.

stieg:
Sure.  I always forget that.

src/logger/fileWriter.c
```diff
@@ -229,14 +231,28 @@ static void flushLogfile(FIL *file)
     }
 }
 
-static int openNewLogfile(char *filename)
+static void file_led_set(bool on) {
```
brentpicasso: LED 3 is the error LED. function should be called enable_file_error_indicator or something similar and succinct.

src/logger/fileWriter.c
```diff
@@ -244,94 +260,152 @@ static int openNewLogfile(char *filename)
...
-    char filename[FILENAME_LEN];
-
-    while(1) {
+        LoggerMessage *msg = NULL;
+        struct file_status fs = { 0 };
+
         while(1) {
-            //wait for the next sample record
-            xQueueReceive(g_sampleRecordQueue, &(msg), portMAX_DELAY);
+                int ok = 0;
```
brentpicasso: should use bool

src/logger/fileWriter.c
```diff
@@ -244,94 +260,152 @@ static int openNewLogfile(char *filename)
...
+                        pr_warning("Unsupported message type\r\n");
+                        ok = false;
                 }
-                LED_disable(3);
-                if (isTimeoutMs(flushTimeoutStart, flushTimeoutInterval)) {
-                    flushLogfile(g_logfile);
-                    flushTimeoutStart = xTaskGetTickCount();
+
+                /* Turns the LED on if things are bad, off otherwise. */
+                file_led_set(ok != 0);
```
brentpicasso: suggest file_led_set( !ok)

src/logger/fileWriter.c
```diff
@@ -244,94 +260,152 @@ static int openNewLogfile(char *filename)
...
+                if (ok != 0) {
```
brentpicasso: suggest:

if ( !ok ) {

src/logger/fileWriter.c
brentpicasso: well, poop. I see that OK is really a result code. Maybe use rc or similar for variable name?

src/logger/fileWriter.c
```diff
@@ -244,94 +260,152 @@ static int openNewLogfile(char *filename)
...
+                        break;
+                case LoggerMessageType_Stop:
+                        ok = logging_stop(&fs);
                         break;
-                    } else {
-                        LED_disable(3);
-                    }
+                default:
+                        pr_warning("Unsupported message type\r\n");
+                        ok = false;
```
brentpicasso: we're conflating bool type with int? I think it might be confusing- should normalize on one data type.

src/logger/fileWriter.c
stieg: The variable `ok` used to be a `bool` until I had to debug an issue before I pushed this patch up for review.  I'll clean it up.

src/logger/fileWriter.c
stieg: Well... I left this as a logging_led because that is what we are using it for.  File led is ambiguous if we have more than one file in operation.

brentpicasso: file_led_set function activates or de-activates LED #3, which is the error LED. So, the function name should talk about it being the error LED.

stieg: Sorry... brain crossed streams.  This has been renamed to `error_led`

PR #53, opened 2015-06-20
disabling background streaming only affectes telemetry link.

brentpicasso:
Disabling background streaming only affects telemetry link. 

When background streaming is disabled, telemetry starts and stops with SD logging.


stieg:
:shipit:

PR #55, opened 2015-06-20
extend filters to include 500 and 1000 logging rates (MK2 specific fix)

brentpicasso:
we were filtering out 500 and 1000Hz sample rates, making it impossible to do those rates on MK2.


src/logger/loggerConfig.c
```diff
@@ -268,8 +268,14 @@ int encodeSampleRate(int sampleRate)
...
 
     switch(sampleRate) {
+#if MAX_SENSOR_SAMPLE_RATE > 100
+    case 1000:
+        return SAMPLE_1000Hz;
+    case 500:
+        return SAMPLE_500Hz;
     case 200:
         return SAMPLE_200Hz;
+#endif
```
stieg: Might it be simpler to have a MAX_SAMPLE_RATE variable that says simply something like

```c
if (sampleRate > MAX_SAMPLE_RATE)
    return SAMPLE_DISABLED;
```

Then you don't put macros in code and that makes it cleaner.  Could also return SAMPLE_MAX_RATE instead of SAMPLE_DISABLED too.

stieg:
I would personally prefer it if we tried to avoid using C preprocessor directives in code unless absolutely necessary.  They really make C code look ugly fast.  Otherwise ship it.

brentpicasso:
Refined to avoid pre-processor directive; also trimmed up the encoding routine.

src/logger/loggerConfig.c
```diff
@@ -282,34 +288,19 @@ int encodeSampleRate(int sampleRate)
         return SAMPLE_5Hz;
     case 1:
         return SAMPLE_1Hz;
-    default:
     case 0:
```
stieg: `case 0` not needed.  But not harmful.

src/logger/loggerConfig.c
```diff
@@ -282,34 +288,19 @@ int encodeSampleRate(int sampleRate)
...
-        return 1;
-    default:
-    case SAMPLE_DISABLED:
-        return 0;
+    if (rate_code == 0) {
+        return SAMPLE_DISABLED;
+    }
+    else {
+        return TICK_RATE_HZ / rate_code;
     }
```
stieg: Can simplify to

```c
return rate_code ? TICK_RATE_HZ / rate_code : SAMPLE_DISABLED;
```

src/logger/loggerConfig.c
```diff
@@ -263,11 +263,17 @@ int getConnectivitySampleRateLimit()
     return sampleRateLimit;
 }
 
-// STIEG: FIX ME! This can be done with math
+/* Filter sample rates to only allow rates we support */
 int encodeSampleRate(int sampleRate)
 {
-
+    if (sampleRate > MAX_SENSOR_SAMPLE_RATE){
```
stieg: Style fail.  Should be `if (sampleRate > MAX_SENSOR_SAMPLE_RATE) {`.  Also braces not needed for one liner :).

stieg:
It looks like you could simplify the encodeSampleRate method by using math there instead of the case statement.  Would just need to validate that it is a supported value somehow.

brentpicasso:
Thanks. I would've done it with math but figured the case statement was easiest enough to only allow the supported rates.

src/logger/loggerConfig.c
brentpicasso: I love ternary operators! but the one time I opt for it to be more readable ;)  

I'll implement your suggestion.

PR #56, opened 2015-06-20
calculate distance outside of lap timing

brentpicasso:
allows distance to be calculated before lap timing starts. Feedback is that users expect this to happen.


src/lap_stats/lap_stats.c
```diff
@@ -446,6 +444,8 @@ static void lapstats_location_updated(const GpsSnapshot *gps_snapshot)
     const GeoPoint *gp = &gps_snapshot->sample.point;
     const float target_radius = degrees_to_meters(config->TrackConfigs.radius);
 
+    update_distance(gps_snapshot);
+
```
stieg: So they get distance if they haven't configured a start finish.  Seems silly to me.  Thoughts?

brentpicasso: feedback is it should start calculating distance even if we're not in a lap - but resets once we start the beginning of a lap.

stieg:
Change seems sane.  Add a unit test please to ensure that this works as expected.

brentpicasso:
added unit test to check for distance increasing before current lap increments.

PR #58, opened 2015-06-27
Remove all usage of { 0 } from the code

stieg:
Removing all usage of zeroing out by using empty types as this seems
to cause the compiler to use much more ROM space than is necessary
(likely because it is creating a large 0 space in ROM and then copying
it in).

Replace these with the memset function, which is much smaller.  This
yeilds over 1K difference in code size in MK1.


brentpicasso:
This is pretty cool. Nice sleuthing!

brentpicasso:
:shipit:

PR #59, opened 2015-06-27
Add a Vagrant image for use on OSX and Windows

stieg:
Adds the ability to build firmware on any machine using a vagrant image that will set itself up automatically.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,59 +19,16 @@ Vagrant.configure(2) do |config|
   # backing providers for Vagrant. These expose provider-specific options.
   config.vm.provider :virtualbox do |vb|
     # Box name info
-    vb.name = 'ASL Debian Jesse Firmware Builder'
+    vb.name = 'ASL Firmware Builder'
 
     # Needed for multi-core support on Vagrant machines.
     vb.customize ["modifyvm", :id, "--ioapic", "on"]
-    vb.memory = 1536
-    vb.cpus = 3
+    vb.memory = 384
+    vb.cpus = 2
+
+    # Enable USB.  Needed for programming RCP units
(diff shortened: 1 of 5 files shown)


Vagrantfile
```diff
@@ -0,0 +1,46 @@
...
+# you're doing.
+Vagrant.configure(2) do |config|
+  # The most common configuration options are documented and commented below.
+  # For a complete reference, please see the online documentation at
+  # https://docs.vagrantup.com.
+
+  # Every Vagrant development environment requires a box. You can search for
+  # boxes at https://atlas.hashicorp.com/search.
+  config.vm.box = 'holms/debian-jessie-amd64'
+  config.vm.box_url = "https://github.com/holms/vagrant-jessie-box/releases/download/Jessie-v0.1/Debian-jessie-amd64-netboot.box"
```
brentpicasso: How was this box selected? is there an official ultra-generic box we can use?

stieg: I pretty much went to their site and found one that I wanted to play with.  I could change it to any Debian derivative (ie Ubuntu).

brentpicasso: Would prefer a generic box that would likely never go away. Plus gives warm fuzzies to people using this vagrant script - if they see a box from an official source, they will have confidence in using it.

stieg: To be honest this VM is on their ATLAS site.  I have a really hard time telling what is legit with Vagrant and what isn't.  If you have a recommendation I am all ears.  Could probably switch it over to https://atlas.hashicorp.com/ubuntu/boxes/trusty64 if you think that is better.  We can also host a box like that on s3 if needed to guarantee that it works.

PR #61, opened 2015-07-06
Support validation of LogMessages and fix several bugs

stieg:
The main intent of this fix is to validate LogMessages.  Specifically
this means that we validate that the sample buffer that a LogMessage
points to has not been over-written/re-used since the LogMessage was
queued.  To achieve this we timestamp both the LogMessage and the
sample buffer.  The LogMessage is deeply copied on the queue while the
sample buffer isn't.  If when the LogMessage is read we find that the
times are different, then we know the LogMessage is pointing to a
buffer that has been overwritten since the LogMessage was queued and
thus we skip it as it is invalid.  This will in turn prevent us from
recording empty log lines as we are (mostly) ensured that the sample
buffer will not change out from underneath us when reading it.  If
that does begin to happen (as is possible since this is a pre-emptive
system) then we must add a lock to these sample structures to prevent
the behavior.

This change also fixes some very bad behavior where we were popping
LogMessages off of the queue into an undefined space.  This was
occuring because we were allocating pointers to point to areas where
the xQueueReceive should copy its contents to instead of physically
allocating the space.  I'm honestly not quite sure how this never
caused any major issues, but somehow we seemed to have gotten away
with it.  Now the space is clearly defined on the stack.

### DIFF
--- a/src/logger/connectivityTask.c
+++ b/src/logger/connectivityTask.c
@@ -279,17 +279,20 @@ void connectivityTask(void *params)
                     break;
                 }
                 case LoggerMessageType_Sample: {
-                    if (should_stream) {
-                        int sendMeta = (tick == 0 ||
-                                        (connParams->periodicMeta &&
-                                         (tick % METADATA_SAMPLE_INTERVAL == 0)));
-                        api_send_sample_record(serial, msg.sample, tick, sendMeta);
+                        if (!should_stream)
+                                break;
+
+                        const int send_meta = tick == 0 ||
+                                (connParams->periodicMeta &&
+                                 (tick % METADATA_SAMPLE_INTERVAL == 0));
+                        api_send_sample_record(serial, msg.sample, tick, send_meta);
+
                         if (connParams->isPrimary)
-                            toggle_connectivity_indicator();
+                                toggle_connectivity_indicator();
+
                         put_crlf(serial);
                         tick++;
-                    }
-                    break;
+                        break;
                 }
                 default:
                     break;
--- a/src/logger/connectivityTask.c
+++ b/src/logger/connectivityTask.c
@@ -209,7 +209,7 @@ void connectivityTask(void *params)
     size_t rxCount = 0;
 
     ConnParams *connParams = (ConnParams*)params;
-    LoggerMessage *msg = NULL;
+    LoggerMessage msg;
 
     Serial *serial = get_serial(connParams->serial);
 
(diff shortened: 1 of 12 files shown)


src/logger/sampleRecord.c
```diff
@@ -9,15 +9,65 @@
...
+size_t init_sample_buffer(struct sample *s, const size_t count)
+{
+        if (s->channel_samples)
+                free_sample_buffer(s);
+
+        const size_t size = sizeof(ChannelSample[count]);
+        s->ticks = 0;
+        s->channel_count = count;
+        s->channel_samples = (ChannelSample *) portMalloc(size);
+
```
brentpicasso: where do we handle the case if we don't successfully allocate memory?

stieg: It would seem that we don't.  I'll add a check so that if the malloc fails we will return size 0

stieg:
@brentpicasso *BUMP*

brentpicasso:
I will test this directly this weekend.

stieg:
Excellent.

stieg:
Any updates @brentpicasso

src/logger/connectivityTask.c
```diff
@@ -246,38 +246,44 @@ void connectivityTask(void *params)
...
+                           connParams->always_streaming))
                         should_reconnect = true;
                     logging_enabled = false;
                     break;
                 }
                 case LoggerMessageType_Sample: {
                     if (should_stream) {
-                        int sendMeta = (tick == 0 || (connParams->periodicMeta && (tick % METADATA_SAMPLE_INTERVAL == 0)));
-                        api_sendSampleRecord(serial, msg->channelSamples, msg->sampleCount, tick, sendMeta);
+                        int sendMeta = (tick == 0 ||
```
brentpicasso: linux style on variable name

src/logger/loggerTaskEx.c
```diff
@@ -108,24 +102,28 @@ static void logging_stopped()
...
-        if (msg->channelSamples != NULL) {
-            vPortFree(msg->channelSamples);
+        const size_t channel_count = get_enabled_channel_count(loggerConfig);
+        struct sample *s = g_sample_buffer;
+        const struct sample * const end = s + LOGGER_MESSAGE_BUFFER_SIZE;
+        int i;
+
+        for (i = 0; s < end; ++s, ++i) {
+                const size_t bytes = init_sample_buffer(s, channel_count);
+                if (bytes) {
```
brentpicasso: shouldn't this be

if (!bytes) {

?

src/logger/loggerTaskEx.c
```diff
@@ -108,24 +102,28 @@ static void logging_stopped()
...
+                        break;
+                }
         }
-        ChannelSample *channelSamples = create_channel_sample_buffer(loggerConfig, channelSampleCount);
-        init_channel_sample_buffer(loggerConfig, channelSamples, channelSampleCount);
-        msg->channelSamples = channelSamples;
-    }
-    return channelSampleCount;
+
+        pr_info_int_msg("Sample buffers allocated: ", i);
```
brentpicasso: maybe debug instead?

brentpicasso:
@stieg  reviewed

src/logger/loggerTaskEx.c
stieg: Yep.  And this is why I often do the `bytes == 0` check as that is harder to mess up

src/logger/loggerTaskEx.c
stieg: Its a low volume message and I want to see it in standard output so I would prefer to keep it as info.  Perhaps we should setup some loose standards about what messages go where if we think our logging is getting too noisy.  Thoughts?

brentpicasso: To me it is about the audience and not the frequency. Info messages are best for the end user who will be interested in stuff that matters to them, like "did GPS initialize correctly" etc.

stieg: Ok.  I'll keep that in mind going forward and will apply that in a second.

brentpicasso:
:+1:   Merge when ready.  Important update; I'll do some direct testing

stieg:
I'll merge it now into maser.  All tests pass.

On Tue, Jul 28, 2015 at 10:42 AM, Brent Picasso <notifications@github.com>
wrote:

> [image: :+1:] Merge when ready. Important update; I'll do some direct
> testing
>
> —
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/61#issuecomment-125658044>
> .
>



--

PR #63, opened 2015-07-09
V2.8.4 can extended

brentpicasso:
Extended CAN IDs were not handled correctly in firmware.  Fixed by properly handling extended ID flags.

Tested with two RaceCapture/Pro units linked together with a short RJ45 cable

Script running on transmit unit:
setTickRate(1)
function onTick()
txCAN(0, 123456, 1, {22})
txCAN(0, 123, 0, {33})
end

Script running on receiving unit:
setTickRate(10)
function onTick()
id, ext, data = rxCAN(0)
if id ~= nil then
 println(id ..' ' ..ext ..' ' ..data[1])
end
end


stieg:
:tractor: it

PR #65, opened 2015-07-14
Add time info support to LUA

stieg:
Adds the ability to get both system uptime (measured in millis)
and full date time (based on gps data).  The date time had to
be broken into an array field that matches the DateTime struct
because our LUA library does not support a value large enough
to house the milliseconds since epoch value.

Note that this needs manual testing since we currently do not
have unit tests for LUA scripting.


brentpicasso:
Looks good, :shipit:  Please update the documentation here as well:
http://autosportlabs.net/RaceCapturePro_Lua_Scripting

stieg:
Will do.

PR #67, opened 2015-07-23
Add GPS Bearing support

stieg:
Adds GPS bearing support into our logging system.  This allows us to know the bearing of the GPS module.  This is effectively the track of the car in navigational terms.  We can use this in combination with compass heading to get drift angle.


brentpicasso:
looks pretty good so far. You have some missing test files, like geo_point_test.c

stieg:
Oops... I'll add those.

stieg:
Closing.  Will open new pull req.

PR #69, opened 2015-07-27
P/stieg/gps bearing

stieg:
Add bearing support to our GPS subsystem.  This is the gps heading effectively which we will use as a navigational track to help calculate drift angle.


brentpicasso:
data from at this weekend's event this bearing code needs more testing. There are spikes in the heading data that might prove the need for a heavier weighting on the smoothing / averaging, or just a flaw that's not being caught by unit tests.  Emailing sample data.

src/gps/geopoint.c
```diff
@@ -23,6 +23,16 @@ static float toRad(float val)
     return val * (M_PI / 180.0);
 }
 
+/**
+ * Converts a given value to degrees.
+ * @param val The provided value in radian form.
+ * @return The degree value.
+ */
+static float toDeg(float val)
```
brentpicasso: correct style

src/gps/gps.c
```diff
@@ -177,6 +201,7 @@ void GPS_sample_update(GpsSample *newSample)
     updateFullDateTime(newSample);
     g_gpsSnapshot.deltaFirstFix = newSample->time - g_timeFirstFix;
     g_gpsSnapshot.previousPoint = prevPoint;
+    g_gpsSnapshot.bearing = gps_bearing(&prevPoint, &(newSample->point));
```
brentpicasso: ah, I think this is what's causing the spikey behavior, calculating it across adjacent GPS points.

Consider running it through our Exponential Moving Average IIR filter - see filter.c. it's what we use to apply smoothing to our analog and timer/pulse channels.

include/gps/gps.h
```diff
@@ -137,6 +138,12 @@ tiny_millis_t getUptimeAtSample();
 
 float getGpsSpeedInMph();
 
+/**
+ * @return The bearing of the GPS path.  Otherwise known as the track.
+ * Relative to previous point.
+ */
+float get_gps_bearing();
```
brentpicasso: function should have void parameter

float get_gps_bearing(void)

src/gps/gps.c
```diff
@@ -159,6 +178,11 @@ GpsSnapshot getGpsSnapshot()
     return g_gpsSnapshot;
 }
 
+float get_gps_bearing()
```
brentpicasso: ditto on void function parameter

include/logger/loggerConfig.h
```diff
@@ -357,17 +358,19 @@ typedef struct _GPSConfig {
...
-		DEFAULT_GPS_LATITUDE_CONFIG,           \
-		DEFAULT_GPS_LONGITUDE_CONFIG,          \
-		DEFAULT_GPS_SPEED_CONFIG,              \
-		DEFAULT_GPS_DISTANCE_CONFIG,           \
-		DEFAULT_GPS_ALTITUDE_CONFIG,           \
-		DEFAULT_GPS_SATELLITE_CONFIG,          \
-		DEFAULT_GPS_QUALITY_CONFIG,            \
-		DEFAULT_GPS_DOP_CONFIG                 \
-         }
+#define DEFAULT_GPS_BEARING_CONFIG {"Bearing", "Degrees", 0, 360, DEFAULT_GPS_SAMPLE_RATE, 1, 0}
```
brentpicasso: logging precision should be zero (heading in whole numbers)

src/gps/gps.c
stieg: I'm not convinced that am EMA filter is the right solution here as the values can change quickly and we don't want any delay in there.  This sort of filter would be much better suited for our gas gauge sender to smooth it out (as it really really needs it).  Perhaps a minimum speed or minimum distance would be good here.  I'm leaning towards the latter since that way we can get gps bearing at lower speeds with minimal extra memory usage.

stieg:
I am thinking I will change the name of the method to get_gps_heading as that is more accurate.  bearing is the wrong term for this properly speaking

stieg:
rebasing onto v2.9.0 and did some force pushing so need to re-open this in a new commit.

src/gps/gps.c
brentpicasso: I agree that it wouldn't be appropriate in this case. Neither would speed based, as the demo data showed glitches on reasonably high speeds (> 10 mph) I think the only reliable way to do it is to only calculate it based on a distance threshold.

PR #74, opened 2015-08-02
P/stieg/fix filewriter logging

stieg:
Fixes the meta issue.  Fixes the f_write issue.  Big fixes :)


src/logger/fileWriter.c
```diff
@@ -34,47 +35,55 @@
...
-static void clear_file_buffer() {
-        fileBuffer.index = 0;
-        fileBuffer.buffer[0] = '\0';
+static void error_led(bool on)
+{
+        on ? LED_enable(3) : LED_disable(3);
 }
 
-static int writeFileBuffer()
+static FRESULT flush_file_buffer()
```
brentpicasso: should use void param

src/logger/fileWriter.c
```diff
@@ -34,47 +35,55 @@
...
 {
-        unsigned int bw;
-        const int rc = f_write(g_logfile, fileBuffer.buffer,
-                               fileBuffer.index, &bw);
-        clear_file_buffer();
-        return rc;
+        FRESULT res = FR_OK;
+        char tmp[32];
+
+        pr_info("Flushing file buffer\r\n");
```
brentpicasso: everywhere we do logging we should prefix it with the module / area doing the logging.  format is <moduleName>: <msg>

So in this case:

pr_info("fileWriter: flushing file buffer\r\n");

applies throughout.

Also, this particular message should be debug.

src/logger/fileWriter.c
```diff
@@ -34,47 +35,55 @@
...
+        size_t chars = get_used(&file_buff);
+        while(0 < chars) {
+                if (chars > sizeof(tmp))
+                        chars = sizeof(tmp);
+
+                pr_info_int_msg("Chars is ", chars);
```
brentpicasso: s/b debug, trace or removed.

src/logger/fileWriter.c
```diff
@@ -34,47 +35,55 @@
...
-        *buffer++ = *data++;
-        index++;
-        if (index >= FILE_BUFFER_SIZE) {
-            *buffer = '\0';
-            writeFileBuffer();
-            index = fileBuffer.index;
-            buffer = fileBuffer.buffer + index;
+        FRESULT res = FR_OK;
+
+        const size_t size = strlen(data);
```
brentpicasso: think there could be a compliment to ring buffer's put_data that is called put_str that would avoid having to use strlen(), taking advantage of the assumption that strings are null terminated? it could avoid repeated called to strlen() in that case.

src/ring_buffer.c
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+#include "mem_mang.h"
+#include "mod_string.h"
+#include "ring_buffer.h"
+
+#include <stdbool.h>
+#include <stddef.h>
+
+static size_t get_end_dist(struct ring_buff *rb, const char *p)
```
brentpicasso: should all of the ringbuffer function be prefixed so we know the functions are related? seems like a common pattern for libraries and similar. 

rb_get_end_dist(...

or

ringbuf_get_end_dist(..

SAM7s_base/fat_sd_at91/ff.c
```diff
@@ -4497,7 +4510,7 @@ int f_putc (
 /* Put a string to the file                                              */
 /*-----------------------------------------------------------------------*/
 
-static int f_puts_unprotected (
+int f_puts (
```
brentpicasso: neat and clean solution.

src/ring_buffer.c
stieg: If I did that we should do it separately.  And also probably come up with a common trait for this so it can apply consistently throughout.  But I think its a good idea.

SAM7s_base/fat_sd_at91/ff.c
stieg: :dart:

src/logger/fileWriter.c
stieg: Not a bad idea.  Should probably fix up some MACROS to make this eaiser.

src/logger/fileWriter.c
stieg: Of course there could be but I don't see the advantage because strlen is only called once per string added.  Plus no matter if I call strlen or write a put_str method, both have to iterate over the string so nothing is really saved IMHO.  Plus then i still need to account for partial string puts which honestly won't make the code any simpler.  So its possible... but I don't quite see the benefit.  Perhaps I am missing something?

brentpicasso: The big thing with a special string function is to possibly avoid iterating
the string twice. If you loop once until the null terminator then you'd
avoid the strlen.

Brent Picasso
Autosport Labs

stieg: I suggest reading the put_data method as I think there is some confusion here.  I don't think your solution will work the way you intend it to work.  Remember that strings can be of arbitrary length, but put_data can only take a finite amount.  And no matter what, we always only iterate over a string once per append call.

stieg:
@brentpicasso Feedback applied.  Have another look and let me know what you think.

src/logger/fileWriter.c
```diff
@@ -149,13 +152,13 @@ static int write_samples_data(const LoggerMessage *msg)
         size_t count = msg->sample->channel_count;
 
         if (NULL == sample) {
-                pr_warning("Logger: null sample record\r\n");
+                pr_warning("_RCP_BASE_FILE_: null sample record\r\n");
```
brentpicasso: Could we use 

fileWriter: <msg>

as it's a bit easier to read (IMO) and is consistent with the other parts of the code that Log stuff?

src/ring_buffer.c
brentpicasso: Can be a separate refactor. I'm already doing it with the cellModem and sim900_ function prefixes.

src/logger/fileWriter.c
stieg: We could.  I have an idea... lets establish what we want going forward.  Then once we have something we are all happy with then we can go back and fix this?  Deal?

stieg:
I'll add in a put_string method tomorrow.

src/logger/fileWriter.c
brentpicasso: Ok. Just keep in mind what the logging output looks like currently - this format will stand out quite differently than the current standard look and feel formatting.

brentpicasso: e.g. current example output

lua: memory usage: 7Kb
BT: Probing baud 115200
GPS: provisioning attempt
GPS: probing baud rate: 921600
GPS: module detected at: 921600
GPS: detected current update rate: 50                                           
GPS: Configuring update rate to 10: win                                         
GPS: Configuring NMEA messages: win                                             
GPS: configure message type: win                                                
GPS: configure navigation data message interval: win                            
GPS: Configuring Gnss Navigation Mode: win                                      
GPS: Disable NMEA messages: win                                                 
GPS: provisioned                                                                
BT: Provision fail                                                              
BT: Probing baud 9600                                                           
BT: Provision fail                                                              
BT: Probing baud 230400                                                         
BT: Provision fail                                                              
BT: Failed to provision module. Assuming already connected.

stieg: Will it?  It will be much more consistent than before (it varied between "logger", "fileLogger", "fileWriter", etc).  It will look like `fileWriter.c: Some unoriginal message`

brentpicasso: Oh, I didn't realize it would swap the string. ok.  Would be nice if the .c was stripped somehow.  That is all :)

stieg: Ahh... yes.  Its a macro.  Lets figure out exactly what we want and how we want to define it... then we can implement a special macro for it and have it just work.

stieg:
@brentpicasso OK... put_string method added along with unit tests for all of ring_buffer.  Can we merge this now?

brentpicasso:
I'll pull it down to test on my unit, just to be safe. then we'll merge!

src/ring_buffer.c
```diff
@@ -0,0 +1,152 @@
...
+                size_t wrap_size = size - dist;
+                memcpy(rb->head, data, dist);
+                memcpy(rb->buf, dist + (char *)data, wrap_size);
+                rb->head = rb->buf + wrap_size;
+        }
+
+        return size;
+}
+
+const char * put_string(struct ring_buff *rb, const char *str)
```
brentpicasso: nice.

src/logger/fileWriter.c
stieg: https://trello.com/c/f5SizLe0/308-establish-logging-prefix-standard-for-firmware-and-go-with-it tracks doing this.

include/ring_buffer.h
```diff
@@ -0,0 +1,91 @@
+/**
```
brentpicasso: Just saw this - can we put ring_buffer into it's own directory  or an existing one, like util?

brentpicasso:
There's a problem with logging.  I don't see any of the logging messages for fileWriter.  example output. Note the last couple of lines:

RaceCapture/Pro MK2 > setLogLevel 7
result="ok";
RaceCapture/Pro MK2 > viewLog
Starting logging mode.  Hit "q" to exit
*** Start! ***
CAN0 init @ 500000 win
CAN1 init @ 500000 win
_RCP_BASE_FILE_: Failed to alloc ring buffer.
timebase/logging/telemetry sample rate: 50/25/25
Start OBD2 task
IMU: init res=0
running lua script len(99)...timebase/logging/telemetry sample rate: 50/25/25
done
lua: memory usage: 7Kb
BT: Probing baud 115200
GPS: provisioning attempt
GPS: probing baud rate: 921600
GPS: module detected at: 921600
GPS: detected current update rate: 50
GPS: Configuring update rate to 10: win
GPS: Configuring NMEA messages: win
GPS: configure message type: win
GPS: configure navigation data message interval: win
GPS: Configuring Gnss Navigation Mode: win
GPS: Disable NMEA messages: win
GPS: provisioned
BT: Provision win
BT: Init complete
Logging started
Logging stopped

include/ring_buffer.h
stieg: Util would work.

PR #76, opened 2015-08-03
Add unit tests for ring_buffer.h

stieg:
Adds unit tests for our ring_buffer struct because its value is
very high in our code and we must be confident that it is working
correctly.


stieg:
Merged as part of Review #74

PR #78, opened 2015-08-10
Update the README file to support completely setting up the MK2 toolchain

stieg:
Because it was incomplete.


stieg:
Trivial and not worth review time.  Merging now.

PR #79, opened 2015-09-08
Bump GPS task stack size to 256 bytes from 200

stieg:


brentpicasso:
i arbitrarily picked 300 vs 256, but 256 is still a big bump. will ship it.

stieg:
Merged onto the release branch.

PR #80, opened 2015-09-08
Bump max track distance from 1 to 5 kilometers

stieg:
Because of customer complaints.  Also fixes headers.


brentpicasso:
Looks good

PR #81, opened 2015-09-10
Turn off the LUA memory optimizations

stieg:
The memory opts are supposed to reduce memory usage while preseving
basic functionality of the various libraries that we implement.
However this was not the case as LUA was crashing whenever someone
would try to access these libraries.  Thus this change works around
the issue simply by disabling the memory optimizations.  The tradeoff
is ~1KB more usage of RAM compared to before.

### DIFF
--- a/lib_lua/src/Makefile
+++ b/lib_lua/src/Makefile
@@ -7,12 +7,18 @@
 # Your platform. See PLATS for possible values.
 PLAT= none
 
+#
+# Disable Memory Opts.
+#
+# Disable the Memory optimizations as they are causing various system
+# libraries not to get registered.  The extra memory seems to be about
+# 1K.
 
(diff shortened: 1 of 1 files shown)


brentpicasso:
so we're still using Lua_Register_light - iirc that was the big one for improving memory usage.

If this works well with MK1, then :shipit:

stieg:
Working on testing MK1.  My bootloader pee'd the bed and I am still hunting for the JTAG adapter which is burried in a box somewhere.

brentpicasso:
Is this good to go for both platforms?

stieg:
Was going to rebase onto the latest MK1 fix and check.

stieg:
Integrated this into another pull.  Hence closing

PR #83, opened 2015-09-16
LUA Math Fix and Memory Games

stieg:
So some of the LUA math methods were broken.  This fixes that.

This also addresses the MK1 and LUA issues that have been plaguing us since 2.8.0 release of firmware.  This works around the issue by releasing LUA resources before we program the LUA script (which is a memory hungry operation because we have done it sub optimally).  By doing this we free up enough of the heap to get things working again.

### DIFF
--- a/src/logger/loggerApi.c
+++ b/src/logger/loggerApi.c
@@ -1,58 +1,61 @@
-/**
- * AutoSport Labs - Race Capture Pro Firmware
+/*
+ * Race Capture Pro Firmware
  *
- * Copyright (C) 2014 AutoSport Labs
+ * Copyright (C) 2015 Autosport Labs
  *
- * This file is part of the Race Capture Pro firmware suite
+ * This file is part of the Race Capture Pro fimrware suite
  *
- * This is free software: you can redistribute it and/or modify it under the terms of the
- * GNU General Public License as published by the Free Software Foundation, either version 3 of the
- * License, or (at your option) any later version.
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
  *
- * This software is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
- * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  *
- * See the GNU General Public License for more details. You should have received a copy of the GNU
- * General Public License along with this code. If not, see <http://www.gnu.org/licenses/>.
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
  */
+
+#include "ADC.h"
 #include "FreeRTOS.h"
-#include "task.h"
-#include "constants.h"
+#include "GPIO.h"
+#include "PWM.h"
+#include "bluetooth.h"
 #include "capabilities.h"
+#include "cellModem.h"
+#include "constants.h"
+#include "cpu.h"
+#include "dateTime.h"
+#include "geopoint.h"
+#include "gps.h"
+#include "imu.h"
+#include "lap_stats.h"
+#include "launch_control.h"
+#include "logger.h"
 #include "loggerApi.h"
 #include "loggerConfig.h"
-#include "modp_atonum.h"
-#include "mod_string.h"
-#include "sampleRecord.h"
-#include "loggerSampleData.h"
 #include "loggerData.h"
-#include "loggerNotifications.h"
-#include "imu.h"
-#include "tracks.h"
 #include "loggerHardware.h"
-#include "serial.h"
-#include "mem_mang.h"
-#include "printk.h"
-#include "geopoint.h"
-#include "timer.h"
-#include "ADC.h"
-#include "imu.h"
-#include "PWM.h"
-#include "cpu.h"
+#include "loggerNotifications.h"
+#include "loggerSampleData.h"
+#include "loggerTaskEx.h"
 #include "luaScript.h"
 #include "luaTask.h"
-#include "logger.h"
-#include "loggerTaskEx.h"
-#include "FreeRTOS.h"
-#include "taskUtil.h"
-#include "GPIO.h"
-#include "gps.h"
-#include "dateTime.h"
-#include "cellModem.h"
-#include "bluetooth.h"
+#include "mem_mang.h"
+#include "mod_string.h"
+#include "modp_atonum.h"
+#include "printk.h"
+#include "sampleRecord.h"
+#include "serial.h"
 #include "sim900.h"
-#include "launch_control.h"
-#include "lap_stats.h"
+#include "task.h"
+#include "taskUtil.h"
+#include "timer.h"
+#include "tracks.h"
+
 #include <stdbool.h>
 
 /* Max number of PIDs that can be specified in the setOBD2Cfg message */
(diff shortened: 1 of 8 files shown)


src/lua/luaTask.c
```diff
@@ -124,117 +146,206 @@ size_t get_ontick_freq()
...
 
-int getShouldReloadScript(void)
+void* getLua()
 {
-    return g_shouldReloadScript;
+    return g_lua;
 }
 
-void setShouldReloadScript(int reload)
+static void _terminate_lua()
```
brentpicasso: why the leading underscore for the function?

src/lua/luaTask.c
```diff
@@ -124,117 +146,206 @@ size_t get_ontick_freq()
...
-    luaopen_string(g_lua);
-    luaopen_math(g_lua);
-    registerBaseLuaFunctions(g_lua);
-    registerLuaLoggerBindings(g_lua);
-    unlockLua();
+        lockLua();
+
+        pr_info("lua: Stopping...\r\n");
+
+        if (LUA_DISABLED == lua_run_state)
```
brentpicasso: can we avoid the use of goto here? We can test if the run state is positive and then terminate lua. It reads more naturally. 

e.g. "If lua is running then terminate it"

src/lua/luaTask.c
```diff
@@ -124,117 +146,206 @@ size_t get_ontick_freq()
...
-        doScript();
-    } else {
-        pr_error("lua: Crash detected- bypassing Lua script for recovery mode\r\n");
-        LED_enable(3);
-    }
+        lockLua();
+
+        lua_getglobal(g_lua, method);
+        if (lua_isnil(g_lua, -1)) {
+                pr_error("Can't find \"");
```
brentpicasso: You could use the printk one-liner for string variables here and elsewhere (pr_error_string ...)

src/lua/luaTask.c
```diff
@@ -124,117 +146,206 @@ size_t get_ontick_freq()
...
+                pr_error(method);
+                pr_error("\"\r\n");
+
+                lua_pop(g_lua, 1);
+                _terminate_lua();
```
brentpicasso: 2 potential problems here:

If the user creates a long running script in lua - such as a loop - outside of the onTick() function, and there is no onTick() defined, this may pull the rug out from underneath. Or, maybe not, if the script runs synchronously 

Also, if no onTick() function is defined and then the user then wants to run in interactive mode later, what happens?

src/lua/luaTask.c
```diff
@@ -124,117 +146,206 @@ size_t get_ontick_freq()
...
+                goto out;
+        }
+
+        if (0 != lua_pcall(g_lua, 0, 0, 0)) {
+                pr_error("lua method error: ");
+                pr_error(lua_tostring(g_lua, -1));
+                pr_error("\r\n");
+
+                lua_pop(g_lua, 1);
+                _terminate_lua();
```
brentpicasso: I'm noting the aggressive termination of Lua on errors seems to be out of scope of the stated intent of the PR.

We need to understand what conditions will cause an error in lua_pcall(), because if there's a runtime error, this new change will now terminate lua. 

What if there's a spurious error in a user's script ? Should we terminate the lua runtime immediately and shut everything down forever? e.g. a spurious divide by zero might cause lua_pcall() to return an error code (IDK if it does)

The current behavior is to log and allow the call to happen again, e.g. it keeps trying.

src/lua/luaTask.c
```diff
@@ -124,117 +146,206 @@ size_t get_ontick_freq()
...
+                goto out;
+        }
+
+out:
+        unlockLua();
 }
 
-void luaTask(void *params)
+void run_lua_interactive_cmd(Serial *serial, const char* cmd)
```
brentpicasso: what if lua is disabled when this function is called?

src/lua/luaTask.c
stieg: Private method.  Has no locks.

src/lua/luaTask.c
stieg: We can, but I would prefer not to.  Its a nice flow structure that is easy to build onto and avoids lots of unnecessary indentation.  I use it when I have to free or unlock resources.

src/lua/luaTask.c
stieg: Good idea.  Will incorporate.

src/lua/luaTask.c
stieg: > If the user creates a long running script in lua - such as a loop - outside of the onTick() function, and there is no onTick() defined, this may pull the rug out from underneath. Or, maybe not, if the initial script runs synchronously.

If this happens, then the script will never load because it will get stuck processing things forever (happens when we first load up the script and the code is parsed).  So not a new issue: we have always had it.

> Also, if no onTick() function is defined and then the user then wants to run in interactive mode later, what happens?

Good call.  Currently it simply would not run because the lua environment would be de-initialized.  I'll review this logic and fix it up :).

src/lua/luaTask.c
stieg: > I'm noting the aggressive termination of Lua on errors seems to be out of scope of the stated intent of the PR.

Correct.  I got a little zealous here :).  I understand the concerns RE killing the script of a pcall fails.  I can pull out the agressive killing.

src/lua/luaTask.c
stieg: Then Segfault because no Lua Environment.  Good catch.  Will fix.

stieg:
Thanks for the good review @brentpicasso.  I'll incorporate that feedback into another patch to address these concerns.

src/lua/luaTask.c
brentpicasso: Cool. I think that would be good to pull out the aggressive killing. I can see a case where a one-time script exists without an onTick() and then the user wants to do interactive mode.

brentpicasso:
:+1:  good changes overall!

src/lua/luaTask.c
brentpicasso: Think all static methods should follow this convention then?

src/lua/luaTask.c
brentpicasso: I see where you did it in the other function, and that function is much longer / complex where this 'cleanup' effect is more effective. This looks a little pedantic, but I understand where you're going.

brentpicasso: maybe change out: label to cleanup:  would be even more explicit to the first time viewer.

src/lua/luaTask.c
stieg: No.  I did this because there is already an extern method named `terminate_lua` that can get called at any time.  I needed a private internal version that was a subset of that public method (in this case one without locks).  Hence why I did that.  Its a common convention in the Linux kernel.  Usually it denotes that you "should really know what you are doing before you use it".  2 underscores exemplifies that even more.

src/lua/luaTask.c
stieg: Ok.  I will do that.

stieg:
@brentpicasso Ready for review again at your leisure.

src/logger/loggerApi.c
```diff
@@ -1,58 +1,61 @@
-/**
- * AutoSport Labs - Race Capture Pro Firmware
+/*
+ * Race Capture Pro Firmware
  *
- * Copyright (C) 2014 AutoSport Labs
+ * Copyright (C) 2015 Autosport Labs
  *
- * This file is part of the Race Capture Pro firmware suite
+ * This file is part of the Race Capture Pro fimrware suite
```
brentpicasso: typo

src/lua/luaTask.c
brentpicasso: :+1:

brentpicasso:
Looks good. If you're satisfied with testing on MK1 and MK2 then please :shipit:

src/logger/loggerApi.c
stieg: Ooohhh... yeah that is from a template... which means its everywhere :(.  Damnit.  Will fix separately.

PR #85, opened 2015-09-19
P/stieg/fix lua printing precision

stieg:
Fixes the printing issue and cleans up some code.
```
lua: memory usage: 7Kb
Lat: 43.085308, Lon: -89.376518
Lat: 43.085342, Lon: -89.376495
Lat: 43.085342, Lon: -89.376495
Lat: 43.085342, Lon: -89.376495
Lat: 43.085342, Lon: -89.376495
Lat: 43.085342, Lon: -89.376495
Lat: 43.085342, Lon: -89.376495
```

### DIFF
--- a/lib_lua/src/luaconf.h
+++ b/lib_lua/src/luaconf.h
@@ -515,7 +515,7 @@
 */
 #define LUAI_UACNUMBER	double
 
-#define LUA_NUMBER_PRECISION 3
+#define LUA_NUMBER_PRECISION 6
 
 
 /*
@@ -790,4 +790,3 @@ union luai_Cast {
 #endif
 
 #endif
-
--- a/src/lua/luaBaseBinding.c
+++ b/src/lua/luaBaseBinding.c
@@ -1,21 +1,36 @@
 /*
- * luaBaseBinding.c
+ * Race Capture Pro Firmware
  *
- *  Created on: May 10, 2011
- *      Author: brent
+ * Copyright (C) 2015 Autosport Labs
+ *
+ * This file is part of the Race Capture Pro fimrware suite
+ *
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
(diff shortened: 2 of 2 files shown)


lib_lua/src/luaconf.h
```diff
@@ -515,7 +515,7 @@
 */
 #define LUAI_UACNUMBER	double
 
-#define LUA_NUMBER_PRECISION 3
+#define LUA_NUMBER_PRECISION 6
```
brentpicasso: In reality, the vast majority of the time we're printing out variables other than Lat/Log (CAN bus values, debug values, analog channel values, IMU values, etc. This will force all output to 6 decimals of precision, causing a usability issue with scripts.

I think we can have the 6 precision if we modify the ftoa conversion to strip trailing zeros (and the decimal point if needed). That way, we can have the best of both worlds - full precision when we need it, and concise output when possible.

Bonus: have lua use our common ftoa source

stieg: > causing a usability issue with scripts.

I'm curious what usability issues you may be referring to.  Example?

> I think we can have the 6 precision if we modify the ftoa conversion to strip trailing zeros (and the decimal point if needed). That way, we can have the best of both worlds - full precision when we need it, and concise output when possible.

Not a bad idea.  Probably want a new method for it (like `ftoa_trim`) so that it doesn't break expectations for others using it.  Should be a pretty easy algorithm.

> Bonus: have lua use our common ftoa source

If you look at line 530 you can see that lua is already using our ftoa source code.

brentpicasso:
The usability impact is just printing any number value to the log or console, even if you're treating it like an integer.  e.g.

i = 1
println("hello: " ..i)

would look like


hello 1.000000

trimming the trailing zeros would have it be much more natural.

brentpicasso:
I may be inaccurate on this. I think the routine might trim some trailing zeros after all.  Worth digging in.

brentpicasso:
it looks like we're stripping trailing zeros already. 

Entering Lua Interpreter. enter 'exit' to leave                                                                  
> i=1                                                                                                            
> println(i)                                                                                                     
1.0                                                                                                              
> i=1.2345678                                                                                                    
> println(i)                                                                                                     
1.234568                                                                                                         
> i=1.23456789                                                                                                   
> println(i)                                                                                                     
1.234568                                                                                                         
> i=1.2000001                                                                                                    
> println(i)                                                                                                     
1.2            
> i=1.1000000                                                                                                    
> println(i)                                                                                                     
1.1

brentpicasso:
Make sure you update CHANGELOG and :shipit:

stieg:
Damn... I always forget the CHANGELOG.

PR #86, opened 2015-09-19
Don't include OBDII channels in the stream when OBDII is disabled

stieg:
Fixes the issues where we were foolishly including OBDII data in
the stream when the OBDII system was disabled.  No longer shall
we do this.


brentpicasso:
:shipit:

PR #87, opened 2015-09-21
Usart interrupt hang

brentpicasso:
if an overrun interrupt fires then we may be stuck in an infinite loop until watchdog occurs. This watches for that flag and clears it.


stm32_base/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -699,6 +699,16 @@ void USART1_IRQHandler(void)
         xQueueSendFromISR(xUsart0Rx, &cChar, &xTaskWokenByPost);
     }
 
+    if (USART_GetITStatus(USART1, USART_IT_ORE_RX) != RESET) {
+        /* Handle Overrun error
+         * This bit is set by hardware when the word currently being received in the shift register is
+           ready to be transferred into the RDR register while RXNE=1. An interrupt is generated if
+           RXNEIE=1 in the USART_CR1 register. It is cleared by a software sequence (an read to the
+           USART_SR register followed by a read to the USART_DR register) */
```
stieg: Fix your comment formatting.

stieg: So... I'm confused.  This gets set every time that RXNE is set?  Or is it set when we don't read a character out of the receive buffer fast enough and the buffer overruns?

stm32_base/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -729,6 +739,15 @@ void USART2_IRQHandler(void)
 //		xQueueSendFromISR( xUsart2Rx, &cChar, &xTaskWokenByPost );
     }
 
+    if (USART_GetITStatus(USART2, USART_IT_ORE_RX) != RESET) {
```
stieg: Repeating code BAD.  Make this a method if you can that checks and handles this case.

stm32_base/hal/usart_stm32/usart_device_stm32.c
brentpicasso: Bit 3 ORE: Overrun error
This bit is set by hardware when the word currently being received in the shift register is
ready to be transferred into the RDR register while RXNE=1. An interrupt is generated if
RXNEIE=1 in the USART_CR1 register. It is cleared by a software sequence (an read to the
USART_SR register followed by a read to the USART_DR register).
0: No Overrun error
From page 1000 of stm32f4 reference manual

1: Overrun error is detected
Note: When this bit is set, the RDR register content will not be lost but the shift register will be
overwritten. An interrupt is generated on ORE flag in case of Multi Buffer
communication if the EIE bit is set.

From my understanding what's happening is that the ORE flag is set when the shift register is getting populated and the data hasn't been read yet. it's like a warning flag. when this happens, we never clear the interrupt in the way the processor wants it to be, so the interrupt fires continuously until watchdog occurs.

stm32_base/hal/usart_stm32/usart_device_stm32.c
brentpicasso: I did this on purpose to make the IRQ handler as fast as possible. inline keyword maybe?

stieg: Dude... let the optimizer handle that.  That is what it is there for.  Writing clean small scope code should be your concern.  If you are thinking about that here, you are over optimizing.  3 cycles to push frames and do a call is negligible considering the speed of this processor.

stm32_base/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -699,6 +699,16 @@ void USART1_IRQHandler(void)
         xQueueSendFromISR(xUsart0Rx, &cChar, &xTaskWokenByPost);
     }
 
+    if (USART_GetITStatus(USART1, USART_IT_ORE_RX) != RESET) {
+        /* Handle Overrun error
```
stieg: Not fixed.  Here is what is _should_ look like:

```C
/*
 * Title
 *
 * Text text text text
 * text text text text
 */

File: stm32_base/hal/usart_stm32/usart_device_stm32.c [hunk 4]
```diff
@@ -648,6 +648,22 @@ int usart3_readLine(char *s, int len)
 // Interrupt Handlers
 ////////////////////////////////////////////////////////////////////////////
 
+inline static void handle_usart_overrun(USART_TypeDef* USARTx)
```
stieg: Don't use `inline`.  Let the optimizer handle that.

stm32_base/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -648,6 +648,22 @@ int usart3_readLine(char *s, int len)
...
+{
+	uint32_t cChar;
```
stieg: Looks like alignment is off somewhere.

stm32_base/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -648,6 +648,22 @@ int usart3_readLine(char *s, int len)
...
+    if (USART_GetITStatus(USARTx, USART_IT_ORE_RX) != RESET) {
```
stieg: So this is a good example of where I would use a return style if check here to avoid indentation.  I'm not saying do this, but consider it.  I might do something like

```C
 if (USART_GetITStatus(USARTx, USART_IT_ORE_RX) != SET)
   return;

...
```

See how that could save you lots of indentation.  This way also works.

stieg:
Fix minor alignment/indentation issues, remove the inline hint and :ship: it

stm32_base/hal/usart_stm32/usart_device_stm32.c
brentpicasso: :+1:

PR #88, opened 2015-09-22
allow user to request a lua runtime bypass in extreme cases 

brentpicasso:
...(if lua script or runtime is causing a repeated watchdog, effectively soft-bricking unit)

* added ability for user to signal a bypass of lua runtime if unit powers up under watchdog reset
User signals lua runtime to be bypassed within 5 seconds of watchdog reset by pressing the front-panel button after power up. if user does not press the front-panel button, Lua runtime will be loaded normally. During this phase the red led will toggle at 250ms intervals.

One usage problem is that pressing the button during this phase will also trigger logging. I think this is an acceptable issue due to the extreme nature of the recovery process, and the limited use it should experience.

Also various clean up of logger messages and function calls.

### DIFF
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -41,7 +41,6 @@
 
 #include <stdbool.h>
 
-#define ERROR_LED 3
 #define DEFAULT_ONTICK_HZ 1
 #define MAX_ONTICK_HZ 30
 #define LUA_STACK_SIZE 1000
@@ -217,16 +216,16 @@ static bool user_bypass_requested(void)
 {
         pr_info("lua: Checking for Lua runtime bypass request\r\n");
         bool bypass = false;
-        LED_disable(ERROR_LED);
+        LED_disable(LED_ERROR);
         for (size_t i = 0; i < LUA_BYPASS_FLASH_COUNT; i++) {
-            LED_toggle(ERROR_LED);
+            LED_toggle(LED_ERROR);
             if (GPIO_is_button_pressed()) {
                 bypass = true;
                 break;
             }
             delayMs(LUA_BYPASS_FLASH_DELAY);
         }
-        LED_disable(ERROR_LED);
+        LED_disable(LED_ERROR);
         return bypass;
 }
 
@@ -328,10 +327,10 @@ static void luaTask(void *params)
         set_ontick_freq(DEFAULT_ONTICK_HZ);
         initialize_script();
 
-        bool should_bypass_lua = (watchdog_is_watchdog_reset() && user_bypass_requested());
+        const bool should_bypass_lua = (watchdog_is_watchdog_reset() && user_bypass_requested());
         if (should_bypass_lua) {
                 pr_error("lua: Bypassing Lua Runtime\r\n");
-                LED_enable(ERROR_LED);
+                LED_enable(LED_ERROR);
         } else {
                 initialize_lua();
         }
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -37,13 +37,19 @@
 #include "lua.h"
 #include "lauxlib.h"
 #include "lualib.h"
+#include "GPIO.h"
 
 #include <stdbool.h>
 
+#define ERROR_LED 3
 #define DEFAULT_ONTICK_HZ 1
 #define MAX_ONTICK_HZ 30
 #define LUA_STACK_SIZE 1000
 #define LUA_PERIODIC_FUNCTION "onTick"
+
+#define LUA_BYPASS_FLASH_DELAY 250
+#define LUA_BYPASS_DELAY_SEC 5
+#define LUA_BYPASS_FLASH_COUNT (1000 / LUA_BYPASS_FLASH_DELAY) * LUA_BYPASS_DELAY_SEC
 
 static enum {
         LUA_DISABLED = 0,
@@ -179,7 +185,7 @@ static bool _load_script(void)
         const char *script = getScript();
         const size_t len = strlen(script);
 
-        pr_info("Loading lua script (len = ");
+        pr_info("lua: Loading lua script (len = ");
         pr_info_int(len);
         pr_info("): ");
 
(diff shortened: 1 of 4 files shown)


src/lua/luaTask.c
```diff
@@ -37,14 +37,20 @@
 #include "lua.h"
 #include "lauxlib.h"
 #include "lualib.h"
+#include "GPIO.h"
 
 #include <stdbool.h>
 
+#define ERROR_LED 3
```
stieg: This should be defined in the LED.h file, not here.

src/lua/luaTask.c
```diff
@@ -37,14 +37,20 @@
...
 #define DEFAULT_ONTICK_HZ 1
 #define MAX_ONTICK_HZ 30
 #define LUA_STACK_SIZE 1000
 #define LUA_PERIODIC_FUNCTION "onTick"
 
+#define LUA_BYPASS_FLASH_DELAY 250
```
stieg: Also probably should live in LED.h.  Thoughts?

src/lua/luaTask.c
```diff
@@ -37,14 +37,20 @@
...
+#define LUA_BYPASS_DELAY_SEC 5
+#define LUA_BYPASS_FLASH_COUNT (1000 / LUA_BYPASS_FLASH_DELAY) * LUA_BYPASS_DELAY_SEC
```
stieg: So what would you ever do if the tick rate changes (Since I assume that is where you are getting the value of 1000).  Fix this plz.

stieg: Also that will break MK1 because tick rate there is 200

src/lua/luaTask.c
```diff
@@ -207,18 +213,35 @@ static bool _load_script(void)
         return true;
 }
 
+static bool user_bypass_requested(void)
+{
+        pr_info("lua: Checking for Lua runtime bypass request\r\n");
+        bool bypass = false;
+        LED_disable(ERROR_LED);
```
stieg: Not needed.

src/lua/luaTask.c
```diff
@@ -207,18 +213,35 @@ static bool _load_script(void)
...
+        for (size_t i = 0; i < LUA_BYPASS_FLASH_COUNT; i++) {
+            LED_toggle(ERROR_LED);
+            if (GPIO_is_button_pressed()) {
```
stieg: How long will this stay true if a user pushes the button?  Does it hold true while button is pressed, or is a bit set until cleared?

src/lua/luaTask.c
```diff
@@ -299,15 +322,16 @@ void run_lua_interactive_cmd(Serial *serial, const char* cmd)
...
 static void luaTask(void *params)
 {
         set_ontick_freq(DEFAULT_ONTICK_HZ);
         initialize_script();
 
-        /* If we are starting up after watchdog, don't init LUA */
-        if (watchdog_is_watchdog_reset()) {
-                pr_error("lua: Crash detected! Bypassing Lua script \r\n");
-                LED_enable(3);
+        bool should_bypass_lua = (watchdog_is_watchdog_reset() && user_bypass_requested());
```
stieg: `const bool`

src/watchdog/watchdog.c
```diff
@@ -1,7 +1,8 @@
 #include "watchdog.h"
 #include "watchdog_device.h"
+#include "printk.h"
```
stieg: Missing `#include <stdbool.h>`

src/lua/luaTask.c
brentpicasso: the function uses delayMs() which accounts for variances in tick rate.

src/lua/luaTask.c
brentpicasso: it is true only while the button is pressed.

stieg:
I can't say I am a big fan of this attack plan, but I don't have a better option other than to not disable LUA after a watchdog.  I doubt users will know how to use this (also update docs plz) or will think to use this, especially while on track.  Seems complicated, even from an advanced user's perspective.  Thoughts?

src/lua/luaTask.c
brentpicasso: No, it is specific to this feature / area of code.

src/watchdog/watchdog.c
brentpicasso: it's actually in watchdog.h, where it's first referenced.

src/lua/luaTask.c
brentpicasso: this was to guarantee consistent flashing (first toggle enables LED). the function won't know the state of the LED when entered.

brentpicasso:
Not much more than what's already described in Trello.

Lua script functionality is becoming a critical for users to enable core operation. While likely a very rare occurrence, a bad script that triggers that somehow segfaults will cause a repeated watchdog, effectively soft-bricking the unit, resulting in a huge support cost. 

At the same time, It would not be appropriate to assume all watchdogs are caused by a bad lua script.  This strikes a balance and allows recovery in these extreme cases.

src/lua/luaTask.c
brentpicasso: :+1:

src/lua/luaTask.c
brentpicasso: :+1:

src/lua/luaTask.c
stieg: Well what about how many times you go through that loop? Won't it be different on MK1 vs MK2?

stieg: Oh... I see.  1000 milliseconds to a second is where you get that value. Not 1000 ticks.  My mistake.

brentpicasso: :+1:

stieg: Also... when using defines everything should be in parens to avoid issues.  So this should read

```C
#define LUA_BYPASS_FLASH_COUNT ((1000 / LUA_BYPASS_FLASH_DELAY) * LUA_BYPASS_DELAY_SEC)
```
or
#define LUA_BYPASS_FLASH_COUNT (LUA_BYPASS_DELAY_SEC * 1000 / LUA_BYPASS_FLASH_DELAY)

src/watchdog/watchdog.c
stieg: Noted.  Its standard Linux practice to include everything you need at the file level as includes in other files can change.  This way removing an include in one file doesn't break the compilation of another.  Hence you should include it here too.

stieg:
I guess it will do for now. :shipit:

PR #89, opened 2015-09-23
Cleanup the sample rate check in loggerTaskEx.c, fix headers and incudes

stieg:
Does some minor cleanup of the logic that chooses when to send
a logging message by replacing an opaque check with a function
whose purpose is more clear.

Also fixes the header and includes per the CONTRIBUTING file
of our repo.

### DIFF
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -1,41 +1,44 @@
-/**
- * AutoSport Labs - Race Capture Firmware
+/*
+ * Race Capture Firmware
  *
- * Copyright (C) 2014 AutoSport Labs
+ * Copyright (C) 2015 Autosport Labs
  *
  * This file is part of the Race Capture firmware suite
  *
- * This is free software: you can redistribute it and/or modify it under the terms of the
- * GNU General Public License as published by the Free Software Foundation, either version 3 of the
- * License, or (at your option) any later version.
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
  *
- * This software is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
- * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  *
- * See the GNU General Public License for more details. You should have received a copy of the GNU
- * General Public License along with this code. If not, see <http://www.gnu.org/licenses/>.
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
  */
 
 #include "FreeRTOS.h"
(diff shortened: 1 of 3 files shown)


stieg:
paging @brentpicasso for a review.

brentpicasso:
Not sure what this review is for. What trello ticket does this relate to?

stieg:
Was related to the logger ticket.  Noticed some non-clear code while I was reviewing it.  So I opted to fix it.

brentpicasso:
Suggest moving this to 2.8.6 with more analysis. line 246 above that changes behavior and it may cause a regression.

src/logger/loggerTaskEx.c
```diff
@@ -243,8 +247,7 @@ void loggerTaskEx(void *params)
                 }
 
                 /* send the sample on to the telemetry task(s) */
-                if (sampledRate >= telemetrySampleRate ||
-                    currentTicks % telemetrySampleRate == 0)
+                if (!isHigherSampleRate(sampledRate, telemetrySampleRate))
```
brentpicasso: Changing this particular logic may cause a regression. Suggest moving to 2.8.6 where we can look at it more carefully.

stieg: Looked at it carefully and its part of the RC.  If it breaks we will know
it.  But could also be convinced to move it to 2.8.6.  Don't see much
difference IMHO.  I tested it locally and it worked fine.

On Thu, Sep 24, 2015 at 2:46 PM, Brent Picasso <notifications@github.com>
wrote:

> In src/logger/loggerTaskEx.c
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/89#discussion_r40363604>
> :
>
> > @@ -243,8 +247,7 @@ void loggerTaskEx(void *params)
> >                  }
> >
> >                  /* send the sample on to the telemetry task(s) */
> > -                if (sampledRate >= telemetrySampleRate ||
> > -                    currentTicks % telemetrySampleRate == 0)
> > +                if (!isHigherSampleRate(sampledRate, telemetrySampleRate))
>
> Changing this particular logic may cause a regression. Suggest moving to
> 2.8.6 where we can look at it more carefully.
>
> —
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/89/files#r40363604>
> .
>



--

brentpicasso: So my concern is you removed the modulus operator-  IIRC it's there to prevent a user from selecting a combination of sample rates in their logger config that would prevent telemetry from broadcasting.  My sense it may need more testing of different sample rate combinations and ensuring telemetry broadcast works.. That's the regression risk I'm seeing, and why I think it should be moved to 2.8.6.

brentpicasso: Also, I anticipated this was not part of RC b/c it's an un-merged PR.

stieg: I'm fine with moving it to 2.8.6, but since we are in this discussion I think we should finish it rather than start it again later.  Help me understand how that mod operator makes much of a difference.  When I was reading the code it seems extraneous.  Do you have an example where this isn't the case?  Possibly if all values are 0, but that is all I can see.

brentpicasso: Thanks. Please attach it to a r/2.8.6 branch and we can move forward with the 2.8.5 release. 

Need to do more analysis on the impact of this, but as I remember it's to prevent cases where we may not queue any telemetry messages if a user selects a certain combinations of sample rates.

stieg:
Closing because we are re-scoping this against 2.8.6.  Will re-open a new req.

PR #92, opened 2015-10-01
Fix MK1 Crash Bug, Cleanup Large Flashing Routines

stieg:
Flashing the tracks DB was causing MK1 to crash because it
was allocating a large chunk of RAM to do it.  We need to
disable LUA before we do this, else we will die due to RAM
starvation.  This fix does that and cleans up both of the
large flashing methods because they really could use it.  Now
both methods are much cleaner, up to CONTRIBUTING spec, and
generally make me happy.

### DIFF
--- a/SAM7s_base/capabilities.h
+++ b/SAM7s_base/capabilities.h
@@ -14,8 +14,17 @@
 #define MAX_VIRTUAL_CHANNELS	10
 #define LOGGER_MESSAGE_BUFFER_SIZE	5
 
+/*
+ * Adds additional memory saving behavior for low memory systems.
+ * These come at a cost of interruption of other services as needed
+ * to save RAM.  Usually quick interruptions, but interruptions
+ * none the less.
+ */
+#define RCP_LOW_MEM	1
+
 
 /* LUA Configuration */
+
 /*
  * What is the maximum length of the script that can be provided?
  * Must be divisible by 256.
--- a/include/lua/luaScript.h
+++ b/include/lua/luaScript.h
@@ -1,24 +1,49 @@
+/*
+ * Race Capture Pro Firmware
+ *
+ * Copyright (C) 2015 Autosport Labs
+ *
+ * This file is part of the Race Capture Pro fimrware suite
+ *
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
+ */
+
 #ifndef LUASCRIPT_H_
 #define LUASCRIPT_H_
 
-#include <stdint.h>
-#include "memory.h"
 #include "capabilities.h"
+#include "memory.h"
 
-#define SCRIPT_ADD_RESULT_OK  		1
-#define SCRIPT_ADD_RESULT_FAIL  	0
+#include <stdint.h>
 
-#define SCRIPT_ADD_MODE_IN_PROGRESS	1
-#define SCRIPT_ADD_MODE_COMPLETE 	2
+enum script_add_result {
+        SCRIPT_ADD_RESULT_FAIL = 0,
+        SCRIPT_ADD_RESULT_OK = 1,
+};
 
-#define MAGIC_NUMBER_SCRIPT_INIT 0xBADDECAF
+enum script_add_mode {
+        SCRIPT_ADD_MODE_IN_PROGRESS = 1,
+        SCRIPT_ADD_MODE_COMPLETE = 2,
+};
 
-#define SCRIPT_PAGE_SIZE 256
-#define MAX_SCRIPT_PAGES SCRIPT_MEMORY_LENGTH / SCRIPT_PAGE_SIZE
+#define MAGIC_NUMBER_SCRIPT_INIT	0xBADDECAF
+#define SCRIPT_PAGE_SIZE	256
+#define MAX_SCRIPT_PAGES	(SCRIPT_MEMORY_LENGTH / SCRIPT_PAGE_SIZE)
 
 typedef struct _ScriptConfig {
-    char script[SCRIPT_MEMORY_LENGTH - 4];
-    uint32_t magicInit;
+        char script[SCRIPT_MEMORY_LENGTH - 4];
+        uint32_t magicInit;
 } ScriptConfig;
 
 void initialize_script();
@@ -27,7 +52,8 @@ int flash_default_script();
 
 const char * getScript();
 
-int flashScriptPage(unsigned int page, const char *data, int mode);
+enum script_add_result flashScriptPage(unsigned int page, const char *data,
+                                       enum script_add_mode mode);
 
 void unescapeScript(char *data);
 
--- a/src/lua/luaScript.c
+++ b/src/lua/luaScript.c
@@ -120,33 +120,34 @@ enum script_add_result flashScriptPage(const unsigned int page,
                 /* Valid cases.  Carry on */
                 break;
         default:
-                pr_error_int_msg("Unknown script_add_mode: ", mode);
+                pr_error_int_msg("lua: Unknown script_add_mode: ", mode);
                 return SCRIPT_ADD_RESULT_FAIL;
         }
 
         static ScriptConfig *g_scriptBuffer;
-        if (g_scriptBuffer == NULL) {
+        if (NULL == g_scriptBuffer) {
                 terminate_lua();
 
-                pr_info("Allocating new script buffer\r\n");
+                pr_debug("lua: Allocating new script buffer\r\n");
                 g_scriptBuffer =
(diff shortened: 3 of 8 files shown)


include/lua/luaScript.h
```diff
@@ -1,24 +1,49 @@
...
-#define SCRIPT_ADD_MODE_IN_PROGRESS	1
-#define SCRIPT_ADD_MODE_COMPLETE 	2
+enum script_add_result {
+        SCRIPT_ADD_RESULT_FAIL = 0,
+        SCRIPT_ADD_RESULT_OK = 1,
+};
 
-#define MAGIC_NUMBER_SCRIPT_INIT 0xBADDECAF
+enum script_add_mode {
+        SCRIPT_ADD_MODE_IN_PROGRESS = 1,
```
brentpicasso: inconsistent with with track_add_mode...  enum.  (missing 0 )

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
+        }
 
-    if (page < MAX_SCRIPT_PAGES) {
-        if (mode == SCRIPT_ADD_MODE_IN_PROGRESS || mode == SCRIPT_ADD_MODE_COMPLETE) {
-            if (g_scriptBuffer == NULL) {
-                g_scriptBuffer = (ScriptConfig *)portMalloc(sizeof(ScriptConfig));
-                memcpy((void *)g_scriptBuffer, (void *)&g_scriptConfig, sizeof(ScriptConfig));
-            }
+        static ScriptConfig *g_scriptBuffer;
+        if (g_scriptBuffer == NULL) {
```
brentpicasso: suggest NULL == g_scriptBuffer

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
+                memcpy((void *)g_scriptBuffer, (void *)&g_scriptConfig,
+                       sizeof(ScriptConfig));
         }
-    } else {
-        pr_error("lua: invalid script index\r\n");
-        result = SCRIPT_ADD_RESULT_FAIL;
-    }
-    return result;
+
+        if (g_scriptBuffer == NULL) {
```
brentpicasso: ditto

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
+                pr_error("Failed to allocate memory for script buffer.\r\n");
+                return SCRIPT_ADD_RESULT_FAIL;
+        }
+
+        char *pageToAdd = g_scriptBuffer->script + page * SCRIPT_PAGE_SIZE;
+        strncpy(pageToAdd, data, SCRIPT_PAGE_SIZE);
+
+        if (mode == SCRIPT_ADD_MODE_IN_PROGRESS)
```
brentpicasso: applies throughout

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
+                       sizeof(ScriptConfig));
         }
-    } else {
-        pr_error("lua: invalid script index\r\n");
-        result = SCRIPT_ADD_RESULT_FAIL;
-    }
-    return result;
+
+        if (g_scriptBuffer == NULL) {
+                pr_error("Failed to allocate memory for script buffer.\r\n");
```
brentpicasso: should prefix with lua:

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
+                return SCRIPT_ADD_RESULT_FAIL;
+        }
+
+        char *pageToAdd = g_scriptBuffer->script + page * SCRIPT_PAGE_SIZE;
+        strncpy(pageToAdd, data, SCRIPT_PAGE_SIZE);
+
+        if (mode == SCRIPT_ADD_MODE_IN_PROGRESS)
+                return SCRIPT_ADD_RESULT_OK;
+
+        pr_info("Completed updating LUA. Flashing... ");
```
brentpicasso: should prefix with lua:

suggest:   "lua: completed updating script. Flashing... "

src/tracks/tracks.c
```diff
@@ -42,44 +62,60 @@ const Tracks * get_tracks()
...
+                /* Valid cases.  Carry on */
+                break;
+        default:
+                pr_error_int_msg("Unknown track_add_mode: ", mode);
+                return TRACK_ADD_RESULT_FAIL;
+        }
+
+        static Tracks *g_tracksBuffer;
+        if (g_tracksBuffer == NULL) {
+                terminate_lua();
```
brentpicasso: We need to not have this happen for MK2. It will have surprise side effects for users. it's an extreme measure to make MK1 work reliably.

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
-     * RAM since our flashing operation is a heavy bugger
-     */
-    terminate_lua();
+        switch (mode) {
+        case SCRIPT_ADD_MODE_IN_PROGRESS:
+        case SCRIPT_ADD_MODE_COMPLETE:
+                /* Valid cases.  Carry on */
+                break;
+        default:
+                pr_error_int_msg("Unknown script_add_mode: ", mode);
```
brentpicasso: missing lua: prefix

src/lua/luaScript.c
```diff
@@ -108,47 +105,60 @@ void unescapeScript(char *data)
...
-                        result = SCRIPT_ADD_RESULT_FAIL;
-                    }
-                    portFree(g_scriptBuffer);
-                    g_scriptBuffer = NULL;
-                }
-            } else {
-                pr_error("lua: script buffer alloc fail\r\n");
-                result = SCRIPT_ADD_RESULT_FAIL;
-            }
+                pr_info("Allocating new script buffer\r\n");
```
brentpicasso: this should be debug, trace or removed.

include/lua/luaScript.h
stieg: Yeah... was thinking about that.  I'll remove the 0 on the other.

src/lua/luaScript.c
stieg: Got rid of it because of the upcoming logging changes.  I'll add it back in

src/lua/luaScript.c
stieg: Will do.

src/lua/luaScript.c
stieg: debug it is.

src/tracks/tracks.c
stieg: A little surprised you would bring this up now.  What side effect are you concerned about specifically?  Its really not that extreme.

brentpicasso: with this change writing tracks would shut down and re-start lua.  If there is stateful information in the user's Lua script then it would get reset. This would be a surprise to the user; to them, updating tracks has nothing to do with Lua scripting. IOW, it violates the principal of least surprise.

stieg: Or are you just concerned that restarting LUA during track map flash would confuse customers?

brentpicasso: yes. 

I think it should only be done for MK1 - possibly with a #define and a comment to remove it from MK2 code after we fork.

stieg: I see your point but I'm not convinced.  When these values change its because someone is programming their device.  This isn't going to change while they are on track or running a session right?  Plus we need to establish the mentality that LUA can and will restart at any time.  What state would they have in there that couldn't restart a LUA restart (other than say timing info)?

Regardless I'll put in a patch to address it.

stieg: Consider this case: LUA is hogging a buttload of ram and people try to flash tracks.  If lua doesn't go down then the program crashes or has a memory leak.  That is worse than pulling LUA to ensure things work no?

brentpicasso: We're now limiting Lua RAM usage so this shouldn't be a problem for MK2.

stieg: Ok.  I'll buy that argument.

On Fri, Oct 2, 2015 at 1:28 PM, Brent Picasso <notifications@github.com>
wrote:

> In src/tracks/tracks.c
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/92#discussion_r41052211>
> :
>
> > +                return TRACK_ADD_RESULT_FAIL;
> > +        }
> > +
> > +        switch (mode) {
> > +        case TRACK_ADD_MODE_IN_PROGRESS:
> > +        case TRACK_ADD_MODE_COMPLETE:
> > +                /* Valid cases.  Carry on */
> > +                break;
> > +        default:
> > +                pr_error_int_msg("Unknown track_add_mode: ", mode);
> > +                return TRACK_ADD_RESULT_FAIL;
> > +        }
> > +
> > +        static Tracks *g_tracksBuffer;
> > +        if (g_tracksBuffer == NULL) {
> > +                terminate_lua();
>
> We're now limiting Lua RAM usage so this shouldn't be a problem for MK2.
>
> —
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/92/files#r41052211>
> .
>



--

SAM7s_base/capabilities.h
```diff
@@ -14,8 +14,17 @@
...
+ * Adds additional memory saving behavior for low memory systems.
+ * These come at a cost of interruption of other services as needed
+ * to save RAM.  Usually quick interruptions, but interruptions
+ * none the less.
+ */
+#define RCP_LOW_MEM	1
+
 
 /* LUA Configuration */
+
```
brentpicasso: extra whitespace?

brentpicasso:
just one tiny thing. assuming you tested this on both platforms - if so, :shaved_ice:

PR #93, opened 2015-10-02
corrected logic for triggering track detection & indicating fixed track mode.

brentpicasso:
trigger a track re-detection if track config changes; correctly selec fixed track mode when appropriate


stieg:
:shipit:

PR #94, opened 2015-10-02
Fix sector enabled logic

stieg:
There was a hilarious bug in the firmware where we were thinking
sectors were enabled when in fact they were not.  This was because
the method that gets a sector at an index returns the finish line
if it has no more sectors (because that is the next logical sector
boundary).  However if no sectors are defined, then sector 0 will
return the finish line, and in our logic we were not taking that
behavior into account.  This fixes that issue.


brentpicasso:
:sheep:

brentpicasso:
oh, add to CHANGELOG

PR #95, opened 2015-10-09
Fix mk1 bt overflow

stieg:
Fixes overflow issues in this unit by

* Making BT task a higher priority
* Increasing BT queue tx/rx queue size

Also adds a little bit of debugging.

### DIFF
--- a/SAM7s_base/usart_at91/usart_device_at91.c
+++ b/SAM7s_base/usart_at91/usart_device_at91.c
@@ -31,26 +31,26 @@
 //         Headers
 //------------------------------------------------------------------------------
 
-#include "usart_device.h"
 #include "FreeRTOS.h"
-#include "task.h"
-#include "queue.h"
-#include "printk.h"
 #include "board.h"
+#include "printk.h"
+#include "queue.h"
+#include "task.h"
+#include "usart_device.h"
+#include "usart_device_at91.h"
 
 #define GPS_BAUDRATE 38400
 #define TELEMETRY_BAUDRATE 115200
-
 #define USART_INTERRUPT_LEVEL 5
-#define USART_QUEUE_LENGTH 300
 
 void usart0_irq_handler (void);
 void usart1_irq_handler (void);
 
-xQueueHandle xUsart0Tx;
-xQueueHandle xUsart0Rx;
-xQueueHandle xUsart1Tx;
-xQueueHandle xUsart1Rx;
+static struct txrx_queue usart_txrx_queues[2];
+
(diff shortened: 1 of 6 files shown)


brentpicasso:
From the description it should look like a few numbers adjusted but actually  includes a significant last minute refactoring that can arguably saved for a future release. Thoughts?

SAM7s_base/usart_at91/usart_device_at91.c
```diff
@@ -141,9 +134,17 @@ static unsigned int createUartMode(unsigned int bits, unsigned int parity, unsig
...
     AT91F_PMC_EnablePeriphClock( AT91C_BASE_PMC,(1 << AT91C_ID_US0) | (1 << AT91C_ID_US1));
-    if (!initQueues()) return 0;
+
+    /* Init device 0 (Telemetry) */
+    if (!init_txrx_queue(0, 512, 512))
+            return 0;
     usart_device_init_0(8, 0, 1, TELEMETRY_BAUDRATE);
+
+    /* Init device 1 (GPS) */
+    if (!init_txrx_queue(1, 512, 128))
```
brentpicasso: Unrelated change with significant risk of GPS regression.  GPS Rx Buffer was reduced from 300 to 128; Tx Buffer increased from 300 512. MK1 treats GPS as a read only device; does not communicate with GPS at all.

brentpicasso:
Rejecting this PR due to inclusion of unrelated changes, including at least one significant regression risk (GPS)

stieg:
The re factor I did because it would have been an ugly hack to not do it.  If you look at the meat of the change most of it was eliminating duplicate code and consolidating code down.  So low risk there.  I will agree with you RE the GPS buffer settings.  I can set them back to 300/300 if you like, or do something like 512/300 (Rx/Tx) so that the chance of regression there is eliminated.  I don't feel like the code is high risk because if something was wrong it would be immediately obvious.

stieg:
Fine.  I'll rebase this against 2.8.6.

brentpicasso:
The refactoring work was sound in concept, definitely. But in principal, would be safest to put against 2.8.6.

Since MK1 does not communicate with GPS module *at all* it would be best to give it a generous Rx queue (say, 512) and a minimal Tx queue (say, 32).

stieg:
> Since MK1 does not communicate with GPS module at all it would be best to give it a generous Rx queue (say, 512) and a minimal Tx queue (say, 32). 

That is exactly what I did.  And I thought it sends the GPS some data in the beginning to set NMEA parameters and data rate etc.

PR #96, opened 2015-10-12
Change the MK1 Telemetry Queue size to 512/512.

stieg:
This is to prevent buffer overflow during high GPS usage scenarios.


brentpicasso:
:shipit:

PR #98, opened 2015-10-19
P/stieg/code cleanup

stieg:
Some code cleanup that was deferred from the 2.8.5 release.  Of note the change that is in loggerTasxEx.c.  @brentpicasso seems to be concerned about it but I was unable to figure out the reason based on the code.  The second conditional was redundant in my investigation.  Hence why I made the change.

### DIFF
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -1,9 +1,9 @@
 /*
- * Race Capture Pro Firmware
+ * Race Capture Firmware
  *
  * Copyright (C) 2015 Autosport Labs
  *
- * This file is part of the Race Capture Pro fimrware suite
+ * This file is part of the Race Capture fimrware suite
  *
  * This is free software: you can redistribute it and/or modify it
  * under the terms of the GNU General Public License as published by
@@ -236,7 +236,8 @@ void loggerTaskEx(void *params)
                  * We only log to file if the user has manually pushed the
                  * logging button.
                  */
-                if (is_logging && sampledRate >= loggingSampleRate) {
+                if (!isHigherSampleRate(sampledRate, loggingSampleRate)
+                    && is_logging) {
                         /* XXX Move this to file writer? */
                         const portBASE_TYPE res = queue_logfile_record(&msg);
                         const logging_status_t ls = pdTRUE == res ?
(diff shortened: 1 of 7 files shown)


src/logger/loggerTaskEx.c
```diff
@@ -246,8 +247,7 @@ void loggerTaskEx(void *params)
                 }
 
                 /* send the sample on to the telemetry task(s) */
-                if (sampledRate >= telemetrySampleRate ||
-                    currentTicks % telemetrySampleRate == 0)
+                if (!isHigherSampleRate(sampledRate, telemetrySampleRate))
```
brentpicasso: we need this line: currentTicks % telemetrySampleRate == 0

otherwise if they have a config that coincidentally does not include the telemetry sample rate we may not queue any telemetry records.  Did you test this for regressions under varying conditions?

stieg: > if they have a config that coincidentally does not include the telemetry sample rate

How is that even possible?  Where would a user set the telemetry sample rate exactly?  I certainly don't see such a location in the App.  One plausible cause is if _all_ channel sample rates are disabled, but even then you would have the interval and UTC time channel (which can not be disabled) so that is really not a valid scenario.

I remember this _was_ an issue long ago, but I'm not sure if it is one now.  Thoughts?

brentpicasso: Did you test this for regressions under varying conditions?

brentpicasso: Example: configuring RCP for different sample rates (example, setting everything to 1Hz) and verifying bluetooth telemetry or cellular telemetry still works?

brentpicasso: If you notice, the loggingSampleRate is a function of determining the highest sample rate present in the config.

Alternatively, the telemetrySampleRate is fixed regardless of the config. IIRC there's a case where it's possible that a certain configuration can prevent telemetry samples from being sent at all. That's why that extra code exists, to ensure that we at least send telemetry on those tick intervals.

stieg:
Tested and indeed it was an issue with disabled channels.  So I fixed the code so that it is now physically impossible to disable all of the channels (interval and UTC time will always be there).

stieg:
@brentpicasso Have a look at this and tell me what you think.  Tested it and with that patch I was never able to disable telemetry.  So I think we have a good solution here.

PR #339, opened 2015-11-02
Code cleanup on the API methods for loggin

stieg:
Cleans up the API methods for logging and makes it so that users can never fully disable telemetry.  This in turn allows us to eliminate some hacks.


stieg:
@brentpicasso Review me plz

Issue #341, opened 2015-11-08
Change telemetry hostname to telemetry.race-capture.com


ryandoherty:
In order to properly handle telemetry data sent to race-capture.com, the domain name use should be changed to telemetry.race-capture.com

brentpicasso:
What are the thoughts of cut-off period for people running older firmware?

brentpicasso:
Also, will there be a way to tell how many people are hitting the old hostname so we can have a clue of when to cut people off?

PR #342, opened 2015-11-08
Update telemetry address

stieg:
The addition of cloud front means we need to use a better qualifier of a
host name as the telemetry destination so as not to interfere with the
CF operations.

telemetry.race-capture.com will always bypass CF and go straight to the
recording backend.


stieg:
Fixes https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/341

brentpicasso:
please target 2.9.0 milestone

PR #343, opened 2015-11-11
Gyro sensitivity training (ie improve the sensitivity of the gyro)

stieg:
The old config used a gyro sensitivity value for a rotational angle
sensitive to up to 2000 degrees per second.  If a race car is ever
spinning that fast.... dang.  This change lowers that sensitivity.
The theory goes that by doing this, we get more bits for precision,
which will improve the sensitivity of the device.

Sensitivity is key for accurate slip_angle readings.


brentpicasso:
This is great. clean, concise changeset. :shipit:

Issue #209, opened 2015-10-26
Bluetooth initialization problem


ryandoherty:
Bluetooh initializes too fast (race condition before BT module is ready)

stieg:
@ryandoherty Huh... Can you elaborate more on this?

brentpicasso:
I think I opened this, and corresponds to the delay before we try to talk
to the module. We need to wait some amount of time before we can
communicate with module.

The delay is not specified and the module does not output a startup message
AFAIK so we have to blindly wait.

Brent Picasso
Autosport Labs
Technology for Race and Street

stieg:
Ahh that is most certainly a part of this fix (it was already there, but added some pauses when we change the serial baud).

stieg:
Closing because merged

Issue #123, opened 2015-10-26
BT disconnects with 2.8.0 and latest prod Android app


ryandoherty:
After 5 seconds of  RCP (mk2 and 2.8.0) sending data over BT it stops sending data.

Most likely due to the new HB logic.

Wireless: timeout
BT: Probing baud115200
BT: cmd: AT
ATBT: cmd rsp: 
BT: nomatch
BT: Provisionfail
BT: Probing baud9600
BT: cmd: AT
ATBT: cmd rsp: 
BT: nomatch
BT: Provisionfail
BT: Probin

stieg:
Looks like the serial altering issue I just fixed.  See https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/346

brentpicasso:
Double check this isn't related to RCP trying to provision the bt module
while android is already connected. If android is already connected, the
module will not respond to AT commands

Brent Picasso
Autosport Labs
Technology for Race and Street

stieg:
I hit this issue while implementing my changes.  Issue was that we were not
giving the serial port enough time to synchronize with the BT device before
we would send the AT command to check if we could communicate with the
hardware.  It was intermittent.  When I put a 15ms delay after changing the
serial port baud rate, the issue completely disappeared.

On Mon, Nov 16, 2015 at 5:00 PM, Brent Picasso <notifications@github.com>
wrote:

> Double check this isn't related to RCP trying to provision the bt module
> while android is already connected. If android is already connected, the
> module will not respond to AT commands
>
> Brent Picasso
> Autosport Labs
> Technology for Race and Street
> On Nov 16, 2015 2:46 PM, "Andrew Stiegmann" <notifications@github.com>
> wrote:
>
> > Looks like the serial altering issue I just fixed. See #346
> > <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/346>
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <
> https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/123#issuecomment-157197439
> >
> > .
>
> >
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/123#issuecomment-157202220>
> .
>



--

Issue #344, opened 2015-11-12
Adjust BT name of RCP units to include last 4 digits of serial


stieg:
Needed because when multiple RCP units are around and you try to pair, it can be very difficult to tell which unit you are looking for.

brentpicasso:
We have APIs for setting the BT name, not yet exposed in the app.

If this feature is implemented, it's important to ensure this is only set
when restoring factory defaults.

On Thu, Nov 12, 2015 at 6:11 AM, Andrew Stiegmann <notifications@github.com>
wrote:

> Needed because when multiple RCP units are around and you try to pair, it
> can be very difficult to tell which unit you are looking for.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/344>.
>



-- 
Brent Picasso
*Autosport Labs <http://www.autosportlabs.com>*
*Technology for Race and Street*

stieg:
Agreed.  It would only apply if a factory reset happens.  That is how I will program the patch.

stieg:
Hrmm... interesting issue here.  The BT device name is set when the unit boots.  So if the unit boots and the BT name has been changed, it will be reflected during that boot.  That seems logical but maybe you may disagree @brentpicasso

stieg:
Closing because merged.

PR #346, opened 2015-11-16
Fix up BT on RCP

stieg:
This patch set improves our BT stack by

* Making BT device names unique by appending last 4 digits of device serial.
* Fixes up issues observed while programming BT unit
* Adds support for programming BT PIN
* Fixes various minor bugs.

This is safe to merge against 2.8.6 as this change does not adjust the LoggerConfig struct size in any way.  The BT name change will only apply after the device undergoes a reset.

### DIFF
--- a/src/devices/bluetooth.c
+++ b/src/devices/bluetooth.c
@@ -1,23 +1,26 @@
-/**
- * AutoSport Labs - Race Capture Firmware
+/*
+ * Race Capture Pro Firmware
  *
- * Copyright (C) 2014 AutoSport Labs
+ * Copyright (C) 2015 Autosport Labs
  *
- * This file is part of the Race Capture firmware suite
+ * This file is part of the Race Capture Pro fimrware suite
  *
- * This is free software: you can redistribute it and/or modify it under the terms of the
- * GNU General Public License as published by the Free Software Foundation, either version 3 of the
- * License, or (at your option) any later version.
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
  *
- * This software is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
- * without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
  *
- * See the GNU General Public License for more details. You should have received a copy of the GNU
- * General Public License along with this code. If not, see <http://www.gnu.org/licenses/>.
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "bluetooth.h"
 #include "FreeRTOS.h"
+#include "bluetooth.h"
 #include "loggerConfig.h"
 #include "mod_string.h"
 #include "printk.h"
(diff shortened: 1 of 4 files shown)


stieg:
This addresses issue #344

stieg:
This also likely fixes issue #123

stieg:
Likely also fixed #209 per conversation with @brentpicasso

src/devices/bluetooth.c
```diff
@@ -1,169 +1,230 @@
...
+                bt_get_version(config) &&
+                bt_set_name(config, deviceName) &&
+                bt_set_pin(config, pin) &&
+                bt_set_baud(config, targetBaud);
+
+        if (status) {
+                pr_info("BT: Init complete\r\n");
+                g_bluetooth_status = BT_STATUS_PROVISIONED;
+        } else {
+                pr_info("BT: Failed to provision module.\r\n");
```
brentpicasso: if app is already connected via BT it will not respond to AT commands and the provisioning will fail, falling to line 220. Line 158 provided a hint that this might be the case, saying we may already be connected. should update this line to be consistent with that.

stieg: Interesting.  I'll re-add that info then.

src/devices/bluetooth.c
```diff
@@ -1,169 +1,232 @@
...
+        flushBt(config);
+        putsBt(config, cmd);
+        readBtWait(config, wait);
+
+        const bool res = 0 == strncmp(config->buffer, rsp, strlen(rsp));
+
+        pr_debug_str_msg("BT: wanted rsp: ", rsp);
+        pr_debug_str_msg("BT: actual rsp: ", config->buffer);
+        pr_debug_int_msg("BT: matched: ", res);
+
```
brentpicasso: This should be trace, or better, simply removed, seems like development code.

src/devices/bluetooth.c
```diff
@@ -1,169 +1,232 @@
...
+                        break;
+        }
+
+        /* Check that we didn't fail to find a workable rate */
+        if (i == sizeof(rates)) {
+                pr_info("BT: Unable to communicate with device.\r\n");
+                return false;
+        }
+
+        pr_info_int_msg("BT: Talking with BT device.  Baud: ", rates[i]);
```
brentpicasso: suggest:

BT: Detected at Baud:

src/devices/bluetooth.c
```diff
@@ -1,169 +1,232 @@
...
+        strlcpy(buf + 6, pin_str, BT_MAX_PIN_LEN + 1);
+
+        return sendBtCommandWaitResponse(config, buf, "OKsetPIN",
+                                         COMMAND_WAIT);
+}
+
+
+static bool bt_find_working_baud(DeviceConfig *config, const int targetBaud)
+{
+        pr_info("BT: Searching for working baud rate...\r\n");
```
brentpicasso: Suggest:

BT: Detecting baud rate...

src/devices/bluetooth.c
```diff
@@ -1,169 +1,232 @@
...
+        const int rates[] = {targetBaud, 230400, 115200, 57600, 9600};
+        size_t i = 0;
+        for (; i < sizeof(rates); ++i) {
+                if(set_check_bt_serial_baud(config, rates[i]))
+                        break;
+        }
+
+        /* Check that we didn't fail to find a workable rate */
+        if (i == sizeof(rates)) {
+                pr_info("BT: Unable to communicate with device.\r\n");
```
brentpicasso: Suggest: 

BT: Could not detect on known baud rates

src/devices/bluetooth.c
```diff
@@ -1,169 +1,232 @@
...
+                bt_set_baud(config, targetBaud);
+
+        if (status) {
+                pr_info("BT: Init complete\r\n");
+                g_bluetooth_status = BT_STATUS_PROVISIONED;
+        } else {
+                pr_info("BT: Failed to provision module.  This may "
+                        "be caused by a device connecting to the BT "
+                        "module.\r\n");
+                g_bluetooth_status = BT_STATUS_ERROR;
```
brentpicasso: Suggest:

BT: Failed to provision module: A client may already be connected.

brentpicasso:
This is a great update. Some minor suggestions to messages.

I'm not able to test directly, but if it works under varying conditions please go ahead and merge after applying additional feedback.

stieg:
Feedback applied.  I tested it with my local device and things seemed pretty solid.  The only thing I haven't tested it with is an unprovisioned BT unit.

PR #347, opened 2015-11-17
Improve MK2 Compilation quality

stieg:
Fixes compilation warnings and ensures that our compiler will no longer successfully compile with warnings.


src/logger/loggerApi.c
```diff
@@ -1642,7 +1642,6 @@ int api_setScript(Serial *serial, const jsmntok_t *json)
                flashScriptPage(page, script, (enum script_add_mode) mode);
 
             rc = flashResult == 1 ? API_SUCCESS : API_ERROR_SEVERE;
-            reload_script = rc == API_SUCCESS && mode == SCRIPT_ADD_MODE_COMPLETE;
```
brentpicasso: What is this in relation to the PR?

brentpicasso: NVM, I see i now.

PR #350, opened 2015-12-14
Avoiding conditional directives that split up parts of statements.

lzmths:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

    https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
    https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/stm32_base/hal/fat_sd_stm32/fatfs/ff.c
+++ b/stm32_base/hal/fat_sd_stm32/fatfs/ff.c
@@ -2877,12 +2877,15 @@ FRESULT f_close (
 )
 {
     FRESULT res;
-
+    int res_test;
 
 #if !_FS_READONLY
     res = f_sync(fp);					/* Flush cached data */
-    if (res == FR_OK)
+    res_test = (res == FR_OK);
+#else
+    res_test = 1;
 #endif
+    if (res_test)
     {
         res = validate(fp);				/* Lock volume */
         if (res == FR_OK) {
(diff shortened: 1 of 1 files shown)


stm32_base/hal/fat_sd_stm32/fatfs/ff.c
```diff
@@ -2877,12 +2877,15 @@ FRESULT f_close (
...
     FRESULT res;
-
+    int res_test;
 
 #if !_FS_READONLY
     res = f_sync(fp);					/* Flush cached data */
-    if (res == FR_OK)
+    res_test = (res == FR_OK);
+#else
+    res_test = 1;
```
stieg: Instead of having an `#else` here why not simply initialize res_test to = 1.  Saves you a couple lines of code.

stieg:
Thanks for the PR.  So this code is not actually maintained. by us directly.  Rather this code is from http://elm-chan.org/fsw/ff/00index_e.html that we simply have in the project.  We can fix things here but it won't make any difference in how the program operates.  I suggest you port your patch to the upstream project there where it will have more effect.

lzmths:
I understand.

RomeroMalaquias:
Hello @stieg , i'm from the same research team of lzmths, and i need to ask if you prefer this code before or after the commit. What do you think about this kind of change?
Thanks in advance.

RomeroMalaquias:
I mean, would you accept if it was not for the fact we submitted the pull request to a code written by third parties?

PR #355, opened 2015-12-27
Fix telemetry destination

stieg:
Changes the destination of telemetry data to `telemetry.race-capture.com`.  This is so that we may better work with Cloudflare.

This resolves issue https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/341


brentpicasso:
:shipit:

PR #356, opened 2015-12-28
Remove MK1 from the mainline build tree

stieg:
Now that MK1 has been forked to its legacy branch this change goes
through and removes all references to MK1 in mainline.  Tested
MK2, unit tests and do_release.sh.


brentpicasso:
Looks good. We might need further cleanup of the Makefile for the lua folder but we can deal with that later.

Issue #340, opened 2015-11-04
Fix/Add `extern "C"` headers for C files used in CPP unit tests


stieg:
To avoid issues with how the CPP compiler mangles method names during testing and compilation of C code, as well as to allow the use of the C compiler during unit test builds, we need to add these headers so that cpp code will be able to appropriately find these methods without symbol lookup errors during link time.

PR #361, opened 2016-01-05
Add CPP guards to C Headers.  Use GCC only for C files.

stieg:
This monster of a change adds some simple things that will allow us to
better test our code as well as cleans up and makes our code more
consistent.

First this change adds the standard C++ guards around our headers in C
code.  This prevents the C++ compiler from mangling the symbols when
the code gets built.

Next, this change alters the compilation process in the test platform
so that we use a C compiler (gcc) instead of a C++ compiler (g++) to
compile C sources.  This ensures that our code is always compiled as C
code, removing unnecessary symbol mangling and side-stepping various
warnings/errors that can occur when compiling C code in C++ land.

Finally this change also fixes up several headers in various source
files to be consistent with our new style.

Issues: #340


stieg:
Since this change is mostly boilerplate, I'm going to forgo a review and simply merge.  Have a look if you like, but I'm sure you will get bored of the repetition quickly.  Resolves #340 as well.

Issue #362, opened 2016-01-06
Adapt lap_stats getAtStartFinish method to work with both circuit and stage tracks


stieg:
Currently the `getAtStartFinish` method returns true IFF the user is at either (or both) the start and finish.  However this is not super useful since you don't know which one they are at.  Also there is a chance that the LUA script could miss the event since GPS and LUA invocation rates are independent of one another.

Ideally a callback is registered so that when these events happen call backs fire and update state appropriately.

stieg:
I'm not going to fix this since I don't see much use for it.  Instead if we want to do this right we should define methods that get called if the users is at start or finish and invoke them in Lua so that its not missed when they were actually there.

Issue #359, opened 2016-01-02
Reset distance channel to 0 when crossing the start


ryandoherty:
RaceCapture/Live is seeing lap data where a new lap starts, but the Distance channel's value starts at values much greater than 0 (0.2, 0.5, etc). This is most likely due to having separate start/finish points and RCP is 'counting' the distance between finish => start, making the Distance channel start at a high value.

This breaks a lot of functionality in other systems and also makes it impossible to compare lap data over distance.

ryandoherty:
Nominating this to be fixed/investigated for the next firmware release.

stieg:
Interesting.  I'll have a look.

ryandoherty:
Tough part is I saw this for telemetry data from a RC => phone => RCL connection, so no raw logs. Need to get the app recording data locally too.

stieg:
So far investigation has found minor issues in how start logic is handled.  A patch is in progress but not known if it will directly solve this issue.  We must be able to support us starting with a distance value > 0 (but not one that is .5 miles) since the implementation of the launch control feature requires us to start a lap with a non-zero distance (because launch control only knows you have started your lap _after_ it has detected a successful launch).

stieg:
Tested and passed as expected.

stieg:
ok to test

stieg:
test this please

stieg:
Need to port fix to 2.8.7

stieg:
PR: https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/393

PR #363, opened 2016-01-06
P/stieg/lap stats fixes

stieg:
This change fixes a few minor issues in our lap_stats code including:

* Ensuring that we can start a lap if RCP is powered up at the start line
* Fixing getAtStartFinish so that it works again (though further work is needed as outlined by #362)
* Fixing a Lap Start triggering issue where we were triggering when we shouldn't have been #359

It also drastically improves the unit tests for this unit of code to help reduce the chance of regressions.


stieg:
ok to build

stieg:
ok to build

stieg:
test this please

stieg:
test this please

stieg:
ok to test

stieg:
test this please

stieg:
test this please

stieg:
test this please

stieg:
test this please

stieg:
test this please

stieg:
test this please

src/lap_stats/lap_stats.c
```diff
@@ -297,86 +313,99 @@ static void sectorBoundaryEvent(const GpsSnapshot *gpsSnapshot)
...
 }
 
+static void process_start_normal(const GpsSnapshot *gpsSnapshot)
+{
+        const GeoPoint point = gpsSnapshot->sample.point;
+        if (!gc_isPointInGeoCircle(&point, g_geo_circles.start))
+                return;
+
+        const tiny_millis_t time = gpsSnapshot->deltaFirstFix;
+        lap_started_event(time, &point, 0);
```
brentpicasso: this function name should contain a verb.  e.g. if you want a event type function name  on_lap_started would be good.

src/lap_stats/lap_stats.c
```diff
@@ -297,86 +313,99 @@ static void sectorBoundaryEvent(const GpsSnapshot *gpsSnapshot)
...
+                return;
+
+        /*
+         * If here, then launch control has determined that the lap
+         * has begun.  Launch control unfortunately causes a delay in
+         * reporting since it has to detect a launch before it can
+         * report it.  Hence we must retroactively calculate the
+         * distance we have traveled and the time elapsed since we
+         * launched from start as it is before this point in time.
+         */
```
brentpicasso: So in other launch control designs, we would have a circular buffer that would contain all of the channel samples so that when we've detected we've launched, we can include all of that previous data back to the point where we actually launched.  We don't use a big circular buffer right now, so we must be using an alternate approach.

Here, it looks like you're just calculating the distance from where we launched from - and the "lap" starts with a pre-calculated distance value?

brentpicasso:
I'm looking at this and I'm wondering if we need to separate lap-timing engines - one to handle circuit, one to handle stage with launch control.  It would reduce the cyclomatic complexity by not having the switching logic in a codebase that's trying to do both, and might be easier to read and understand.

src/lap_stats/lap_stats.c
stieg: ???  This function is named this way so that we process start logic without the involvement of launch control.  Just because we are processing start logic doesn't imply that a start has happened (that is what the `lap_started_event` is for).  I could rename it to `process_start_no_launch_control` if that would make it clearer.

src/lap_stats/lap_stats.c
stieg: > So in other launch control designs, we would have a circular buffer that would contain all of the channel samples so that when we've detected we've launched, we can include all of that previous data back to the point where we actually launched. We don't use a big circular buffer right now, so we must be using an alternate approach.

Correct.  Using a big circular buffer is not a good design in my opinion because you have to detect a launch before you run out of buffer space.  If that doesn't happen, you get a buffer overrun and lose data.  My design on launch control works on the principle of keeping the launch control zone small and figuring out the point where you first moved above a threshold speed.  So long as the launch control circle is small, this distance should be fairly accurate.

stieg:
> It would reduce the cyclomatic complexity by not having the switching logic in a codebase that's trying to do both, and might be easier to read and understand.

I do not agree with this.  The fundamental difference between a circuit track and a stage track is very minor.  When you break it down a circuit track is simply a stage track with a special condition at the end  (the finish point is the same as the start point).  I don't see much of any gains that would be had by doing this split but rather see tons of duplicated logic and complexity in our sampling code if we were to split them out into separate units.  And its not just a simple split between circuit and stage.  In example, you want launch control on the 0th lap of a circuit race if your race starts from a standstill (think many kart race scenarios).

I understand your concerns about the logic being a bit confusing.  I didn't put time into simplifying the code at all with this change because it was outside the scope of my goal.  Perhaps we should file a ticket to simplify this code instead?

src/lap_stats/lap_stats.c
brentpicasso: :+1:

src/lap_stats/lap_stats.c
brentpicasso: One arguable goal could be that we want all samples from all channels from the point when we launched.  In the case here, our "lap" or "run" only starts when we've tripped the threshold? So some data is dropped in theory.   Maybe that's Ok, if it's minimal amounts of data.  What's the estimate of the amount of lost data from the point when the person actually launches to the point where the firmware thinks it's happened?

brentpicasso:
Sounds good. A ticket to refactor / simplify the code would be good I think, and then additional work may come from that based on how it goes. Thanks.

src/lap_stats/lap_stats.c
```diff
@@ -1,24 +1,46 @@
...
 #include "tracks.h"
-#include "test.h"
 
 #include <stdint.h>
 
-#include "auto_track.h"
-#include "launch_control.h"
-
+/* Make the radius 3x the size of start/finish radius.*/
+#define GEO_TRIGGER_RADIUS_MULTIPLIER 3
```
brentpicasso: is this actually used anywhere?

stieg: Yep.  Used on the Schmitt style GeoTriggers that protect start and finish from false positives.

brentpicasso: it's weird, I don't see it elsewhere in the PR, yet it seems to be first added here.

stieg:
Filed https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/369 to track refactor of the code.

stieg:
test this please

src/lap_stats/lap_stats.c
stieg: Well.. good eyes.  This got accidentally dropped when I refactored the method that creates the GeoTriggers.  Restored.  Good call.

stieg:
test this please

src/lap_stats/lap_stats.c
brentpicasso: Any thoughts on the above? nothing actionable on the current release, but thoughts for the future.

brentpicasso:
AFAICT this looks good. we can smoke test further when we fully qualify 2.9.0.

src/lap_stats/lap_stats.c
stieg: Its such a small distance (up to 11 meters, the radius of our s/f circles) that its not worth the effort.  It would have to get much larger than that for it to validate the engineer/hours to improve upon it.  I would say maybe you lose a second of data as is on a moderate race car, and that is in the worst case.  We have plenty of other things to worry about besides this matter if you ask me.

stieg:
I'm going to drop in an entry into the change log.  Then I'll merge this.

PR #368, opened 2016-01-08
[RFC] Add official unit support

stieg:
This patchset adds in official units support as outlined at [units wiki page](https://github.com/autosportlabs/RaceCapture-Pro_firmware/wiki/Units-Supported).  Effectively it enforces that we use only supported units and defines them in a way that will allow us to later add support for converting between any two units within the same unit group.

This change will probably need a bump of the X version on firmware (because we are now enforcing unit values instead of simply allowing freeform).  Thus this is only a request for comments pull so people can get eyes on it.  Please **DO NOT MERGE**.

Enjoy.


stieg:
test this please

stieg:
Need to add support for validation of a new channel when invoked via the LUA addVirtualChannel methods

PR #372, opened 2016-01-11
convert temp and speed units to match the defaults

brentpicasso:
...provided in the channel config (degrees F / MPH). This will suffice until we have formal unit and conversion support in firmware.


src/OBD2/OBD2.c
```diff
@@ -24,6 +24,16 @@ int OBD2_get_current_PID_value(int index)
     return OBD2_current_values[index];
 }
 
+static float to_farenheight(float celcius)
+{
+    return celcius * 1.8 + 32.0;
+}
+
+static float to_mph(float kph)
```
stieg: Hrmm... a method like this already exists in lap_stats.c (converts km to miles).  I would rename the methods to be more explicit like `celcius_to_farenheight` as that is more readable in code that simply `to_farenheight` which doesn't explicitly say what the original unit was.

stieg:
Looks good but there are merge conflicts.

src/OBD2/OBD2.c
brentpicasso: Ok. I'll update the names per suggested.  This will get worked over pretty hard in 2.9.0, though.

brentpicasso:
Ok. fixed conflicts, will merge.

PR #373, opened 2016-01-11
R/2.8.6 actual

brentpicasso:
Here's a release candidate


stieg:
Seems sane.  As with any firmware release lets smoke test this before we release it to the general public.

stieg:
test this please

Issue #352, opened 2015-12-20
Command line terminal mode causes watchdog


brentpicasso:
When entering terminal mode from command prompt (startTerminal command)  RCP will watchdog.

does not reproduce in 2.8.4

repro steps:
> startTerminal 4 115200 0

(wait for watchdog)

stieg:
Fixed.

PR #375, opened 2016-01-11
Fix terminal pass-through mode on firmware

stieg:
This terminal pass-through mode regressed in 2.8.5 (caused a watchdog
every time it was run).  Cause was due to the priority shift of the
tasks.  Since this caused a busy while loop in a high priority task,
lower priority tasks (such as the logger which resets the watchdog)
were starved of resources and thus the issue arose.

This patch addresses the issue by draining both tx and rx buffers on
every pass and then delays for a bit before running again.  This
prevents this busy while from starving the other tasks.

Issues: #352

### DIFF
--- a/src/logger/loggerCommands.c
+++ b/src/logger/loggerCommands.c
@@ -45,6 +45,9 @@
 
 #include <stdbool.h>
 
+/* Time (ms) to wait for input before continuing */
+#define TERM_WAIT_MS 5
+
 void TestSD(Serial *serial, unsigned int argc, char **argv)
 {
     int lines = 1;
@@ -96,12 +99,11 @@ void ResetConfig(Serial *serial, unsigned int argc, char **argv)
 static void StartTerminalSession(Serial *fromSerial, Serial *toSerial, uint8_t localEcho)
 {
         bool stay = true;
+        const size_t delay_ticks = msToTicks(TERM_WAIT_MS);
         while (stay) {
-                /* Delay so we don't starve other tasks (like watchdog) */
-                delayMs(5);
-
                 char c;
-                while (fromSerial->get_c_wait(&c, 0)) {
+                /* Delay so we don't starve other tasks (like watchdog) */
+                while (fromSerial->get_c_wait(&c, delay_ticks)) {
                         if (c == 27)
                                 stay = false;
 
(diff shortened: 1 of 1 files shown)


src/logger/loggerCommands.c
```diff
@@ -93,44 +95,57 @@ void ResetConfig(Serial *serial, unsigned int argc, char **argv)
...
 static void StartTerminalSession(Serial *fromSerial, Serial *toSerial, uint8_t localEcho)
 {
+        bool stay = true;
+        while (stay) {
+                /* Delay so we don't starve other tasks (like watchdog) */
+                delayMs(5);
 
-    while (1) {
-        char c = 0;
-        if (fromSerial->get_c_wait(&c, 0)) {
```
brentpicasso: The existing code will work perfectly if you change the 0 to some value. Right now the 0 means 'don't wait at all' . The parameter is number of ticks to wait.  Suggest you change it to:

if (fromSerial->get_c_wait(&c, msToTicks(IDLE_DELAY_MS))) {

where IDLE_DELAY_MS = 5  or whatever you want it to be.

brentpicasso: behind the scenes it calls this function: http://www.freertos.org/a00118.html.  While it's waiting the task scheduler will be free to work on other things.

stieg: Interesting approach.  I'll take a look at doing it that way.

src/logger/loggerCommands.c
```diff
@@ -93,44 +98,56 @@ void ResetConfig(Serial *serial, unsigned int argc, char **argv)
...
-            if (c == 27) break;
-            if (localEcho) fromSerial->put_c(c);
-            if (c == '\r' && localEcho) fromSerial->put_c('\n');
-            toSerial->put_c(c);
-            if (c == '\r') toSerial->put_c('\n');
+        bool stay = true;
+        const size_t delay_ticks = msToTicks(TERM_WAIT_MS);
+        while (stay) {
```
brentpicasso: why not use break as we had before? it is less complex.

stieg: Because of the nested `while` statements.

brentpicasso:
Gotcha. you're draining rx/tx as much as possible with the inner while.  I like.  :shipit:

Issue #376, opened 2016-01-12
Add bitOp support to Lua Runtime (MK2)


brentpicasso:
Need the ability to do bitwise operations for Lua scripting, to enable bitwise mapping for CAN bus messages.

https://luarocks.org/modules/luarocks/luabitop

stieg:
Resolving since @brentpicasso merged this fix.

Issue #185, opened 2015-10-26
implement double as number for Lua


ryandoherty:
to fix extended CAN ID issue. (29 bit IDs)

stieg:
@brentpicasso fixed this as part of the bitops fix to LUA.  @brentpicasso Did you test if this resolves the 29 bit CAN value issue?

brentpicasso:
Yes, it does, b/c our problem was with CAN IDs being destroyed by our use
of float. I had tested it back when it was first identified.

On Thu, Jan 14, 2016 at 3:17 PM, Andrew Stiegmann <notifications@github.com>
wrote:

> @brentpicasso <https://github.com/brentpicasso> fixed this as part of the
> bitops fix to LUA. @brentpicasso <https://github.com/brentpicasso> Did
> you test if this resolves the 29 bit CAN value issue?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/185#issuecomment-171813433>
> .
>



-- 
Brent Picasso
*Autosport Labs <http://www.autosportlabs.com>*
*Technology for Race and Street*

stieg:
I meant have you verified that the fix actually works.

brentpicasso:
I have verified it in the past via a temporary change to LUA_NUMBER to double,  I have not checked it again since we added bitOp just recently.

stieg:
Ok.  That is probably good enough.  Going to close this for now.

PR #378, opened 2016-01-13
P/brentpicasso/lua bitop 2.9.x

brentpicasso:
* Enable bit operations for Lua
* Convert lua number to double, which has the side effect of fixing 29 bit CAN IDs.


stieg:
Addresses #185 and #376

lib_lua/src/luaconf.h
```diff
@@ -506,9 +506,12 @@
 #if defined(LUA_NUMBER_INTEGER)
 #define LUA_NUMBER      int
 #else
-#define LUA_NUMBER      float
+#define LUA_NUMBER      double
 #endif
 
+#define LUA_NUMBER double
+#define LUA_NUMBER_DOUBLE LUA_NUMBER
```
stieg: Why is this duplicated here?

brentpicasso: bitOp wants to see LUA_NUMBER_DOBULE defined for some reason.

lib_lua/src/luaconf.h
```diff
@@ -527,7 +530,7 @@
 #if defined(LUA_NUMBER_INTEGER)
 #define lua_number2str(s,n)     modp_itoa10((n),s)
 #else
-#define lua_number2str(s,n)     modp_ftoa((n),s, LUA_NUMBER_PRECISION)
+#define lua_number2str(s,n)     modp_dtoa((n),s, LUA_NUMBER_PRECISION)
```
stieg: We should use the stdlib functions as we are using those now and will be getting rid of the modp methods in a later release

stieg: Err.. by stdlib I mean the methods in string.h that will convert a double to a char array.

brentpicasso: for 2.9.x or the later release? unclear.

lib_lua/src/luaconf.h
stieg: I mean why the re-definition in line 512?

stieg:
Minor nits.  Address and :shipit:

lib_lua/src/luaconf.h
brentpicasso: left over cruft. fixed.

lib_lua/src/luaconf.h
brentpicasso: I'll save the modp_xxxx for a holistic sweep through when we do it in a later release.

Issue #379, opened 2016-01-13
Fix BT password regression introduced with new BT features.


stieg:
Need to fix the regression @brentpicasso found.  Outlined in the following chat:

```
​[22:09] 
brent Ok, I tried again and saw something really weird.

I had already paired the RCP Bt module in android.  (specified pwd, everything)..  I relaunched the app and when I ran the app android challenged me to re-pair and specify password ​*again*​.  After that, I was able to connect again no problem
​[22:10] 
App connected and read data from RCP
​[22:10] 
That's very odd
​[22:11] 
I wonder if it has something to do with the 0000 pin.. I thought I remembered there was something special about that number.
​[22:12] 
Yes, I power cycled the BT module and it did it again
​[22:12] 
after re-launching the app
​[22:16] 
Correction: I power cycled RCP which caused it to send the BT module new configuration commands
​[22:16] 
​*that*​ causes android to issue the password challenge again
​[22:16] 
I'll take a video.
​[22:32] 
ryan That is really craY
​[22:40] 
brent I did a quick test of disabling the setting of password
​[22:40] 
apparently setting the password again (even if it's the same value) causes the problem
​[22:42] 
stieg Oh... Interesting
​[22:42] 
Ok... That should be easy to fix
```
Demo'd here: https://www.youtube.com/watch?v=cu9ZFWHxg1o

stieg:
Wonderful.  It would seem that there is no way to read the current PIN using AT commands from the device.  Here is the full documentation: https://www.olimex.com/Products/Components/RF/BLUETOOTH-SERIAL-HC-06/resources/hc06.pdf

So we either need to simply leave the default password or need to come up with a creative way to know that the password changed and thus needs to be updated.

brentpicasso:
Yeah, it's crap. :(

Before, the one way I knew the module was not provisioned was to check if it's at 9600 baud. If it was at 115200, I assumed it was already provisioned.

Maybe setting the Name / PIN could be an explicit action.

stieg:
#383 should solve those issues.  Similar to what you described @brentpicasso

stieg:
Fixed and verified.

PR #382, opened 2016-01-13
Further refine BT configuration process

stieg:
It turns out on the HC-06 (the BT device we use for MK1 and MK2) that
if you set the PIN (even if it doesn't get changed) you cause the
pairing between the tablet to be broken.  This is unfortunate.  To
remedy this problem, this patch changes the meaning of the BT name
and pin values to be new ones set at the next boot of RCP.  Once
that change occurs, the values are cleared.  This prevents the original
problem while still allowing customers to change the BT settings.

It also maintains backwards compatibility with the original code so
that if we detect that the BT module is in factory mode, it does the
proper initial configuration.

Issues: #379

### DIFF
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -28,6 +28,7 @@
 #include "capabilities.h"
 #include "versionInfo.h"
 
+#include <stdbool.h>
 #include <stddef.h>
 
 CPP_GUARD_BEGIN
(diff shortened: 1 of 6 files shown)


stieg:
Before you ask... I'm testing it now

stieg:
Hrmm... seems to partially work and partially not work.  Needs more diagnosis.

include/logger/loggerConfig.h
```diff
@@ -434,25 +435,16 @@ typedef struct _TrackConfig {
...
 
 typedef struct _BluetoothConfig {
-    unsigned char btEnabled;
-    char deviceName [BT_DEVICE_NAME_LENGTH];
-    char passcode [BT_PASSCODE_LENGTH];
-    unsigned int baudRate;
+        unsigned char btEnabled;
+        unsigned int baudRate;
+        char new_name [BT_DEVICE_NAME_LENGTH];
+        char new_pin [BT_PASSCODE_LENGTH];
```
brentpicasso: s/b BT_PIN_LENGTH  (being pedantic )

brentpicasso: Also, I'm wondering if we use the HC-05 in the future, would we re-use the same struct? Because then, we'd be able to get the device into configuration mode and do it the right way.  In this case, we'd want the struct member names simply name / pin.   Thoughts? Or would we create a new struct for this?

src/devices/bluetooth.c
```diff
@@ -206,16 +232,31 @@ int bt_init_connection(DeviceConfig *config)
          * factory level (9600).  This ensures things go slow enough for the
          * HC-06 processor + code to handle it.
          */
+        int baud = bt_find_working_baud(config, btConfig);
+        switch (baud) {
+        case 0:
+                pr_info("BT: Failed to communicate with device.  Assuming it "
```
brentpicasso: so if the BT module is already paired with android you'll arrive on this line.  So the text should say "Assuming already configured, moving on" or some such. 

Because it could either be configured / connected or simply disconnected.  Sigh, HC-06...

src/devices/bluetooth.c
```diff
@@ -206,16 +232,31 @@ int bt_init_connection(DeviceConfig *config)
...
+                        "is not connected.\r\n");
+                g_bluetooth_status = BT_STATUS_ERROR;
+                return DEVICE_INIT_FAIL;
+        case 9600:
```
brentpicasso: I think you also want to check if the current baud is BT_AT_CMD_BAUD then you want to fall through and assume it's already configured?

brentpicasso:
So far clean implementation. Noted some logic concerns.

include/logger/loggerConfig.h
stieg: If we move to another module lets address that issue then.  I think we would probably use the same struct with additions because its super basic.

src/devices/bluetooth.c
stieg: Addressed.

src/devices/bluetooth.c
stieg: This is now addressed.

PR #383, opened 2016-01-14
MOAR fixes for BT.  Fast init support. Name + PIN change support.

stieg:
Ok... so there is a lot happening here.  Here is the brief overview:

* Setting the PIN causes BT to need to re-pair with devices.  Because
  of this, we must only set the PIN when a user requests a change.
  This is implemented here.

* Adjust the init process to be as fast as possible.  This is because
  there is a race condition between the initialization code and a
  BT device trying to pair with the unit.  If the BT device connects,
  it makes init impossible.  Thus this code must do only what is
  necessary.  Killed version info and dropping baud down to 9600
  as a result

* Implement legacy init behavior.  To be inline with what the original
  BT code did, we must detect factory settings on the HC-06 and set
  RCP defaults.  We do this by no longer setting the 9600 baud and
  looking for it during init.  If we see it, we assume its never been
  initialized (beacuse 9600 baud is the default from the factory)
  and initialize it accordingly.

* Add re-try logic to the baud probing.  Because the unit can't move
  as fast as we do.  Thus we need to be a bit flexible.

* Adjust the communication parameters to SANE values.  This thing is
  slow and works in wonkey ways (like expecting a 500ms pause before
  interpreting the AT command).  So this change adjusts those parameters
  to reflect this behavior yet move as fast as the module will allow.
  Adjust these at your own peril.

Did I miss anything... probably.  But that is what you need to know.

Enjoy!

Issues: #379

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -7,8 +7,6 @@
   sensitivity (8x more to be precise).
 * Improve the MK2 compilation routine by adding stricter warning flags and
   by making all warnings into errors.
-* Add uniqueness to BT device name by appending last 4 digits of unit serial
-  to the BT device.  Applies after logger reset.
 * Fix up BT code so that we now properly set BT name and Pin.  Addresses race
   issues observed in BT module during programming as well.
 * Added Lua sleep() function. When called will sleep for the specified number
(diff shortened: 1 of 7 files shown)


brentpicasso:
Looks solid. :shipit: 

Do you need second eyes to test it?

brentpicasso:
Did a cursory test on my MK2 hardware. looks good AFAICT

stieg:
We will get its final test on the 2.9.0 final validation.  I tested the crap out of it yesterday.

Issue #164, opened 2015-10-26
remove obsolete top level obd2SampeRate value from struct _OBD2Config


ryandoherty:
This breaks the overall config struct, so need to save it for when we do a minor rev change (not bugfix release)

stieg:
Fixed.

PR #384, opened 2016-01-14
Remove "obd2SampleRate" value from OBD2Config struct

stieg:
Its not used.  So nuke it from orbit.

Issues: #164


brentpicasso:
:shipit:

Issue #360, opened 2016-01-03
Switch accelerometer to SAE standard orientation


brentpicasso:
X should be Forward / aft G's
Y should be lateral G's
Z should be vertical G's

stieg:
The following links clarify:

* https://claraty.jpl.nasa.gov/man/software/development/conventions/coord_frames.php
* https://law.resource.org/pub/us/cfr/ibr/005/sae.j1733.1994.html
* http://www.formula1-dictionary.net/Images/motion_yaw_pitch_roll.jpg

PR #385, opened 2016-01-15
Set the default IMU calibration to the SAE standard

stieg:
This change sets the default IMU config to conform to the SAE standard.
It also creates enum types for the channels and modes and ups the precision
on the gyro portions since this is now more sensitive.  It also addresses
the fallout from making these default changes.

Issues: #360


include/logger/loggerConfig.h
```diff
@@ -216,59 +216,62 @@ typedef struct _GPIOConfig {
...
-         MODE_IMU_NORMAL,                       \
-         IMU_CHANNEL_YAW,                       \
-         DEFAULT_GYRO_ZERO,                     \
-         0.1F                                   \
-         }
+#define MIN_IMU_RAW	0
+#define MAX_IMU_RAW	4097
+
+#define DEFAULT_ACCEL_ZERO	2048
+/* LY330ALH zero state voltage output is 1.5v */
```
brentpicasso: LY330ALH is from MK1; I think we need to reset this number to be the actual yaw zero value for the MPU-9150 / 9250

include/logger/loggerConfig.h
```diff
@@ -216,59 +216,62 @@ typedef struct _GPIOConfig {
...
+#define IMU_GYRO_CONFIG(name, mode, chan) {     \
+                IMU_GYRO_CH_CONFIG(name),       \
+                        mode,                   \
+                        chan,                   \
+                        DEFAULT_GYRO_ZERO,      \
+                        0.1F                    \
+                        }
 
+#define IMU_CONFIG_DEFAULTS {                                                 \
+                IMU_ACCEL_CONFIG("AccelX", IMU_MODE_INVERTED, IMU_CHANNEL_Y), \
```
brentpicasso: need some code comments here to explain why we're doing this mapping

brentpicasso:
This PR should be re-worked to not do the SAE mappings at the higher level, but should be remapped within the HAL layer (imu_device_stm32.c), function at line 72  (imu_device_read). 

Rationale: If defined at a higher level, then users will see in the app config, some channels inverted or not inverted, which will appear confusing, and adding cognitive load to understanding that we're mapping ISO to SAE from the interface we're presenting them.

By doing it at a lower level we're in effect converting the IMU device to be "SAE compliant" . The result is, the default configuration for the IMU would have X mapped to X, Y mapped to Y, no channel inversions, etc.

include/logger/loggerConfig.h
brentpicasso: NVM, see top level PR comment.

stieg:
You are right.  Will re-tool the patch.

include/logger/loggerConfig.h
stieg: Yeah something doesn't seem right for me here either.  I'll play with this a bit.

brentpicasso: A simple update to the zero value that's appropriate for the MPU-9150/MPU-9250 would do the trick I think.

stieg: It does.  In next patch.

stieg:
@brentpicasso There we go.  Check it out now.

brentpicasso:
Looks good. I assume it's been bench tested; go ahead and :shipit:

stieg:
Yep.  Tested and verified.

On Thu, Feb 11, 2016 at 10:07 AM, Brent Picasso <notifications@github.com>
wrote:

> Looks good. I assume it's been bench tested; go ahead and [image:
> :shipit:]
>
> —
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/385#issuecomment-182936154>
> .
>



--

PR #388, opened 2016-01-17
report correct values for min/max in filewriter

brentpicasso:
min / max from meta data was incorrectly filtered through sample rate decoder. Fix this defect.


stieg:
:shipit:

PR #396, opened 2016-01-19
Fix pred time value if predicted time unit not in operation.  Bump version

stieg:
Fixes predictive time to return 0 if the unit is not operational.  Otherwise it can return the wrong value.  Also bumps the version to 2.8.7.


brentpicasso:
:shipit:

Issue #397, opened 2016-01-21
getAltitude() Lua binding


brentpicasso:
Create a Lua binding for fetching the current altitude.

PR #405, opened 2016-01-27
P/stieg/get gps altitude

stieg:
Adds support for getGpsAltitude method.  Also adds a useful clean method in our makefile because I am sick of not having it.


stieg:
Also fixes issue #397

brentpicasso:
looks good

Issue #404, opened 2016-01-26
Lua txCAN timeout parameter has no effect


brentpicasso:
The timeout message is available in the lua function, but has no effect down in the STM32 HAL. 

need a way to enable a timeout, or remove the timeout parameter.

Also - in the HAL layer the CAN_device_tx_msg always returns 1; should return 1 or 0 indicating success.

stieg:
Seems that the timeout portion was skipped in the hal implementation.  I see methods for CAN message status and cancellation in the stm32f4xx_can.h file:

```C
/* CAN Frames Transmission functions ******************************************/
uint8_t CAN_Transmit(CAN_TypeDef* CANx, CanTxMsg* TxMessage);
uint8_t CAN_TransmitStatus(CAN_TypeDef* CANx, uint8_t TransmitMailbox);
void CAN_CancelTransmit(CAN_TypeDef* CANx, uint8_t Mailbox);
```

So probably best to implement the timeout here in HAL.

stieg:
This is a 2.9.0 feature since it changes behavior (albeit for the better).

brentpicasso:
Yeah. the Rx method conveniently uses the FreeRTOS queue for timeout.

Since only Lua uses the txCAN function (so far) - we can implement a cheap
sleep() delay loop up until the timeout if the STM32 CAN transmit function
fails to actually send the message. Wouldn't be terrible.

-Brent

On Wed, Jan 27, 2016 at 2:16 PM, Andrew Stiegmann <notifications@github.com>
wrote:

> This is a 2.9.0 feature since it changes behavior (albeit for the better).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/404#issuecomment-175889226>
> .
>



-- 
Brent Picasso
*Autosport Labs <http://www.autosportlabs.com>*
*Technology for Race and Street*

stieg:
Patch tested and merged.

PR #406, opened 2016-01-27
P/stieg/can tx timeout support

stieg:
Adds support for timeout values on the MK2 CAN Tx method.

Tested using the following LUA script:
```lua
function p_uptime(msg)
	println("Uptime: " .. getUptime() .. msg)
end

function p_uptime_status(result)
	p_uptime(", Status: " .. result)
end

function onTick() 
	data = {11,22,33}

	p_uptime(" At Test 1")
	res = txCAN(0, 1234, 0, data)
 	p_uptime_status(res)


	p_uptime(" At Test 2 (delay 92ms)")
	res = txCAN(0, 1234, 0, data, 92)
 	p_uptime_status(res)

	p_uptime(" At Test 3 (no delay)")
	res = txCAN(0, 1234, 0, data, 0)
 	p_uptime_status(res)
end
```

PR #407, opened 2016-01-31
Add timeout support to the CAN Tx method on MK2

stieg:
Previously ignored the timeout value and always assumed that the
message would send in time.  Now we will pay attention to the value
and will cancel the message if it has not sent in the specified
timeout.

A timeout value of 0 means we don't care whether it sent or not and
will always return success.

Issue: #404


stieg:
Feedback applied from #406.

PR #408, opened 2016-02-02
P/stieg/merge rct mainline

stieg:
Merges code from the RCT codebase into v2.9.0 for mainline inclusion when we fold
2.9.0 into master.


brentpicasso:
like the use of #defined() .  Glad to get some various latent RCT firmware fixes in the mainline now. 

:shipit:

PR #413, opened 2016-02-07
Fix Lua Stack Smash. Improve OpenOCD Debug capabilities.

### LINKED ISSUE #411: LUA is smashing its stack during parsing
User @imstimpy [reported on the forums](https://www.autosportlabs.org/viewtopic.php?p=23733#23733) that he was hitting a crash bug.  After analysis of the script it was found that the method of iteration over a table was causing the crash.  Deeper analysis shows that the culprit is a stack-overflow.  The resulting LUA code, while valid, caused the internal parser to blow its stack (already at 1000 characters).  Obviously the fix using the iterator prevented the stack from doing this.  Solution is to give the LUA task a larger stack to help avoid this.

stieg:
Fixes #411 by upping the stack size on the LuaTask.

To trace the issue back to this, had to improve how we debug in OpenOCD. These changes are included as well (but are in separate commits). This PR also introduces a 'DEBUG' make variable so that we can build DEBUG builds that are useful for tracing down issues like this without having to make many many changes.

### DIFF
--- a/main.c
+++ b/main.c
@@ -53,11 +53,14 @@
 #include <app_info.h>
 #include <stdbool.h>
 
-#define FATAL_ERROR_SCHEDULER	1
-#define FATAL_ERROR_HARDWARE	2
+#define FLASH_PAUSE_DELAY_S 	5
+#define FLASH_DELAY_S		1
 
-#define FLASH_PAUSE_DELAY 	5000000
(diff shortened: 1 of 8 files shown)


brentpicasso:
looks sane and solid. :shipit:

PR #415, opened 2016-02-08
OBD2 Update for more codes and validation of code

imhoffj:
Added more OBD2 PIDs and validation of lua script with OBD2 to read out
values. Here is the Gist for the shift light: https://gist.github.com/imhoffj/e2f1d9d273f59e15747f. You already had the code, just need to send "12" PID for RPM and then use the shift light. I verified working and added some debugging, but also added some more PID values and looking into finding more for built in turbo and other items.

### DIFF
--- a/src/OBD2/OBD2.c
+++ b/src/OBD2/OBD2.c
@@ -56,6 +56,13 @@ static float kph_to_mph(float kph)
     return kph * 1.60934;
 }
 
+static float kPa_to_psig(float kPa)
+{
+    return (kPa * 0.1450377377)-14.6959494;
+}
+
+//look into data bytes beyond 4
+//need 4F,50, 55,56,57,58,64
 static int decode_pid(unsigned char pid, CAN_msg *msg, int *value)
 {
     int result = 0;
@@ -64,14 +71,13 @@ static int decode_pid(unsigned char pid, CAN_msg *msg, int *value)
             msg->data[0] >= 3 &&
             msg->data[1] == CUSTOM_MODE_SHOW_CURRENT_DATA &&
             msg->data[2] == pid ) {
-
-
+    	
         result = 1;
-
+        
         int A = msg->data[3];
         int B = msg->data[4];
-        //int C = msg->data[5]; //save for later
-        //int D = msg->data[6];
+        int C = msg->data[5];
+        int D = msg->data[6];
 
         switch(pid) {
         case 0x04: //calculated engine load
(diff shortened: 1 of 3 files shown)


src/OBD2/OBD2.c
```diff
@@ -110,12 +116,140 @@ static int decode_pid(unsigned char pid, CAN_msg *msg, int *value)
...
+        case 0x19: //Oxygen sensor voltage - Bank 2, Sensor 2
+        case 0x1A: //Oxygen sensor voltage - Bank 2, Sensor 3
+        case 0x1B: //Oxygen sensor voltage - Bank 2, Sensor 4
+            *value = A / 200;
+            break;
+        case 0x1F: //Run time since engine start
+            *value = (A * 256) + B;
+            break;
+        case 0x22: //Fuel rail Pressure (relative to manifold vacuum) kPa
+            *value = ((A * 256) + B) * 0.079;
```
stieg: Why the different multiplier here vs in 0x23?  Are they not both in kPa?

src/OBD2/OBD2.c
```diff
@@ -110,12 +116,140 @@ static int decode_pid(unsigned char pid, CAN_msg *msg, int *value)
...
+            break;
+        case 0x23: //Fuel rail Pressure (diesel, or gasoline direct inject) kPa 
+            *value = ((A * 256) + B) * 10;
```
stieg: It would be clearer to write `(A << 8 | B)` instead of the math since that is the intention here.

src/OBD2/OBD2.c
```diff
@@ -110,12 +116,140 @@ static int decode_pid(unsigned char pid, CAN_msg *msg, int *value)
         case 0x11: //throttle position %
             *value = A * 100 / 255;
             break;
+        case 0x14: //Oxygen sensor voltage - Bank 1, Sensor 1
```
stieg: Note that we use the Linux standard for code.  So C-89 style comments only please `/* ... */`

src/OBD2/OBD2.c
```diff
@@ -56,6 +56,13 @@ static float kph_to_mph(float kph)
     return kph * 1.60934;
 }
 
+static float kPa_to_psig(float kPa)
```
stieg: Please add a comment about the meaning of `psig`.  I never knew the relative term.  TIL.

src/logger/luaLoggerBinding.c
```diff
@@ -733,6 +733,11 @@ int Lua_ReadOBD2(lua_State *L)
         unsigned char pid = (unsigned char)lua_tointeger(L, 1);
         int value;
         if (OBD2_request_PID(pid, &value, OBD2_PID_DEFAULT_TIMEOUT_MS)) {
+        	 if (DEBUG_LEVEL) {
```
stieg: Not needed.

src/lua/luaTask.c
```diff
@@ -22,6 +22,7 @@
 
 #include "FreeRTOS.h"
 #include "GPIO.h"
+#include "OBD2.h"
```
stieg: Needed?  I see no other changes to the file.

stieg:
ok to build

stieg:
ok to test

stieg:
Looks pretty good.  We don't work off of master on this project.  Instead we work off of the /r branches.  Since this is improving an existing feature I would say this can should get on the next .Z release.  Once you have applied the feedback will you re-open the patch against r/2.8.8?  Thanks for the patch @imhoffj

brentpicasso:
Thank you @imhoffj for the PR! This is great. Out of curiosity, what source did you pull the additional codes - was it from wikipedia or a difference source? How many have you been able to test?

imhoffj:
Wiki and http://www.outilsobdfacile.com/obd-mode-pid.php. I am looking into testing the turbo and electric, but trying to research out the custom items. I will update against the r/2.8.8 this week.

stieg:
Hey @imhoffj,

stieg:
ok to test

stieg:
Hey @imhoffj,

   Was hoping you had time to implement the feedback we gave.  If you don't let us know and we can do some work on the patch if we find some time.  It would be great to get this into 2.8.8.

imhoffj:
I have made the updates in a separate pull request - still learning the Github flow so apologies. I incorporated all comments except  "It would be clearer to write (A << 8 | B) instead of the math since that is the intention here." because I wanted to keep it standard to anyone looking into it to keep it as math.

stieg:
Fair.  Its not the way I would do it... but since the result is the same I won't argue it.  I put my feedback in the other patch (just white space bits is all).  I'll close this one then.  Thanks for the patch.  Looking forward to the new features it will add.

Issue #182, opened 2015-10-26
ublox 3G module support


ryandoherty:
support both modules in same firmware - need an auto detect feature to switch between driver implementations. there's an AT command that indicates the model #.

stieg:
ANNNNNNNNDDDDDDD ITS DONE

PR #416, opened 2016-02-08
3G Support!

stieg:
Holy crap this took way too long to do.  But here it is.  This patch implements 3G support for RCP with backwards support for SIM900.  In other words, it is a fuax driver loading mechanism.  After some extensive testing it seems to work quite well and is very fast in bringing up the modems.

Also in this change is a lot of new features that will be useful when support for WiFi chip is added.  This includes part of a new AT command framework, introduction of SerialBuffers, and various improvements to help me debug this project.

The commits are bit of a mess... sorry.  I have learned my lesson with this one.


stieg:
Solves #182

include/logger/loggerConfig.h
```diff
@@ -445,32 +445,39 @@ typedef struct _BluetoothConfig {
...
+        char apnPass [CELL_APN_PASS_LENGTH + 1];
+        char dns1[DNS_ADDR_LEN];
+        char dns2[DNS_ADDR_LEN];
 } CellularConfig;
 
 
-#define DEVICE_ID_LENGTH 36
-#define TELEMETRY_SERVER_HOST_LENGTH 100
+#define DEVICE_ID_LENGTH 35
+#define TELEMETRY_SERVER_HOST_LENGTH 95
```
brentpicasso: why not 99 to be consistent with -1  reduction of DEVICE_ID_LENGTH?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+}
+
+static void cell_serial_tx_cb(const char *data)
+{
+        static bool pr_tx_pfx = true;
+
+        for(; *data; ++data) {
+                if (pr_tx_pfx) {
+                        printk(serial_dbg_lvl, "[cell] tx: ");
+                        LED_toggle(0);
```
brentpicasso: left over debug code?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+
+        for(; *data; ++data) {
+                if (new_line) {
+                        printk(serial_dbg_lvl, "[cell] rx: ");
+                        new_line = false;
+                }
+
+                switch(*data) {
+                case('\r'):
+                        printk(serial_dbg_lvl, "<cr>");
```
brentpicasso: curious, why display <cr> instead of the actual carriage return character?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+{
+        serial_buffer_append(sb, "\r");
+        serial_buffer_tx(sb);
+        serial_buffer_clear(sb);
+        const size_t rv = serial_buffer_rx_msgs(sb, wait, rx_msgs,
+                                                rx_msgs_size);
+
+        /*
+         * URC Pause (must be > 20ms)
+         * TODO: Add support for handling URCs.  Right now we ignore them.
```
brentpicasso: what is a URC?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+         *       This will likely require a re-write of how cellular is
+         *       done.
+         */
+        if (NULL == at_cfg)
+                at_cfg = get_safe_at_config();
```
brentpicasso: spidey senses tingling. is there a race condition possible where a detected modem config could get wiped by this code?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+        case CELLULAR_NETWORK_DENIED:
+                return "Denied.  Call provider.";
+        case CELLULAR_NETWORK_REGISTERED:
+                return "Registered";
+        default:
+                return "Unknown";
+        }
+}
+
+int cellular_disconnect(DeviceConfig *config)
```
brentpicasso: should this return bool?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+
+        /*
+         * SIM900 requres at least 1 second of pull for powerdown
+         * and then 1 second of non-pull for powerup.  This however
+         * can screw with Ublox-Sara, which needs to autobaud very
+         * shortly (< 3sec) after powerup.  This balance seems to work
+         * "Well enough"(tm).
+         */
+        cell_pwr_btn(true);
+        delayMs(1100);
```
brentpicasso: extract magic number pls

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+}
+
+/**
+ * Puts the device in a state that is ready to use.
+ */
+static int cellular_init_modem(struct serial_buffer *sb)
+{
+        reset_modem();
+
+        if (!autobaud_modem(sb, 5, 200))
```
brentpicasso: magic numbers

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+        json_objEnd(serial, 0);
+        serial->put_c('\n');
+
+        pr_debug_str_msg("sending auth- deviceId: ", deviceId);
+
+        /* Parse the incoming JSON */
+        jsmn_parser parser;
+        jsmntok_t toks[AUTH_MAX_JSMN_TOKENS];
+
+        for (size_t tries = 20; tries; --tries) {
```
brentpicasso: magic number.  And why 20 attempts?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+        if(!methods->open_telem_connection(sb, &cell_info, &telemetry_info,
+                                           telemetryConfig)) {
+                telemetry_info.status =
+                        TELEMETRY_STATUS_SERVER_CONNECTION_FAILED;
+                print_registration_failure();
+                return DEVICE_INIT_FAIL;
+        }
+
+        /* Auth against RCL */
+        if (!auth_telem_stream(sb, telemetryConfig)){
```
brentpicasso: Would it be useful to disambiguate a condition where we just couldn't get a definitive response from back-end? e.g. never got an auth'd or not-auth'd response?  This way,  we say it's rejected even if there's a network issue.

src/devices/cellular.c
brentpicasso: regarding attempts; what happens if our cellular connection drops and we have many connection attempts remaining?  Do we keep attempting to communicate with a non-existent back-end?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+        }
+
+
+        telemetry_info.status = TELEMETRY_STATUS_CONNECTED;
+        telemetry_info.active_since = getUptime();
+
+	return DEVICE_INIT_SUCCESS;
+}
+
+int cellular_check_connection_status(DeviceConfig *config)
```
brentpicasso: does this function actually get used? can't find it in the diff.

src/devices/sara_u280.c
```diff
@@ -0,0 +1,467 @@
...
+
+        serial_buffer_reset(sb);
+
+        /* APN settings */
+        serial_buffer_printf_append(sb, "AT+UPSD=0,1,\"%s\"", host);
+        serial_buffer_printf_append(sb, ";+UPSD=0,2,\"%s\"", user);
+        serial_buffer_printf_append(sb, ";+UPSD=0,3,\"%s\"", password);
+
+        /* Dynamic IP */
+        //serial_buffer_append(sb, ";+UPSD=0,7,\"0.0.0.0\"");
```
brentpicasso: comment style

src/devices/sara_u280.c
```diff
@@ -0,0 +1,467 @@
...
+        return status;
+}
+
+static bool sara_u280_activate_pdp(struct serial_buffer *sb,
+                                   const int pdp_id)
+{
+        return sara_u280_gprs_psd_action(sb, pdp_id, 3);
+}
+
+#if 0
```
brentpicasso: remove dead code?

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+                }
+        }
+}
+
+size_t cellular_exec_cmd(struct serial_buffer *sb,
+                         const size_t wait,
+                         const char **rx_msgs,
+                         const size_t rx_msgs_size)
+{
+        serial_buffer_append(sb, "\r");
```
brentpicasso: is this designed to send multiple AT commands at once, and wait for all the respective replies? As seen in the setup functions (DNS settings, etc)

src/devices/sim900.c
```diff
@@ -19,124 +19,396 @@
...
+        return sim900_get_imei(sb, ci) &&
+                sim900_get_signal_strength(sb, ci) &&
+                sim900_get_subscriber_number(sb, ci);
+}
+
+static bool sim900_register_on_network(struct serial_buffer *sb,
+                                       struct cellular_info *ci)
+{
+        /* Check our status on the network */
+        for (size_t tries = 10; tries; --tries) {
```
brentpicasso: magic number

src/devices/sara_u280.c
```diff
@@ -0,0 +1,467 @@
...
+        return CELLULAR_NETWORK_REGISTERED == ci->net_status;
+}
+
+static bool sara_u280_setup_pdp(struct serial_buffer *sb,
+                                struct cellular_info *ci,
+                                const CellularConfig *cc)
+{
+        /* Check GPRS attached */
+        bool gprs_attached;
+        for (size_t tries = 10; tries; --tries) {
```
brentpicasso: magic number (dup of sim 900 tries count of 10)

brentpicasso: magic numbers applies througout.  Might be good for a config include or similar that applies to both modem technologies.

src/gsm/gsm.c
```diff
@@ -0,0 +1,241 @@
...
+        char *num_end = strstr(num_start, "\"");
+        if (!num_end)
+                goto parse_fail;
+
+        *num_end = '\0';
+        strncpy(ci->number, num_start, sizeof(ci->number));
+        return true;
+
+parse_fail:
+        pr_warning("[gsm] Failed to prase phone number\r\n");
```
brentpicasso: spelling

src/gsm/gsm.c
```diff
@@ -0,0 +1,241 @@
...
+        const size_t msgs_len = ARRAY_LEN(msgs);
+
+        serial_buffer_reset(sb);
+        serial_buffer_append(sb, cmd);
+        const size_t count = cellular_exec_cmd(
+                sb, READ_TIMEOUT, const_msgs, msgs_len);
+        const bool status = is_rsp_ok(const_msgs, count);
+
+        if (!status) {
+                pr_warning("[gsm] Failed to read signal strength\r\n");
```
brentpicasso: with these failure messages, would it be useful to also include the response received, for diagnostic purposes?

src/gsm/gsm.c
```diff
@@ -0,0 +1,241 @@
...
+                pr_warning("[sara_u280] Failed to read IMEI\r\n");
+                return false;
+        }
+
+        strncpy(cell_info->imei, msgs[0], sizeof(cell_info->imei));
+
+        return 1;
+}
+
+enum cellular_net_status gsm_get_network_reg_status(
```
brentpicasso: this is great detail :+1:

src/jsmn/jsmn.c
```diff
@@ -323,3 +323,40 @@ int jsmn_isNull(const jsmntok_t *tok)
 {
     return strncmp("null", tok->data, 3) == 0;
 }
+
+/* TODO: rename these with jsmn_ prefix */
```
brentpicasso: think will this TODO be handled with this PR or later?

src/logger/loggerApi.c
```diff
@@ -291,8 +255,8 @@ int api_getStatus(Serial *serial, const jsmntok_t *json)
     json_objEnd(serial, 1);
 
     json_objStartString(serial, "telemetry");
-    json_int(serial, "status", (int)sim900_get_connection_status(), 1);
-    json_int(serial, "dur", sim900_active_time(), 0);
+    json_int(serial, "status", (int)cellular_get_connection_status(), 1);
+    json_int(serial, "dur", cellular_active_time(), 0);
     json_objEnd(serial, 0);
```
brentpicasso: Do we need to update the app to reflect any additional status detail for telemetry?

src/serial/serial_buffer.c
```diff
@@ -0,0 +1,173 @@
...
+#include "string.h"
+#include "taskUtil.h"
+
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdio.h>
+
+static void ctrl_char_strip(char *data)
+{
+        /* Strip all control characters from our messages */
```
brentpicasso: is this for trailing control characters only?

src/serial/serial_buffer.c
```diff
@@ -0,0 +1,173 @@
...
+        for(; *data >= 32; ++data);
+        *data = 0;
+}
+
+bool serial_buffer_create(struct serial_buffer *sb,
```
brentpicasso: We have quite a bit of RAM available in the CCR block of the STM32 (the 64K region). if we define these buffers statically, we can conserve the 128K RAM block for Lua and other things that need dynamic RAM. Thoughts?

src/util/mod_string.c
```diff
@@ -162,7 +162,8 @@ char *strtok_r(char * __restrict s1, const char * __restrict s2, char ** __restr
                 *p++ = 0;
             }
         } else {
-            p = s = NULL;
+            s = NULL;
+            p = NULL;
```
brentpicasso: change b/c style guide preferences? just curious.

brentpicasso:
Architecture looks solid! :+1: 

primary concerns are handling edge cases where network disappears (e.g. while we're trying to auth)  and so on. 

I'll work on testing this locally as well.

include/logger/loggerConfig.h
stieg: Word alignment to avoid unnecessary padding.

src/devices/cellular.c
stieg: No.  Right now this is a callback method called by the serial library so I can see data on a singular serial bus.

src/devices/cellular.c
stieg: Because it was easier to debug a character I could see.

src/devices/cellular.c
stieg: Unexpected Response Code.  These come in after an AT command.  Its the reason for the X ms backoff.

src/devices/cellular.c
stieg: Negative.

brentpicasso: ok

stieg: Currently this only sets the backoff time on AT commands.  Ublox only needs 25ms where as sim900 need 250 (because it is slooooooowwwwww).

src/devices/cellular.c
brentpicasso: I meant the LED toggle.

src/devices/cellular.c
stieg: Yes... but since its a common API endpoint I opted to leave it as is now because otherwise clang would complain about it during tests

src/devices/cellular.c
stieg: WILCO.

src/devices/cellular.c
brentpicasso: gotcha

src/devices/cellular.c
stieg: I don't recall why 20 off-hand.  Probably a left-over that should get updated.  If the connection drops at this point we will never auth and the modem will reset and try again.  I can try to optimize a bit here but we need to be careful otherwise we can find ourselves in a loop where we never auth.

src/devices/cellular.c
stieg: I can look into doing this when I add support for multi-socket systems as I will have more knowledge about the cause of failure at that time.  For now its not easily do-able.

src/devices/cellular.c
stieg: It does.  It is used by the connectivityTask IRRC.

src/devices/sara_u280.c
stieg: Will fix.

src/devices/sara_u280.c
stieg: Oops.

src/devices/cellular.c
stieg: Yep.

stieg: Well... it can send multiple commands in one query... but you only get one response back when you chain commands like that.

src/devices/sara_u280.c
stieg: Ok.  Sounds right.

src/gsm/gsm.c
stieg: hehe.

src/gsm/gsm.c
stieg: The failure here is determined by not receiving an OK at the end of the command.  Not sure how a message would help since its not a parsing failure.

src/jsmn/jsmn.c
stieg: I should just f-ing do it now.  Nothing is more permanent than temporary.

stieg: I was tired.

src/logger/loggerApi.c
stieg: Yes.  I forgot to file this.  There is a new json value that has a string representation of the cellular status.  Just need to print it.

src/devices/cellular.c
brentpicasso: If we put a check for the string 'ERROR'  - thiis is what the modem will respond with if we've dropped the connection. Would be good to have as it's an explicit check - even though the as-is code will eventually fall through.

src/serial/serial_buffer.c
stieg: This was adapted from what existed previously.  It strips any non-visible characters.  It is designed for trailing control characters since that is how our serial rx methods parse lines (on \n)

src/serial/serial_buffer.c
stieg: That is a very good point that I did not think of.  The problem will be how to handle URCs in the next implementation of this.  I will review.

src/devices/cellular.c
brentpicasso: I just noticed the auth is sent only once at the top, and the loop tries on just subsequent responses. I wonder if the server can handle multiple auth requests per connection - @ryandoherty ?

src/util/mod_string.c
stieg: I don't recall off-hand actually.  Both are equally valid and I might have just been in a mood at that point :package:

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+        serial->put_c('\n');
+
+        pr_debug_str_msg("sending auth- deviceId: ", deviceId);
+
+        /* Parse the incoming JSON */
+        jsmn_parser parser;
+        jsmntok_t toks[AUTH_MAX_JSMN_TOKENS];
+
+        for (size_t tries = 20; tries; --tries) {
+                const int chars = serial_buffer_rx(sb, 1000);
```
brentpicasso: This timeout should be set much higher - the auth can take multiple seconds to come back. I see it all the time when testing modules.  5000 ms might be a good number, and reducing the 20 to maybe 3 attempts - especially if we're only sending one auth at the top.

src/devices/cellular.c
brentpicasso: So for the moment, a network failure at the instant we send the the auth packet will show as a auth failure. I think it'll be OK edge case since it'll retry quickly right after that.

src/devices/cellular.c
brentpicasso: If it works well I like it; I always assumed we had to send / process one command at a time. :+1:

src/jsmn/jsmn.c
brentpicasso: :+1:

src/logger/loggerApi.c
brentpicasso: Did the interface change? I'm trying to remember if it was sending numeric codes before or if it was sending string values as well.

src/devices/cellular.c
```diff
@@ -0,0 +1,534 @@
...
+
+int32_t cellular_active_time()
+{
+        const tiny_millis_t since = telemetry_info.active_since;
+        return 0 == since ? 0 : (int) getUptime() - since;
+}
+
+const char* cellular_get_net_status_desc()
+{
+        switch(cell_info.net_status) {
```
brentpicasso: The other status codes are returned as numbers and the app decodes it. I think we should remain consistent with this and let the app convert that into textual strings.

src/devices/cellular.c
stieg: Removed

src/devices/cellular.c
stieg: > If we put a check for the string 'ERROR' - thiis is what the modem will respond with if we've dropped the connection. Would be good to have as it's an explicit check - even though the as-is code will eventually fall through. 

I am of the opinion this will have no added value in the end.  I'll just reduce the number of attempts from 20 down to 3.  If it doesn't get an auth response, it will be because of a bad connection and will reset and start over.

src/devices/cellular.c
stieg: Precisely... it would show as an auth failure for a bit until the next retry, which assuming they were still out of range, would lead to a network connection failure and a correct status.  Probably not work correcting now.

src/serial/serial_buffer.c
stieg: For now this won't change because of how we allocate the tasks.  But when we add the wifi support I will incorporate this in.

src/devices/cellular.c
stieg: Ok

src/devices/cellular.c
stieg: Per our discussion we will now use keys.  Code included.

stieg:
@brentpicasso All feedback is now addressed.  Time for another pass over.

stieg:
Feedback applied.

Issue #418, opened 2016-02-09
ZOMG!!! SHIFT!!!! (Fix the impossibly high readout on PWM inputs)


stieg:
When the vehicle gets turned off sometimes the tach likes to read unrealistically high values (like 15,000,000).  Not super good if you have logic that reacts based on these values.  Need to fix this so that this no longer happens.

stieg:
So the solution here is to fix the broken overflow interrupts on the timers.  Once that is fixed, a no signal readout should cause the value to go down to 0.  Also should add an upper limit (of say 20K) to prevent the stupid high readouts.  @brentpicasso has a patch for this somewhere :D

stieg:
#434 should dramatically help the high rpm issue.  See ticket #430

PR #429, opened 2016-02-15
Make the RPM input timers reset themselves when no signal detected

stieg:
Previously the logic that would detect a buffer overflow in the timer
was disabled because of unexpected interrupts occuring.  These were
caused by the UpdateRequestSource flag not being set to local events
only.

This patch sets that value for the timers 0 & 1 and enables the interrupt
code that allows for clearing of the value when overflow hits (ie
when we lose signal from the engine).  Timer 3 still has emotional issues.

Part of the fix for issue #418


brentpicasso:
This looks pretty sane to me.  STM32 voodoo stuff otherwise.

PR #434, opened 2016-02-18
Rpm Software Filtering

stieg:
We have seen issues in the field where voltage echos (aka Fly-back) in the sender wires cause our unit to accidentally register an extra RPM pulse.  Here is [one such example from an Acura Integra](https://www.race-capture.com/events/24-hours-of-lemons-where-the-elite-meet-to-cheat-day-1-south-track/device/team-sheen-181?laps[]=96&laps[]=93).  As you can see, those aren't real values.  Unfortunately, this makes things unusable, so we need a fix.

This patch addresses the issue by adding in a "quiet period" into the IRQs that register the triggers.  If an IRQ occurs during this period, then we save the timer counter value locally, but ignore the pulse.  Then when the real pulse comes in, we are able to accurately calculate the period to within a much narrower window.  Multiple fly-back signals can cause a little bit of noise, but nothing so severe that it can't be overcome.

This is currently designed against 2.8.8 so we can get a solution out to the field sooner (since this is a widespread issue).  In 2.9.0 I will add support for adjusting the quiet period as part of the config.  For now I assume that the user's vehicle will not redline at 20,000 or more.

### DIFF
--- a/src/timer/timer.c
+++ b/src/timer/timer.c
@@ -28,12 +28,22 @@ static Filter g_timer_filter[CONFIG_TIMER_CHANNELS];
 
 int timer_init(LoggerConfig *loggerConfig)
 {
-    for (size_t i = 0; i < CONFIG_TIMER_CHANNELS; i++) {
-        TimerConfig *tc = &loggerConfig->TimerConfigs[i];
-        timer_device_init(i, tc->timerSpeed, tc->mode);
-        init_filter(&g_timer_filter[i], tc->filterAlpha);
-    }
-    return 1;
+        for (size_t i = 0; i < CONFIG_TIMER_CHANNELS; i++) {
+                TimerConfig *tc = &loggerConfig->TimerConfigs[i];
+
+                /*
+                 * Software Filter Hack Represents 20K Max.  For release
+                 * in 2.8.8.  Have to divide value by pulses per revolution
+                 * since this must scale depending on the number of pulses
+                 * per engine rotation.
+                 */
+                const uint16_t qp_us = 3000 / tc->pulsePerRevolution;
+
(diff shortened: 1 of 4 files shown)


src/timer/timer.c
```diff
@@ -26,14 +26,31 @@
...
+        if (MODE_LOGGING_TIMER_RPM != tc->mode)
+                return 0;
+
+        /*
+         * Software Filter Hack Represents 20K Max.  For release
+         * in 2.8.8.  Have to divide value by pulses per revolution
+         * since this must scale depending on the number of pulses
+         * per engine rotation.
+         */
+        return 3000 / tc->pulsePerRevolution;
```
brentpicasso: magic number for now until we have it configurable in 2.9.0?

stm32_base/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -245,107 +238,139 @@ static void init_timer_2(size_t prescaler, unsigned int slowTimerMode)
...
-    return period;
+        return channel < MK2_TIMER_CHANNELS ? g_state[channel].period : 0;
+}
+
+
+/*
+ * = = = IRQ methods below this point = = =
+ */
+
+void update_device_state(const size_t chan, const uint16_t p_ticks,
```
brentpicasso: please make the p_ticks and p_h_ticks variables more verbose, or document the function. That would help

kosenko-max:
Is this going to disallow somehow reading 0-4500Hz turbo speed sensor?

stieg:
No.  Just change the timing mode to Hz instead of RPM.  This will bypass the filter.

kosenko-max:
Cool. Thanks.

PR #438, opened 2016-02-18
Fix the Lua timing issue so that it invokes the onTick method regularly.

stieg:
There was a bug in the code that was effectively like using `sleep(1000)`
at the top of the `onTick` invocation method.  This fixes the execution
order so that we properly invoke onTick on regular intervals, regardless
of how long it took to execute the previous Lua code.


brentpicasso:
:shaved_ice:

PR #439, opened 2016-02-18
Quiet down the lua memory messages

stieg:
Converts them from debug to trace since they are noisy when
debug is active.


brentpicasso:
:sake:

Issue #414, opened 2016-02-08
MAX_VIRTUAL_CHANNELS = 30; is too small


kosenko-max:
Hi. Sorry for posting here...

But I've found significant problem that is not letting me to do Evo X CAN BUS support. Standard RAX Patch logger returns 8 groups with 4 numbers in each. That's already 32 - above limit of virtual channels. Plus I have some more data outside of RAX: MAF Airflow, WGDC corrections, transmission pressure/temperature/rpm per each gearset, knockdata. Plus I'd like to read wheels speed, brake signal, steering signals from engine CAN. So I literally need much more virtual channels. And I didn't even start to use any calculated data for the primary purpose of Virtual Channels.

I don't understand from the code why limit of virtual channels on MK2 was set to 30.
[https://github.com/autosportlabs/RaceCapture-Pro_firmware/blob/aef0769b5539f0647a76b3a00b1774ba29a8c343/stm32_base/capabilities.h#L14
](url)

Could you move that up to 100 at least or should I do Evo X support as a custom firmware? 

Thank you.

stieg:
I don't recall off-hand why we set that limit either.  Probably for historical reasons (MK1).  We may be able to bump this up now.  Just need to evaluate impact first.

stieg:
> Hi. Sorry for posting here...

Why?  Its where it belong. :D

kosenko-max:
Is it possible to somehow get firmware with this change before release? I need to populate 74 channels http://pastebin.com/02i5Dp6L and this limit is really not letting me to do any tests.

Thank you.

stieg:
Hey @kosenko-max,

   We don't usually provide releases ahead of time for several reasons (stability, QC, debugging etc).  But if you were so inclined you could build it yourself once the patch was in the build tree.  Of course you would be on your own in terms of support (since you would not be running an official version) but I assume you already know that.

kosenko-max:
It's ok, I thought maybe you have some automated build system which you can point onto 2.3.7 with DEFINE patch only and build a firmware file for testing.

I don't have build environment yet - mostly doing Lua for now to integrate Evo X CAN.

stieg:
Yeah we don't have a public build system with nightly releases yet.  Its on
the list of things to get done.

On Fri, Feb 12, 2016 at 10:56 PM, Max <notifications@github.com> wrote:

> It's ok, I thought maybe you have some automated build system which you
[...]

PR #440, opened 2016-02-19
Change Max Virtual Channels from 30 to 100

kosenko-max:
Based on #414 request propose to let MK2 use more channels and see how it works.

### DIFF
--- a/stm32_base/capabilities.h
+++ b/stm32_base/capabilities.h
@@ -11,7 +11,7 @@
 //configuration
 #define MAX_TRACKS				240
 #define MAX_SECTORS				20
-#define MAX_VIRTUAL_CHANNELS	30
+#define MAX_VIRTUAL_CHANNELS	100
 #define LOGGER_MESSAGE_BUFFER_SIZE	10
 
 /*


stieg:
ok to test

stieg:
Looks sane to me.  Doesn't interfere with the `LoggerConfig` struct in any way.  Ideally this would go in 2.9.0 but should be able to work in 2.8.8.  Only thing I would want to do before merging this would be to test it locally.

stieg:
Tested this locally.  While I am able to add 100 channels and have things work locally, I have stumbled onto another bug.  File Writing isn't working and the ChannelSample type isn't getting populated like it should.

@kosenko-max Since this is a new feature (expansion of virtual channel support) lets target this towards 2.9.0 instead of 2.8.8 (which is a bugfix release).  I'll build upon your patch and get this into 2.9.0.  Thanks for the submission.

kosenko-max:
I'm still unable to make build system to setup logging on my car. If you see that 100 channels is a feature and File Writing or ChannelSample bugs are not in effect with 30 channels - yes, it's probably 2.9.0. Although I'm not sure when it will come.

stieg:
> I'm still unable to make build system to setup logging on my car. If you see that 100 channels is a feature and File Writing or ChannelSample bugs are not in effect with 30 channels - yes, it's probably 2.9.0.

stieg:
Dang it.  ok to test.

stieg:
Hey @kosenko-max,
   I hear ya.  I don't know why its breaking off-hand either.  Definitely a regression though with the file writing.  I'll look at patching up the vagrant image so that you can build with it.  Then, if you are so inclined, you can build your own version of the firmware before its released to play with.

stieg:
Solved the bug.  See #444

stieg:
Just tested this with a workaround to the issue... works fine as far as I can tell.

stieg:
Since this will go against 2.9.0, I have re-opened another PR with patches from this PR rebased onto r/2.9.0.  See PR autosportlabs/RaceCapture-Pro_firmware/pull/446.  Going to close this one.

Issue #444, opened 2016-02-21
File Logging Completely If Virtual Channel With Sample Rate of 50Hz created


stieg:
If you add a virtual channel in the Lua environment with a sample rate of 50hz, then logfile writing will no longer function.  It looks like its functioning, but no data will get written and no errors will be reported.  Even the Big Red Light will fail to work.  How to reproduce:

1.  Start with a clean RCP with a reset config
2.  Ensure logging to sdcard is working
3.  Add the following Lua script to the Lua environment:

 ```lua
 -- Adds one virtual Channel
 addChannel("Foo", 10)
 function onTick() end
 ```
4.  Try logging.  Won't work.

Seems like this regression has been around for a while as I see it on 2.8.4 firmware.

stieg:
Regression began in the 2.8.4 release.  2.8.3 was operational.

stieg:
RCA: we aren't sending messages to the fileWriter.  This is happening because we are not including the virtual channels in our sample rate calculations.

stieg:
Merged

PR #445, opened 2016-02-21
Factor the sample rates of virtual channels into our calculations

stieg:
It would seem that we forgot to include virtual channels into our
sample rate values.  As a result, logging to the fileWriter was
failing to work.  This ticket fixes that issue by factoring in
virtual channels when we calculate those values.

Addresses issue #444


brentpicasso:
looks sane.  shipit

PR #448, opened 2016-02-22
Make Logging a top priority

stieg:
Acquiring a data snapshot at the precise time is of high importance
in a telemetry system.  If any one unit is able to prevent the logger
from aqcuiring its information at the precise moment, then the logger
data could be stale and lost.

Thus to prevent this scenario, we need make the logger task have the
highest priority in the system.  When this is the case, the logger
will always trump any other task needing to run when the preemptive
scheduler is invoked.  Thus there is no chance in missing an update.
a data snapshot is determined when our


brentpicasso:
:shipit: This will get generally wrung out when we test the 2.9.0 branch as a whole.

PR #449, opened 2016-02-22
Support faster onTick, Improve setTickRate method

stieg:
To be able to play better with customers who want faster Lua rates, I am adjusting our setTickRate method to allow higher invocation rates.  This is sane because Lua is our lowest priority process (only idle task is lower).  So it will never block a task that isn't more important.

This patch also improves the setTickRate to return the time (in ms) it will sleep between onTick invocations and enables the ability to send back errors to Lua when folks invoke the commands improperly.

### DIFF
--- a/include/lua/luaTask.h
+++ b/include/lua/luaTask.h
@@ -38,7 +38,7 @@ void* getLua(void);
  */
 size_t get_lua_mem_size();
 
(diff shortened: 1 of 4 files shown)


brentpicasso:
looks sane. :shipit:

Issue #424, opened 2016-02-11
addChannel("Something", 30) hangs Lua silently


kosenko-max:
addChannel("Something", 30) hangs Lua silently - it took a lot of time to find out what's going on since there were no messages and system was not even accepting code updates and constantly restarting.

stieg:
Well that is annoying.  Will investigate.

brentpicasso:
I think it relates to the 30 of the sample rate you specified. Does it
happen with one of the supported sample rates?

https://autosportlabs.net/RaceCapturePro_Lua_Scripting#addChannel.28_name.2C_sampleRate.2C_.5Bprecision.5D.2C_.5Bmin.5D.2C_.5Bmax.5D.2C_.5Bunits.5D_.29

It looks like it returns a disabled sample rate, which may be causing the
hang downstream. Looks like we need to better validate the specified sample
rate and fail the addChannel() function as necessary.

On Thu, Feb 11, 2016 at 7:33 AM, Andrew Stiegmann <notifications@github.com>
wrote:

> Well that is annoying. Will investigate.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/issues/424#issuecomment-182918424>
> .
>



-- 
Brent Picasso
*Autosport Labs <http://www.autosportlabs.com>*
*Technology for Race and Street*

kosenko-max:
Do you actually need any update rate specified on a virtual channel? I struggle to understand any meaning of it at all...

stieg:
Yes you do.  The logging mechanism takes samples at X hz.  When you update a rate on a virtual channel, that will update the sample value that will be read by the logger.  Since these are independent processes you need to tell the logging mechanism how often you want the sample updated in the telemetry stream.  It assumes that you will be updating it at that rate.  Otherwise it just sends the previous value that it got.

kosenko-max:
I see. Once you log anything at 1000Hz - is it actually going to make any difference that other channels  are not that fast? You still need to try to make 1000 lines a second, so you save only on reading values from the table of channels. But channels anyway updated at their own pace.

It looks to me that having a single log rate for everything and adjust rest of the stuff to that rate as much as possible would be way more clear approach for some future versions. It makes sense to give some priorities for CAN requests since they fight for same bandwidth. Well, I still didn't checked which processor you use, so maybe that could be a limit then?

stieg:
[...]

PR #451, opened 2016-02-22
P/stieg/validate lua add channel

stieg:
Adds validation to the lua addChannel method.  Adds code that allows for centralized validation and creation of sane default ChannelConfig structs.  Adds tests to make it all work.

Addresses Issue #424

### DIFF
--- a/src/logger/luaLoggerBinding.c
+++ b/src/logger/luaLoggerBinding.c
@@ -26,11 +26,13 @@
 #include "LED.h"
 #include "OBD2.h"
 #include "PWM.h"
+#include "channel_config.h"
 #include "dateTime.h"
 #include "gps.h"
 #include "imu.h"
+#include "luaBaseBinding.h"
 #include "lap_stats.h"
-#include "lap_stats.h"
+#include "lauxlib.h"
 #include "logger.h"
 #include "loggerConfig.h"
 #include "loggerData.h"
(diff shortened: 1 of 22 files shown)


src/logger/luaLoggerBinding.c
```diff
@@ -782,29 +784,83 @@ int Lua_FlashLoggerConfig(lua_State *L)
...
+                cc.min = lua_tointeger(L, 4);
+        case 3:
+                cc.precision = lua_tointeger(L, 3);
+        }
+
+        switch(validate_channel_config(&cc)) {
+        case CHAN_CFG_STATUS_OK:
+                break;
+        case CHAN_CFG_STATUS_MAX_LT_MIN:
+                return luaL_error(L, "Max is less than min");
```
brentpicasso: I see some inconsistent uses of luaL_error - here you're returning the result and in other places above you don't. Should they all return as on this line?

src/logger/luaLoggerBinding.c
```diff
@@ -782,29 +784,83 @@ int Lua_FlashLoggerConfig(lua_State *L)
...
-    lua_pushinteger(L, create_virtual_channel(cc));
+        const size_t args = lua_gettop(L);
+        if (args < 2 || args > 6)
+                return incorrect_arguments(L);
 
-    return 1;
+        const char *label = lua_tostring(L, 1);
+        switch (validate_channel_config_label(label)) {
+        case CHAN_CFG_STATUS_OK:
+                break;
```
brentpicasso: suggest setting the label text on this line so it's consistent with how you do it on line 824

src/logger/luaLoggerBinding.c
stieg: Ahh... so that is an idiom.  `luaL_error` never actually returns.  So me using a `return` statement is designed to make more logical sense.  I'll clean up the code to always use returns.

brentpicasso: I thought that was a case. Agree on the idiom.

stieg: I should clarify too that by "it never returns" I mean that the lua script will continue if possible, it just long jumps out of the C code.  See the luaL_error method description for more detailed info.

brentpicasso: long jump = nuclear option!

brentpicasso:
This looks like a great improvement! fantastic work.

stieg:
@brentpicasso is this good to go in?

PR #454, opened 2016-02-25
[RFC] AT Command Engine

stieg:
This patchset has a working version of the new AT command engine along with all unit tests.  This engine is what we will use to handle all commands going to and from the various communication devices we support.  WiFi & Cellular will make use of it.  It implements asynchronous call backs so none of the commands are blocking.  It expects the `at_task` method to be invoked frequently as this is what checks for incoming data as well as invokes sends.

This is not intended for merge.  Rather this is a forum for feedback on the code itself.  The full WiFi patch will include all you see here plus the driver for the WiFi chip and the task management.

### DIFF
--- a/src/modem/at.c
+++ b/src/modem/at.c
@@ -37,6 +37,7 @@ static const struct at_rsp_status_msgs {
         AT_STATUS_MSG("OK", AT_RSP_STATUS_OK),
         AT_STATUS_MSG("FAILED", AT_RSP_STATUS_FAILED),
         AT_STATUS_MSG("ABORTED", AT_RSP_STATUS_ABORTED),
+        AT_STATUS_MSG("ERROR", AT_RSP_STATUS_ERROR),
 };
 
 static bool is_timed_out(const tiny_millis_t t_start,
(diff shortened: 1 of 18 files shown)


src/modem/at.c
```diff
@@ -0,0 +1,414 @@
...
+        case AT_RX_STATE_URC:
+                /* We are in the middle of receiving a unexpected message. */
+                process_urc_msg(ati, msg);
+        case AT_RX_STATE_READY:
+                /*
+                 * We are starting a new message series.  Gotta figure out what
+                 * type of message this is before we process it.  Then we process
+                 * it appropriately.
+                 */
+                ; /* C is stupid after labels.  Workaround :\ */
```
Jeff-Ciesielski: If you start a bracketed block here after the label, this problem goes away.

stieg: Good to know.  I'll give it a shot and see if it helps.  Seems like a brain dead C ism IMHO... but what are you going to do.

Jeff-Ciesielski: Yeah, I think it's just a weird visage of how C handles lexical scoping and the differences between C89/99.

stieg:
Got the feedback I wanted.  Thanks.  Closing.

Issue #426, opened 2016-02-12
Fix USB autosuspend bug in MK2


stieg:
Turns out that there is a bug in the autosuspend feature of the USB stack on MK2.  If the units USB gets put into suspend mode it will cause the unit to never recover.  This makes the device unusable until you unplug and re-plug the USB cord.

On Linux systems you can work around the issue by pasting the following udev rule into `/etc/udev/rules.d/99-asl-rcp-mk2.rules`

```
SUBSYSTEM=="usb", ACTION=="add", ATTRS{idVendor}=="16d0", ATTRS{idProduct}=="07f1", ATTRS{version}==" 2.00", ATTR{power/autosuspend}="-1"
```

That will cause sysfs to never put the device into autosuspend mode, thus side-stepping the issue.  This ticket tracks fixing the USB code so we can work with USB autosuspend.

PR #458, opened 2016-02-29
Remove de-init code in USB driver when USB gets suspended

stieg:
This causes the device to never return to an awake state when the
USB subsustem goes into sleep mode.  The simple fix here is to
remove the call since power consumption is not a high priority for
us.  The better fix would be to fix it so that we return from sleep
when we move away from suspend.  Patches welcome.

Fixes issue #426

### DIFF
--- a/stm32_base/hal/usb_stm32/usbd_cdc_vcp.c
+++ b/stm32_base/hal/usb_stm32/usbd_cdc_vcp.c
@@ -183,23 +183,44 @@ static bool check_tx_overrun(void)
     return false;
 }
 
-static bool check_suspended(void)
+/**
+ * @return true if the device is in suspended mode, false otherwise.
+ */
+static bool is_usb_suspended(void)
 {
-    /* Checks to see if the USB bus is suspended */
-    if (USB_OTG_dev.regs.DREGS->DSTS & 0x1)
-        return true;
-
-    return false;
+        /* Checks to see if the USB bus is suspended */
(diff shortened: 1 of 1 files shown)


brentpicasso:
awesome. :shipit:

stieg:
Well @Jeff-Ciesielski is sending me warning about doom with this because of ISR storms when no USB is connected.  So need further testing and validation before I merge this.

stieg:
The double not clamp is to ensure that its treated as a bool.  Its probably not necessary.  I can get rid of it as needed.

Jeff-Ciesielski:
Lack of a true bool type has always been a bit of an annoyance in C.  On the one hand, I could ask the question "who would do something like decrement/increment/modify something marked bool".  On the other, I've actually seen someone perform math on a bool causing some really subtle bugs :(.

I can see the benefit, but do you think perhaps a "CLAMP_TO_BOOL" macro would make sense for readability?

stieg:
> I've actually seen someone perform math on a bool causing some really subtle bugs :(.

Oh god.

> I can see the benefit, but do you think perhaps a "CLAMP_TO_BOOL" macro would make sense for readability?

Sure.  Can easily do that.

brentpicasso:
Looks sane. I've never reproduced this, so I'll just test the normal operation on my kernel after this gets merged to ensure everything worked as before.

:shipit:

Issue #435, opened 2016-02-18
Slow timer / fast timer calculations off


brentpicasso:
When calculating uSec, need to account properly for various prescaler settings. 

Medium timer mode calculations is accurate  ( 1 signal : 1 measured)
Fast measures 1/2 of actual  (1 signal : 0.5 measured)
Slow measures 10x of actual (10 signal : 1 measured)

PR #460, opened 2016-02-29
Fix Timer pre-scalar skew in MK2 HAL layer

stieg:
Turns out that we were not correctly adjusting our ticks per
microsecond (us) calculation when the prescaler changed.  This was
resulting in incorrect period readings of 1/2 the speed in fast mode,
and 10x the value in slow mode.  This patch addresses that issue and
consolidates the code a bit more to make it more readable and
understandable.

Fixes issue #435

### DIFF
--- a/stm32_base/hal/timer_stm32/timer_device_stm32.c
+++ b/stm32_base/hal/timer_stm32/timer_device_stm32.c
@@ -292,19 +292,19 @@ void reset_device_state(const size_t chan)
         s->q_period_ticks = 0;
 }
 
-bool timer_device_init(const size_t channel, const uint32_t speed,
+bool timer_device_init(const size_t chan, const uint32_t speed,
                        const uint16_t quiet_period_us)
 {
-        if (channel >= MK2_TIMER_CHANNELS)
+        if (chan >= MK2_TIMER_CHANNELS)
                 return false;
 
-        struct config *c = &g_config[channel];
-        c->prescaler = speed_to_prescaler(channel, speed);
+        struct config *c = &g_config[chan];
+        c->prescaler = speed_to_prescaler(chan, speed);
         c->q_period_us = quiet_period_us;
 
-        reset_device_state(channel);
+        reset_device_state(chan);
 
-        switch (channel) {
+        switch (chan) {
         case 0:
                 init_timer_0(c);
                 return true;
@@ -320,9 +320,9 @@ bool timer_device_init(const size_t channel, const uint32_t speed,
         return false;
 }
 
-void timer_device_reset_count(size_t channel) {}
+void timer_device_reset_count(size_t chan) {}
 
-uint32_t timer_device_get_count(size_t channel)
+uint32_t timer_device_get_count(size_t chan)
 {
         return 0;
 }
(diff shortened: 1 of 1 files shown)


stm32_base/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -309,15 +327,16 @@ uint32_t timer_device_get_count(size_t channel)
...
+static uint32_t ticks_to_us(const size_t chan, const uint16_t ticks)
 {
-        /* 1 tick == 2us */
-        return (uint32_t) ticks << 1;
+        /* Our pre-scalars are setup to never have less than 1 tick per us */
+        const uint32_t us_per_tick = get_clk_speed(chan) / 1000000;
+        return us_per_tick ? g_config[chan].prescaler * ticks / us_per_tick : 0;
 }
 
 uint32_t timer_device_get_usec(size_t channel)
```
brentpicasso: nit. inconsistent use of chan or channel throughout code

stm32_base/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -309,15 +327,16 @@ uint32_t timer_device_get_count(size_t channel)
...
 }
 
-uint32_t ticks_to_us(uint16_t ticks)
+static uint32_t ticks_to_us(const size_t chan, const uint16_t ticks)
 {
-        /* 1 tick == 2us */
-        return (uint32_t) ticks << 1;
+        /* Our pre-scalars are setup to never have less than 1 tick per us */
+        const uint32_t us_per_tick = get_clk_speed(chan) / 1000000;
+        return us_per_tick ? g_config[chan].prescaler * ticks / us_per_tick : 0;
```
brentpicasso: should this static function do any parameter validation or do we think the higher level functions are taking care of it? thinking of the array index.

stm32_base/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -271,21 +299,11 @@ bool timer_device_init(const size_t channel, const uint32_t speed,
                 return false;
 
         struct config *c = &g_config[channel];
-        c->prescaler = speed_to_prescaler(speed);
+        c->prescaler = speed_to_prescaler(channel, speed);
```
brentpicasso: what's the risk of setting the prescaler to 0 if we've passed in an invalid channel? The timer may misbehave?

brentpicasso:
Looks good, a few comments about variable checking and other nits.

stm32_base/hal/timer_stm32/timer_device_stm32.c
stieg: Yep.  It will go way to fast and we will notice it as the timer seeming to never work.

stm32_base/hal/timer_stm32/timer_device_stm32.c
stieg: I was leaving that to the higher level methods.  The way I see it, methods designed to be invoked from other c files should do the validation.  Otherwise it becomes redundant.

stieg:
All comments addressed.  Good to go @brentpicasso ?

Issue #456, opened 2016-02-28
Enhance Virtual Channel methods to include getChannel ability


kosenko-max:
Currently it's cumbersome to get data from channels - you have to make local variables at the time of update. It would be cleaner code if getChannel would read Virtual Channel data.

stieg:
Hey @kosenko-max, Would you provide an example?  I am afraid I'm not quite following you here.  TIA.

kosenko-max:
IDC calculation requires to remember IPW in one place and come to RPM in
another place. It can be done by storing last IPW in variables, but that
increases code size and makes it inconsistent with other log data access
methods.

stieg:
Got it.  Will look into it.

stieg:
New method now [documented on the wiki](https://autosportlabs.net/RaceCapturePro_Lua_Scripting#getChannel.28_.3CChannel_ID_.7C_Channel_Name.3E_.29  Co).  Closing

PR #461, opened 2016-03-01
Add a get_virtual_channel method

stieg:
To help reduce the number of globals that must be kept in the Lua
runtime, this method allows devs to grab virtual channel values so
they can create meta-channels using data from other virtual channels
without having to store all the channel values in global lua runtime
space.

Fixes issue #456 

Note... I am merging this against 2.8.8 because that branch has patches I wanted
to use in this commit.  It will be part of 2.9.0 release since we are going to fold 2.8.8
into the 2.9.0 release.

### DIFF
--- a/src/logger/luaLoggerBinding.c
+++ b/src/logger/luaLoggerBinding.c
@@ -60,8 +60,9 @@
 #define LUA_DEFAULT_SERIAL_PARITY 0
 #define LUA_DEFAULT_SERIAL_STOP_BITS 1
 
-
 char g_tempBuffer[TEMP_BUFFER_LEN];
+static int lua_get_virtual_channel(lua_State *ls);
+
 
 static int lua_get_uptime(lua_State *L)
 {
@@ -161,6 +162,7 @@ void registerLuaLoggerBindings(lua_State *L)
     lua_registerlight(L,"getBgStream", Lua_GetBackgroundStreaming);
 
     lua_registerlight(L, "addChannel", Lua_AddVirtualChannel);
+    lua_registerlight(L, "getChannel", lua_get_virtual_channel);
     lua_registerlight(L, "setChannel", Lua_SetVirtualChannelValue);
 
     /* Timing info */
@@ -848,6 +850,36 @@ int Lua_AddVirtualChannel(lua_State *L)
 bug:
         return luaL_error(L, "BUG. Should never get here.  "
                              "Please inform AutosportLabs");
+}
+
+static int lua_get_virtual_channel(lua_State *ls)
+{
+    if (lua_gettop(ls) != 1)
+            return incorrect_arguments(ls);
+
+    const bool is_num = lua_isnumber(ls, 1);
+    const bool is_str = lua_isstring(ls, 1);
+    if (!is_num && !is_str)
+            return luaL_error(ls, "This method only accepts channel names "
+                              "or channel IDs");
+
+    /*
+     * Numbers can be passed in as 123 or "123" in lua.  So handle
+     * that case first
+     */
+    VirtualChannel *vc;
+    if (is_num) {
+            vc = get_virtual_channel(lua_tointeger(ls, 1));
+    } else {
+            const int idx = find_virtual_channel(lua_tostring(ls, 1));
+            vc = get_virtual_channel(idx);
+    }
+
+    if (!vc)
+            return luaL_error(ls, "Virtual channel not found!");
+
+    lua_pushnumber(ls, vc->currentValue);
+    return 1;
 }
 
 int Lua_SetVirtualChannelValue(lua_State *L)


brentpicasso:
looks sane. :shipit:

makes me think we could have better unit tests around the lua stuff...

stieg:
We could do that somewhat, but would have to stub out the lua_State object to do it.

kosenko-max:
Cool - channel by name. That would be extremely useful for my script rewrite in attempt to fit 100 channels into memory https://gist.github.com/kosenko-max/c5cb086292c287e45004

stieg:
It supports both name and index.

On Tue, Mar 1, 2016 at 12:02 PM, Max <notifications@github.com> wrote:

> Cool - channel by name. That would be extremely useful for my script
> rewrite in attempt to fit 100 channels into memory
> https://gist.github.com/kosenko-max/c5cb086292c287e45004
>
> —
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/461#issuecomment-190835034>
> .
>



--

kosenko-max:
Thanks a lot for doing it flexible.

stieg:
Using it by index would be best (because it doesn't have to look up the index of the channel associated with the name), but it will work with a name too (just slower).

kosenko-max:
I have a suggestion - Lua actually always generalizes all strings. So same
string always same object. And usage of strcmp over whole array of channels
looks very excessive - comparing pointers should be more than enough. And
speed would not degrade that much.

I can't really use indexes, because I don't know them. And that would force
me to store links to them.

kosenko-max:
I'm talking about https://github.com/autosportlabs/RaceCapture-Pro_firmware/blob/26ab29d3fd52cfe94112f21bc0310747c9ec754d/src/virtual_channel/virtual_channel.c#L44

stieg:
> Lua actually always generalizes all strings. So same string always same object. And usage of strcmp over whole array of channels looks very excessive - comparing pointers should be more than enough. And speed would not degrade that much.

As I understand it we can't do that because these parameters are passed from the Lua environment to the C environment via the Lua stack.  That stack is only valid for the length of the call, so any C string on there (or referenced by a pointer there) could be relocated without warning.

> I can't really use indexes, because I don't know them. And that would force me to store links to them.

You already store them.  They are the values returned when you create a channel.  What is returned is the index of the virtual channel

kosenko-max:
I'm rather confused - are you saying that whole Lua heap memory is
reallocated between calls? I somewhat doubt it. Lua keeps all strings in
one hashtable and verifies that same string is always same pointer. So if I
register channel with some name and later request value of it by name - it
would be indeed same pointer.

In a new script I don't store indexes of channels anymore.

stieg:
> I'm rather confused - are you saying that whole Lua heap memory is reallocated between calls?

No.  Rather lua provides a stack for each call into C land that you can read parameters from and push parameters to when returning values.

> Lua keeps all strings in one hashtable and verifies that same string is always same pointer.

Sure.  But does that pointer always reflect the same address?  What happens if the Lua GC compacts the space?  Is it not possible that string could move in memory to allow for more efficient compaction?  I know its true in the Java world, but my familiarity with the Lua interpreter is not as deep.

> So if I register channel with some name and later request value of it by name - it would be indeed same pointer.
> In a new script I don't store indexes of channels anymore.

When you say value, do you mean the value last set for the channel, or the value of the channel index?  I'm not sure how you are going to be able to invoke `setChannel` without maintaining a variable that allows you the virtual channel id (What I have been referring to as the index, because that is what it is under the hood).

kosenko-max:
I would need to check Lua sources to see if pointer can move due to GC.

By value I mean data value. And for me there is a difference of source data
I use - channels uploaded all in one cycle vs calculated channels for which
link is necessary.

Anyway, I can find how to overcome slow find channel logic by storing
channel pointer in the channels definition array and some hashtable. But I
though it worth checking if strcmp is overkill or it could use compare by
string hash at least.

stieg:
> By value I mean data value.

Ok.  Glad we are on the same page there.

> And for me there is a difference of source data I use - channels uploaded all in one cycle vs calculated channels for which link is necessary.

Right... this make sense as to why you asked for the `getChannel` method.  My thought process there is since you already have a channel reference somewhere in the lua code (since you need it for `setChannel`), you can just pass that into the `getChannel` call to get the channel value.  I am having a hard time understanding the inconvenient nature of it since you have to keep a reference to the virtual channel anyways in order to be able to use `setChannel`.  If you feel there are too many variables, perhaps use a single global table to look up channel names?

> But I though it worth checking if strcmp is overkill or it could use compare by string hash at least.

It is unfortunately.  But its what we have because I haven't yet added hash table functionality.  I meant to do it for this release, but since this one is wrapping up, I probably will push it back.

kosenko-max:
It's OK, I can do it by the pointer.

Anyway my intent is to make universal Lua CAN messenger and suggest you to
port it into firmware method once it works Lua or extend planned CAN mapper
to request-response based option. I don't see any other options how I could
fit into memory otherwise.

Issue #436, opened 2016-02-18
Enhance the Version Information in Firmware


stieg:
Currently our firmware version is only 3 digits (X.Y.Z).  This works for production level, but lacks features like the ability tell if the firmware is an official release build or if its something that is locally built or experimental.  Effectively we don't have enough information.

To fix this I propose we include the following:
* A flag which tells us if the build is a DEV, EXPERIMENTAL, or OFFICIAL.  This would be controlled by the Make system
* A spot for us to include the SHA-1 hash of the git commit so we can know exactly what point in the tree the firmware came from.

PR #463, opened 2016-03-02
P/stieg/improved version and make

stieg:
Add very useful version info into the firmware including git descriptions and build type.  Also fixes years of crustiness in our Makefiles and makes the top level one better, but far from best.  Now you can build all of our projects from the TLD.  Its magical.  Also cleans up version.mk and moves it into the TLD Makefile.

I left the ./do_release.sh bits around for now.  These will go away at a later time and date.  For now its simply a wrapper to execute the makefile with the `package` target (along with parallelism).

Fixes issue #436


brentpicasso:
Looks good. I trust your testing. 

Could you add an issue to the 1.5.0 RC App to show the additional status info this firmware is providing?

Thanks, 
Brent

stieg:
No problem

Issue #437, opened 2016-02-18
telemetry and GPS LEDs are swapped


brentpicasso:

stieg:
Man... this turned out to be a bigger project than I thought.  But now we get consistent behavior across all platforms.

stieg:
Issue merged

PR #465, opened 2016-03-03
Fix up the LEDs

stieg:
This patch fixes a long standing annoying issue where we had issues controlling which LEDs were getting activated because they would change between platforms.  This solves this by declaring an enum type to represent each light we know about.  We then map that to the appropriate light in the
driver for the device to ensure proper behavior.  We also maintain the classic ability to set by index (used by Lua).  The lights are read left to right, top to bottom, with the red light always as light 0.  For RCT, the pattern follows the Red, Green, Blue strategy for indexing.

Also fixes a memory corruption bug that could happen if a user passed in an out of bounds index.


stieg:
Fixes #437

main.c
```diff
@@ -80,19 +80,19 @@ static void fatalError(const enum fatal_error type)
         taskDISABLE_INTERRUPTS();
 
         for(;;) {
-                LED_enable(1);
-                LED_enable(2);
+                led_enable(LED_GPS);
```
brentpicasso: if we have a fatal error we should be at least flashing the error LED as well. maybe flash all LEDs?

src/OBD2/OBD2_task.c
```diff
@@ -36,7 +36,7 @@
 
 #define OBD2_TASK_STACK 	100
 
-#define OBD2_FEATURE_DISABLED_DELAY_MS 2000
+#define OBD2_FEATURE_DISABled_DELAY_MS 2000
```
brentpicasso: Your refactoring tools are lacking.

src/LED/led.c
```diff
@@ -0,0 +1,81 @@
...
+
+/*
+ * LED names are the same strings as the enums in code.
+ */
+#define LED_NAME_ENTRY(a)	{#a, a}
+static struct led_names {
+        const char *name;
+        const enum led led;
+} led_names[] = {
+        LED_NAME_ENTRY(LED_ERROR),
```
brentpicasso: Where are the string representation of the LED names defined?

main.c
stieg: You are talking something like a panic function eh.  Not unlike what Linux does.  That would be a wise idea as I have had need for this in code.

src/OBD2/OBD2_task.c
stieg: oops :P.  Will fix

src/LED/led.c
stieg: Same as the Enums.  That is what the `#` character does in pre-processor.  So for the error led the string is "LED_ERROR".

main.c
brentpicasso: I'd say in the short term add the error LED as part of the grouping.

src/LED/led.c
brentpicasso: That's clever, but it could change the API through an innocent refactoring. What about decoupling it? 

Kind of reminds me the risks of externalizing the numeric enums into our API.

stieg: Agreed.  Will fix.

stieg:
This good to go @brentpicasso

src/util/panic.c
```diff
@@ -0,0 +1,99 @@
...
+                delay_seconds(FLASH_DELAY_S);
+
+                for (int c = 0; c < cause; ++c) {
+                        led_enable(LED_GPS);
+                        led_enable(LED_LOGGER);
+                        delay_seconds(FLASH_DELAY_S);
+                        led_disable(LED_GPS);
+                        led_disable(LED_LOGGER);
+                        delay_seconds(FLASH_DELAY_S);
+                }
```
brentpicasso: thought we'd also flash the error led on panic?

stieg: Oh... I just turned it on solid.  The green lights flash an encoded message.

brentpicasso: ok!

brentpicasso:
:ship:

Issue #464, opened 2016-03-03
Further enhancement to RPM filtering


brentpicasso:
Based on customer testing feedback:

For 2.8.8
* In RPM only mode
  * Remove dividing duration based on PulsePerRevolution. Rationale: ignition systems attempt a constant dwell time throughout the RPM range, so in some cases we may end the filter too early.  Also, makes it simpler for users to understand the filter length if it is a constant value.
  * Have a sanity filter that ignores RPM values that are 2x the max channel config value. This will filter out any additional spurious signals.

For 2.9.0
Add config values for filter length (uS) and rising / falling edge flags
Have config default be 4.5 ms, falling edge

stieg:
> Remove dividing duration based on PulsePerRevolution. Rationale: ignition systems attempt a constant dwell time throughout the RPM range, so in some cases we may end the filter too early.

This doesn't sound right.  Logic:

* On a V8 that red lines at 8K, you get 4 sparks per revolution.
* Assume that you are revving it at 8K
* 4 sparks per revolution means we are seeing a pulse rate equivalent to 32K if there was a single pulse per revolution.  If I didn't scale the quiet period based on this info, their RPM readings would clip once they got above 5K (since the current max is 20K RPM).

The solution here is to use the "max" value set by the user.  Assume the max RPM is there, then we can calculate the absolute minimum value we can use.  Easier to use and sane/logical.  No extra field required.

> Add config values for filter length (uS) and rising / falling edge flags

Happy to add the edge trigger setting.  Filter length I don't want to expose.  Lets use the above calculation?

vtjballeng:
Smart coils and ignitors will vary dwell with RPM and noise may not necessarily happen within a fixed filter length window.  Having a variable filter length based on RPM is advisable.  The length should be also divisible based on the pulses per revolution .

Example math.  Let's just say you choose 5ms at 6,000rpm.  Per cylinder, that is 60000/6000 = 10ms, so you are leaving half the period open for a valid signal.  Taking this further and scaling appropriately, at 2,000rpm your filter length would be 60000/2000 = 30ms/2 = 15ms.

If you don't scale this based on the pulses per revolution, you will either drop the signal or allow noise.  

If your sanity filter is a simple cut for 2x the max value, it will only solve a small percentage of the problem.

[...]

PR #467, opened 2016-03-06
Timer Improvements

stieg:
Lots of stuff happening here.  Here is the main gist (which is how the commits are broken up):

* Added in dynamic RPM filter calculations.  This generates a quiet period based on the maximum expected value of the RPM field along with the pulse per revolution value and some other constants.
* Expanded filtering to all timing aspects
* Added the ability to control whether or not we trigger on the rising edge or falling edge of the signal
* Added the ability to manually set the quiet period of the timer, or disable it, or set it to auto.  The values are > 0, 0, and < 0 respectively.
* Moved the timerConfig to its own file so its easier to work on

Addresses issue #464 and then some.


include/timer/timer_config.h
```diff
@@ -0,0 +1,64 @@
+/*
```
brentpicasso: we have the timer config in include/timer/timer_config.h, but the main config is still in include/logger. maybe we should have a top level config directory to organize the location of configs?

include/timer/timer_config.h
```diff
@@ -0,0 +1,64 @@
...
+#define MODE_LOGGING_TIMER_RPM		0
+#define MODE_LOGGING_TIMER_FREQUENCY	1
+#define MODE_LOGGING_TIMER_PERIOD_MS	2
+#define MODE_LOGGING_TIMER_PERIOD_USEC	3
+
+#define TIMER_SLOW	0
+#define TIMER_MEDIUM	1
+#define TIMER_FAST	2
+
+#define TIMER_FILTER_VALUE_AUTO	-1
```
brentpicasso: white space

src/logger/loggerApi.c
```diff
@@ -1091,21 +1091,26 @@ static void getTimerConfigs(size_t channelId, void ** baseCfg, ChannelConfig **
...
 {
     TimerConfig *timerCfg = (TimerConfig *)cfg;
 
-    int iValue = modp_atoi(value);
-    if (NAME_EQU("st", name))
-        timerCfg->slowTimerEnabled = (iValue != 0);
-    if (NAME_EQU("mode", name))
-        timerCfg->mode = filterTimerMode(iValue);
-    if (NAME_EQU("alpha", name))
+    if (!strcmp("mode", name))
```
brentpicasso: we should make all of these == 0.  the  not (!) while technically correct, has a higher cognitive load because I think someone will think of it as "not something" first.

src/logger/loggerApi.c
```diff
@@ -1118,11 +1123,14 @@ static void sendTimerConfig(Serial *serial, size_t startIndex, size_t endIndex)
         TimerConfig *cfg = &(getWorkingLoggerConfig()->TimerConfigs[i]);
         json_objStartInt(serial, i);
         json_channelConfig(serial, &(cfg->cfg), 1);
-        json_uint(serial, "st", cfg->slowTimerEnabled, 1);
+         /* DEPRECATED.  Only here for compatibility */
+        json_uint(serial, "st", 0, 1);
```
brentpicasso: we should check if the app fails if the value is missing... if it doesn't fail, we should remove it.

include/timer/timer_config.h
stieg: Eventually... yes.  But this is a large scale refactor that I did not want to get into now as already this got a bit out of scope (but was necessary for some of the new api methods related to the config).

src/logger/loggerSampleData.c
```diff
@@ -175,34 +175,6 @@ float get_analog_sample(int channelId)
 }
 #endif
 
-#if TIMER_CHANNELS > 0
```
brentpicasso: we removed this bock - what replaced it?

include/timer/timer_config.h
stieg: Huh?  Where?  You mean between values and define characters?  Those are simply tab characters as is the Linux standard if so.

src/logger/loggerApi.c
stieg: Ok.  Should make this a system wide macro instead of a local one.

src/logger/loggerApi.c
stieg: Was simply doing this for API compatibility.  Happy to kill it if the App doesn't mind.

src/logger/loggerSampleData.c
stieg: It was moved into timer_config.c.  In there the pre-processor directive isn't needed since the code will be discarded by the linker if not used.

src/timer/timer.c
```diff
@@ -21,34 +21,55 @@
...
 
-static uint16_t get_timer_quiet_period(const TimerConfig *tc)
+/**
+ * Calculates the highest quiet period usable based on the timer
+ * configurations.  We calculate this by figuring out the expected
+ * maximum HZ of the engine based in the info provided.  Then we simply
+ * take the inverse of that and multiply it by the number of micro seconds
+ * in a second.
+ */
+static unsigned int get_timer_auto_quiet_period(const TimerConfig *tc,
```
brentpicasso: better name would be timer_calc_quiet_filter_period

src/timer/timer.c
```diff
@@ -21,34 +21,55 @@
...
-         */
-        return QUIET_PERIOD_US / tc->pulsePerRevolution;
+        const float max_rpm = tc->cfg.max;
+        const float ppr = (float) tc->pulsePerRevolution;
+        const float max_hz = max_rpm * ppr * MAX_RPM_MULT / period;
+
+        return max_hz ? US_IN_A_SEC / max_hz : 0;
+}
+
+static unsigned int get_timer_quiet_period(const TimerConfig *tc)
```
brentpicasso: Related, exposed functions could benefit from the 'timer_' prefix so they're consistent with the module they're representing

timer_get_quiet_period

src/timer/timer.c
```diff
@@ -93,3 +114,24 @@ void timer_reset_count(size_t channel)
 {
     timer_device_reset_count(channel);
 }
+
+float get_timer_sample(const int cid)
```
brentpicasso: timer_get_sample

src/logger/loggerSampleData.c
brentpicasso: I see, it's moved to timer.h

include/timer/timer_config.h
brentpicasso: there's no space before the -1 is what I mean, sorry.

brentpicasso:
Overall looks pretty good, aside from notes.

 just homing in on the core quiet period algorithm. Can test on the bench shortly

include/timer/timer_config.h
stieg: I don't follow.  There is a tab character there.  That is just how it lines up with the tab.

brentpicasso: screenshot from 2016-03-07 10 50 26

This is what I was seeing in github. if there's a tab there, no prob.

src/logger/loggerApi.c
brentpicasso: sgtm

src/logger/loggerApi.c
stieg: Well... what version of the app.  Seems like a wide range of apps to support.  I say we just leave it in and kill it as part of 3.0.  Already have a ticket filed for it.

src/timer/timer.c
stieg: For static methods I think we don't need the `timer_` prefix.  It is unnecessary since no-one will be using the method outside the scope of this file.  Hence I think its acceptable to simply name this `get_quiet_period`.  I will name the above `calc_quiet_period`.  It simple that way and easy to follow

include/timer/timer_config.h
stieg: Ahh... the browser is fooling you.  Try adjusting your zoom.

PR #473, opened 2016-03-16
corrected units for OBD2 speed conversion

### LINKED ISSUE #472: KPH to MPH conversion is incorrect
Should be kph * 0.621371

brentpicasso:
resolves #472

### DIFF
--- a/src/OBD2/OBD2.c
+++ b/src/OBD2/OBD2.c
@@ -53,7 +53,7 @@ static float celcius_to_farenheight(float celcius)
 
 static float kph_to_mph(float kph)
 {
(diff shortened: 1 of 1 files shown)


src/OBD2/OBD2.c
```diff
@@ -53,7 +53,7 @@ static float celcius_to_farenheight(float celcius)
 
 static float kph_to_mph(float kph)
 {
-    return kph * 1.60934;
+    return kph * 0.621371;
```
stieg: This is good.  I'm going to file a ticket to centralize where we do our unit conversions so we can unit test them going forward.

PR #477, opened 2016-03-18
remap CAN peripheral to alternate port, since USB was added

brentpicasso:
We moved the hardware around on the latest board revision, to make room for the USB jack. the end.


stieg:
Ha.  I initially read this as for MK2 instead of RCT.  Funny stuff.  :shipit:

PR #481, opened 2016-03-21
Fix blocking of telemetry messages

brentpicasso:
fixed regression where a higher sample rate somewhere in the system blocks all telemetry messages


stieg:
@brentpicasso I threw my code in here since it is much inline with what you are doing, but applies it to all places instead of just the one.  Tested it on my tablet and it seems to work well, but the reading of the configs is slow.  Looking into it now.  LMK what you think.

brentpicasso:
Looks good.  

Side note, In the variable names throughout the code we need to disambiguate the coded sample rates with sample rates in Hz.  because the word 'rate' == Hz.  The number of ticks between samples isn't a rate.

brentpicasso:
I mean, rate != Hz.

PR #483, opened 2016-03-21
Put BT back in targetBaud if failed to detect baud rate

### LINKED ISSUE #482: If a BT client is already connected,  RCP leaves baud rate at 9600
if RCP starts up and BT is already connected it leaves the default baud rate at 9600 (the last baud rate it attempted), and then the tablet never can communicate, b/c wrong baud rate.

Under this condition it should reset it to the default programmed baud rate. 

BT: Trying Baudrate: 115200
BT: Trying Baudrate: 57600                                                                                           
BT: Trying Baudrate: 9600                                                                                            
BT: Could not detect device using known baud rates.                                                                  
BT: Failed to communicate with device.                                                                               
BT: Failed to provision module. A client may already be connected.

stieg:
Fixes #482

### DIFF
--- a/src/devices/bluetooth.c
+++ b/src/devices/bluetooth.c
@@ -43,6 +43,7 @@
  * your own peril.
  */
 #define BT_BACKOFF_MS	1000
+#define BT_BAUD_RATES	{115200, 57600, 9600}
 #define BT_COMMAND_WAIT	600
 #define BT_DEFAULT_NAME	"RaceCapturePro"
 #define BT_DEFAULT_PIN	"1234"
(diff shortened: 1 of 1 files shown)


src/devices/bluetooth.c
```diff
@@ -230,6 +234,10 @@ int bt_init_connection(DeviceConfig *config)
         switch (baud) {
         case 0:
                 pr_info("BT: Failed to communicate with device.\r\n");
+
+                /* Restore the targetBaud rate in case alreay in command mode */
```
brentpicasso: typo  - already

brentpicasso:
:ship:

PR #484, opened 2016-03-21
Better Serial Logging Support

stieg:
Because what we had before was not very controllable.  This patch does the following:

* Adds the ability to enable logging on a user specified serial channel
* Adds persistence for that setting so that it can persist through reboots
* Adds ability to flash NVRAM from our serial command line
* Gives cleaner names to the serial channels so its easy to see them while debugging.

This may be very helpful in diagnosing field issues with new modules as they are released.  So that is why this is going in with 2.9.0


include/serial/serial.h
```diff
@@ -67,20 +80,20 @@ Serial * get_serial(serial_id_t port);
 void configure_serial(serial_id_t port, uint8_t bits, uint8_t parity,
                       uint8_t stopBits, uint32_t baud);
 
-int serial_get_c_wait(const Serial *s, char *c, const size_t delay);
```
brentpicasso: why removing const?

src/devices/cellular.c
```diff
@@ -66,57 +66,6 @@ static const struct at_config* get_safe_at_config()
         return &cfg;
 }
 
-static void cell_serial_tx_cb(const char *data)
-{
```
brentpicasso: why is this being removed?   Is this from an unrelated change?

brentpicasso:
I think there's noise in this PR from unmerged changes, like the BT changes.  Could you merge stuff in so only the relevant changes are in place? Thanks

stieg:
@brentpicasso The Fix for the BT Baud issue is in here.  Once I merge that it it will disappear from this commit.  Let me do that now.

include/serial/serial.h
stieg: Because I had to put in some state trackers into the Serial object, thus making Serial mutable.  Hence why the removal of the const.

src/devices/cellular.c
stieg: This is being removed because it is no longer usable code now that we handle printing all serial messages from within the serial code.  This was here because it was the code that used to get invoked when I needed to print only the messages for the cellular device.  This patch is a much improved version of what this was.

brentpicasso:
This looks like a better candidate for 2.10 as we're strictly in bug fixing mode.

stieg:
Got a :shipit: per slack convo:
```
brent which will be aux
[19:56] 
It's a nit. feel free to ship it
[19:56] 
:simple_smile:
[19:56] 
Good thing to add and will be very helpful
```

PR #486, opened 2016-03-22
Improve BT scheduling

stieg:
Improves BT scheduling a bit by not unnecessarily waiting around.  It achieves this by only sleeping if both there was no message to tx or rx.  Then and only then will the code sleep.  Else it will process messages until such time that there are none to process.  Hopefully this moves things along as fast as possible.

### DIFF
--- a/src/logger/connectivityTask.c
+++ b/src/logger/connectivityTask.c
@@ -278,6 +278,7 @@ void connectivityTask(void *params)
         bool should_reconnect = false;
 
         while (1) {
+                portTickType xLastWakeTime = xTaskGetTickCount();
             if ( should_reconnect )
                 break; /*break out and trigger the re-connection if needed */
 
@@ -286,8 +287,7 @@ void connectivityTask(void *params)
                     connParams->always_streaming ||
                     logger_config->ConnectivityConfigs.telemetryConfig.backgroundStreaming;
 
-            const char res = receive_logger_message(sampleQueue, &msg,
-                                                    IDLE_TIMEOUT);
+            const char res = receive_logger_message(sampleQueue, &msg, 0);
 
             /*///////////////////////////////////////////////////////////
             // Process a pending message from logger task, if exists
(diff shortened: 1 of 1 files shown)


src/logger/connectivityTask.c
```diff
@@ -373,6 +373,15 @@ void connectivityTask(void *params)
                 pr_info_str_msg(connParams->connectionName, ": timeout");
                 should_reconnect = true;
             }
+
+            /*
+             * If we didn't rx or tx a message, sleep for a bit so as not
+             * starve the other tasks.
+             */
+            if (pdFALSE == res && !msgReceived) {
```
brentpicasso: I don't see how this benefits. when there are no messages to receive, xQueueReceive will block, freeing up the scheduler to service other tasks. That's how FreeRTOS works with queues.

What problem are we trying to solve?

stieg: The current code will always block if there is no message to send, but if there is a message received and needing a reply, it will wait for the above send to timeout, then process the received message.  This change fixes that by only sleeping when we both don't send anything and don't receive anything.  

What this is trying to solve is to not wait unnecessarily to process messages received by the BT stack.  In effect it is a minor optimization.  I tested it yesterday and BT responded a bit faster, but was still more sluggish than I would like.

brentpicasso:
Did a stopwatch test of reading config on r/2.9.0 and on this branch.  Android device was Samsung 10.5" tablet. Timing was from pressing button to where "RaceCapture/Pro MK2 v2.9.0" showed up again in status bar, indicating fully complete. 

r/2.9.0 = 7.7 seconds to read config on average; 6 tests
this branch: 7.7 seconds to read config on average; 6 tests

It's my recommendation is to reject this change as it adds complexity for no measurable benefit.

stieg:
Ok.  Nak'd.

On Tue, Mar 22, 2016 at 11:43 AM, Brent Picasso <notifications@github.com>
wrote:

> Did a stopwatch test of reading config on r/2.9.0 and on this branch.
> Android device was Samsung 10.5" tablet. Timing was from pressing button to
> where "RaceCapture/Pro MK2 v2.9.0" showed up again in status bar,
> indicating fully complete.
>
> r/2.9.0 = 7.7 seconds to read config on average; 6 tests
> this branch: 7.7 seconds to read config on average; 6 tests
>
> It's my recommendation is to reject this change as it adds complexity for
> no measurable benefit.
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/486#issuecomment-199897308>
>



--

PR #488, opened 2016-03-22
Make BT init when connected to device not slow

stieg:
Our handling of BT devices when they are not responding is unfortunately
a bit slow.  This was because I was using a very safe and high backoff
value, trying more baud rates than was probably needed, and being a bit
redundant.  This change fixes that by doing the following:

* Reducing the backoff time from 1000ms to 500ms.  This + the response
  time of 600ms should be sufficient to make the minimum 1000ms RTT time
  noted by the docs.
* Don't try the target baud if its one of the pre-fabricated baud rates.
* Don't try the 57600 baud rate since that is not really ever used.


stieg:
Addresses #488

Jeff-Ciesielski:
:+1:  Looks fine to me.  Seems like it should cut the retry time pretty significantly.

brentpicasso:
Measured 10 seconds restore

BT: Trying Baudrate: 9600                                                       
BT: Could not detect device using known baud rates.                             
BT: Failed to communicate with device.                                          
[BT] Restoring to target baud: 115200
<long delay here>                                           
BT: Failed to provision module. A client may already be connected.

brentpicasso:
Long delay after restoring default target baud

Also, each iteration is still taking about 1 second, not 500ms

brentpicasso:
I see > 1 second after the "trying 9600" pops up in log.

stieg:
I added one minor change that will help speed up the init when its connected.  But at this point this is as fast as it will get with the more resilient features added.

stieg:
> Long delay after restoring default target baud

Fixed that.

> Also, each iteration is still taking about 1 second, not 500ms

Yes that is expected.  It tries to ping the device 3 times before failing.

brentpicasso:
:shipit:

PR #490, opened 2016-03-22
P/stieg/better bt telemetry

stieg:
Ignore the BT init improvement change here as that is already part of a separate change.  The main change here is the placement of the max telemetry limits into the ConnParameters struct.  This allows us to more granularity control whether or not we should send a telemetry message when compared against the old method where we just naively set the lower of the two rates.


stieg:
After convo with @brentpicasso we will hold off with this until the next major release.  There are multiple ways to approach this and while this can work, probably better do re-evaluate the approach.

PR #492, opened 2016-03-23
Minor fixup to serial buffer printing.

stieg:
It wasn't printing nicely before.  This fixes it so it does print nicely.  Pretty trivial really.


brentpicasso:
:shipit:

Jeff-Ciesielski:
:+1:

PR #493, opened 2016-03-23
Consolidate array_utils into macros.h

stieg:
Simple fix that removes an almost unnecessary header file `array_utils.h"`and puts its contents into the more versatile header file `macros.h`.


brentpicasso:
:shipit:

Jeff-Ciesielski:
:+1:

Issue #489, opened 2016-03-22
[gsm] Failed to read phone number


stieg:
Bug with the ublox module.  Failing to read phone number.  Need to fix before release.

PR #494, opened 2016-03-23
Add retry for CNUM command on Ublox driver

stieg:
AT&T cards seem to have some strange behavior that needs additional
setup that isn't entirely clear.  Running CNUM the first time seems
to trigger that setup, but at the cost of the command failing.  This
little hack causes us to try getting the number 3 times before failing.

It also adjusts the code so that the failure to grab any of the SIM
info does not cause other bits of SIM info to fail to be grabbed as
well.


stieg:
Fixes #489

brentpicasso:
:shipit:

PR #499, opened 2016-04-04
P/stieg/ring buffer fixes

stieg:
Re-implements ring buffer so its more sane with more features.  Also adds a new thread safe version that wraps the newly re-written version and implements its use in our logging code to prevent the annoying partial writes we see.

### DIFF
--- a/include/util/ring_buffer.h
+++ b/include/util/ring_buffer.h
@@ -42,7 +42,8 @@ size_t ring_buffer_peek(struct ring_buff *rb, void *data,
                         size_t size);
 size_t ring_buffer_put(struct ring_buff *rb, const void *data,
                        size_t size);
-
+size_t ring_buffer_write(struct ring_buff *rb, const void *data,
+                         size_t size);
 CPP_GUARD_END
 
 #endif /* __RING_BUFFER_H__ */
--- a/include/util/ring_buffer.h
+++ b/include/util/ring_buffer.h
@@ -29,68 +29,20 @@
 
 CPP_GUARD_BEGIN
 
-struct ring_buff {
-    char *buf;
-    size_t size;
-    char *head;
-    char *tail;
-};
+struct ring_buff;
 
-/**
- * Copies size bytes from the ring buffer into data.
- * @return The number of bytes copied.
- */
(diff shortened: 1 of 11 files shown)


Jeff-Ciesielski:
Haven't had a chance to do a super in-depth dive, but given a few once overs, it looks pretty sane.

I'm not terribly comfortable with the ringbuf_put behavior though. IMO, this is a landmine just waiting to happen.  I can see a few situations where this behavior could potentially be desirable, but in a SPSC scenario, it imposes an undue prerequisite that the user absolutely understands the consequences of performing a put without ensuring the buffer has capacity and could lead to subtle bugs since you can't guarantee that this behavior is safe by default.

I think a simple way around this would be to provide an 'overwrite' flag (or wrapper function) that allowed this behavior, and rejected it by default.  That way, you still get the FIFO overwrite behavior if you really want/need it, but the average case wouldn't do anything dangerous.  It would still be up to the user to decide what to do with the leftover data, but it wouldn't make any assumptions that could lead to data corruption on the consumer side.

stieg:
> I'm not terribly comfortable with the ringbuf_put behavior though. IMO, this is a landmine just waiting to happen.

I hear you and would love suggestions.  I am using both behaviors as is and its quite nice.  Yes its possible for someone to screw this up... but I'm not sure where to balance flexibility and guarantees of sanity.  I could create a `write` method that does put without the clobbering if that would help?

Jeff-Ciesielski:
>Yes its possible for someone to screw this up... but I'm not sure where to balance flexibility and >guarantees of sanity. 
I think at the API level, it just needs to be clear what's going to happen.  We're programming firmware in C, unsafe code is just part of the gig :).   I don't think we need to prevent the ability to do this, just make it clear via the name of the function that it will have consequences (and provide the ability to work consequence free if desired without needing to think about it)

>I could create a write method that does put without the clobbering if that would help?

I think so.  If it did the put, and only inserted as much data as there was room for (returning that amount so the user can handle on their end), I think that would be a reasonable compromise.

stieg:
@Jeff-Ciesielski Done.  Have a look.

Jeff-Ciesielski:
:+1: LGTM

PR #502, opened 2016-04-07
Add CDC-ACM support

Jeff-Ciesielski:
All necessary functionality appears good, you can interact with a shell!

Note: This is completely different USB hardware than what is present on the F4 / RCP and the drivers share nothing in common (aside from an interface).  

Also: We'll need to do some CAN testing.  The two pieces of hardware share an interrupt vector.


Jeff-Ciesielski:
Almost forgot: The majority of non ST supplied code is in USB_CDC_device_stm32.c, the rest is from their vendor code.

stieg:
Ok to test

platform/rct/Makefile
```diff
@@ -214,14 +228,15 @@ inc-y += $(RC_PLATFORM_DIR)/mk2/hal/fat_sd_stm32/fatfs
...
+CFLAGS := -DUSE_STDPERIPH_DRIVER -DSTM32F30X -MD -std=gnu99
 CFLAGS += -mcpu=cortex-m3 -mthumb
 CFLAGS += $(VERSION_CFLAGS)
 CFLAGS += $(BOARD_DEFINES)
 CFLAGS += -fno-builtin-sprintf
 CFLAGS += $(addprefix -I,$(inc-y))
 CFLAGS += -O0 -g
-LDFLAGS := -mcpu=cortex-m3 -mthumb -TProject/STM32_FLASH.ld -Wl,-cref,-u,Reset_Handler,-Map=$(OUTDIR)/main.map,--gc-sections
+LDFLAGS := -mcpu=cortex-m3 -mthumb -TProject/STM32_FLASH.ld		\
+-Wl,-cref,-u,Reset_Handler,-Map=$(OUTDIR)/main.map,--gc-sections
```
stieg: Instead of wrapping lets just do a `LDFLAGS += ...` line.

platform/rct/Project/FreeRTOSConfig.h
```diff
@@ -159,5 +159,5 @@ standard names. */
 #define vPortSVCHandler SVC_Handler
 #define xPortPendSVHandler PendSV_Handler
 #define xPortSysTickHandler SysTick_Handler
-
+void __attribute__((used)) vTaskSwitchContext(void);
```
stieg: What is happening here.  Comments would be useful.

Jeff-Ciesielski: Oops, snuck through.  I'm going to delete that line for now. It'll be required once we enable link time optimisation.

platform/rct/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -66,27 +68,27 @@ static int initQueues()
     int success = 1;
 
     /* Create the queues used to hold Rx and Tx characters. */
-    xUsart0Rx = xQueueCreate(UART_QUEUE_LENGTH,
+    xUsart0Rx = xQueueCreate(UART_RX_QUEUE_LENGTH,
                              (unsigned portBASE_TYPE)sizeof(signed portCHAR));
-    xUsart0Tx = xQueueCreate(UART_QUEUE_LENGTH + 1,
+    xUsart0Tx = xQueueCreate(UART_TX_QUEUE_LENGTH + 1,
```
stieg: Lets get rid of this annyoing `+ 1` crap here since it invalidates the point of using the `#define` above.

platform/rct/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -66,27 +68,27 @@ static int initQueues()
...
-    xUsart2Tx = xQueueCreate(UART_QUEUE_LENGTH + 1,
+    xUsart2Tx = xQueueCreate(UART_TX_QUEUE_LENGTH + 1,
                              (unsigned portBASE_TYPE)sizeof(signed portCHAR));
     if (xUsart2Rx == NULL || xUsart2Tx == NULL) {
         success = 0;
         goto cleanup_and_return;
     }
 
-    xUsart3Rx = xQueueCreate(UART_QUEUE_LENGTH,
+    xUsart3Rx = xQueueCreate(UART_RX_QUEUE_LENGTH,
```
stieg: Also... this is going to break quite hard with the new virtual serial code.  Oh well...

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
-#include "USB-CDC_device.h"
```
stieg: Please include our file header in all files you create going forward.  It is in our CODING_STANDARD.md doc.

stieg: Also please sort your includes.

stieg: If you are using emacs (which I think you are) highlight the includes and issue `sort-lines`

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
+	volatile uint32_t USB_Tx_ptr_in;
+	volatile uint32_t USB_Tx_ptr_out;
+	volatile uint32_t USB_Rx_ptr_in;
+	volatile uint32_t USB_Rx_ptr_out;
+	volatile int _init_flag;
+	xSemaphoreHandle _rlock;
+	xSemaphoreHandle _wlock;
+} usb_state;
+
+/* ST's USB hardware doesn't include an internal pullup resistor on
```
stieg: Nice comment here.  Please use our multi-line comment formatting of 

```C
/*
 * Hi I am a
 * multi-line comment.
 * Bye!
 */
```

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
+static int USB_CDC_force_reenumeration(void)
+{
+	GPIO_InitTypeDef GPIO_InitStructure;
+
+	RCC_AHBPeriphClockCmd(RCC_AHBPeriph_GPIOA, ENABLE);
+	GPIO_InitStructure.GPIO_Pin = GPIO_Pin_12;
+	GPIO_InitStructure.GPIO_Speed = GPIO_Speed_50MHz;
+	GPIO_InitStructure.GPIO_Mode = GPIO_Mode_OUT;
+	GPIO_InitStructure.GPIO_OType = GPIO_OType_PP;
+	GPIO_InitStructure.GPIO_PuPd = GPIO_PuPd_NOPULL;
```
stieg: So I take that it would be unwise to use the GPIO pull up?

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
+int USB_CDC_device_init(const int priority)
+{
+	/* Perform a full USB reset */
+	USB_CDC_force_reenumeration();
+
+	/* Create read/write mutexes */
+	usb_state._rlock = xSemaphoreCreateMutex();
+	usb_state._wlock = xSemaphoreCreateMutex();
+
+	if (usb_state._rlock == NULL || usb_state._wlock == NULL) {
```
stieg: NIT: Use Yoda checks.  So `NULL == usb_state._rlock`.  This way if there were an accidental single equals... compiler catches it.

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
Jeff-Ciesielski: Correct.  Once the USB MAC takes over, this is an unpulled open drain output. (which happens to be pulled up externally)

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
-int USB_CDC_device_init()
+	return 0;
+}
+
+void USB_CDC_Write(portCHAR *cByte, int len)
 {
-    vcp_setup();
+	xSemaphoreTake(usb_state._wlock, portMAX_DELAY);
+	while (len--) {
+		/* If the transmit buffer is full, wait until the USB
```
stieg: Multi-line comments again.  Fix applies throughout.

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
-              USB_OTG_FS_CORE_ID,
-              &USR_desc,
-              &USBD_CDC_cb,
-              &USR_cb);
-    _init_flag = 1;
+		usb_state.USB_Tx_Buffer[usb_state.USB_Tx_ptr_in] = *cByte++;
 
-    return 0;
+		/* Handle wrapping */
+		if(usb_state.USB_Tx_ptr_in == VIRTUAL_COM_PORT_DATA_SIZE - 1) {
```
stieg: Hrmm... why not use a ring_buff here instead of rolling your own?

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
+			usb_state.USB_Tx_ptr_in++;
+		}
+
+	}
+	xSemaphoreGive(usb_state._wlock);
 }
 
+/* TODO: Deprecate this.  It's unused, and having to use strnlen this
+ * deep in the driver stack gives me a wiggins */
 void USB_CDC_send_debug(portCHAR *string)
```
stieg: I eliminated it in MK2 so sounds good to be consistent here :)

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
```diff
@@ -1,52 +1,250 @@
...
+				/* Enable the receive of data on EP3 */
+				SetEPRxValid(ENDP3);
+			}
+
+			/* Note: We don't need to handle wrapping in
+			 * the receive calls because we always fill
+			 * the rx buffer from the start and drain to completion */
+			byte_received = true;
+		} else {
+			if (delay == portMAX_DELAY) {
```
stieg: Should this really be treated differently even if it is portMAX_DELAY.  I wouldn't.

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
Jeff-Ciesielski: Because I need to poke at the guts in the interrupt handlers and manually control the underlying storage due to the way the USB MAC handles packet storage in a separate area of memory.  I'd have needed an intermediary buffer (which seemed inefficient).  I can look into updating it in the future, but for now, I think this was the better solution.

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
Jeff-Ciesielski: I need to double check, but I think portMAX_DELAY means 'wait forever' (although, 4 billion ticks is a long time)

platform/rct/hal/usart_stm32/usart_device_stm32.c
Jeff-Ciesielski: Fair enough, Wasn't sure if there was a reason for them, didn't want to pee in the pool. (Though in the future, hopefully we can keep cross-driver contamination out of the mix entirely, I only touched this because I was out of heap space)

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
stieg: I think getting something working is the highest priority.  But #2 priority will be trimming away all unnecessary memory usage just because of the limited space we are working with.  I would suggest filing some issues against this so its not forgotten as we move forward.  Unfortunately in software nothing is more permanent that temporary.

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
stieg: Its just always the largest value you can pass.  In the queues it is designed to mean that, but in practice you should always program as if the lock time expired and be able to handle that corner case.  Altering the behavior here is probably a bad idea so I would rather you be consistent with the rest of the system and handle that super rare case where the timeout expires.  Should be trivial to do...

stieg:
My only concern at a high level is the number of buffers used to make all this work.  Seems like way more than should be needed, but if its just to stand up an initial version then so be it.  In the latter case, please file tickets to optimize it so its not forgotten.  Looks great otherwise.

platform/rct/hal/usb_stm32/USB-CDC_device_stm32.c
Jeff-Ciesielski: Sorting includes is preferential and can, in fact, cause some (admittedly not well designed) code to work.  That needs striken out of the standard if it currently exists.  It doesn't serve a purpose.

stieg: I respectfully disagree.  You sort includes so that people can easily find if includes are there or not and to be able to easily see if someone has duplicated the includes.  If code is effected by order of includes, then exposing this flaw during development is very important to do so that it can be fixed.  Order of includes should never matter in C so long as they are all at the top of the file.  Ensuring they are all there and not duplicated is why we do it.

stieg:
Code looks good.  The sorting of includes is important, but is not a blocker.  :shipit:

PR #504, opened 2016-04-09
Improve lua code

stieg:
Moves the Lua code to the Lua directory (making it consistent with the headers) and removes all methods that don't need to be public.  This is simply cleaner.  Since there is no functional change in the code here, this is safe to put into 2.9.1.


src/lua/luaLoggerBinding.c
```diff
@@ -740,71 +666,244 @@ int Lua_ReadOBD2(lua_State *L)
...
+        lua_pushinteger(L, chan_id);
+        return 1;
+
+bug:
+        return luaL_error(L, "BUG. Should never get here.  "
+                             "Please inform AutosportLabs");
 }
 
-int Lua_SetVirtualChannelValue(lua_State *L)
+static int lua_get_virtual_channel(lua_State *ls)
```
brentpicasso: inconsistent naming change here?

src/lua/luaLoggerBinding.c
```diff
@@ -740,71 +666,244 @@ int Lua_ReadOBD2(lua_State *L)
...
 
-int Lua_IsLogging(lua_State *L)
+static int Lua_IsLogging(lua_State *L)
 {
     lua_pushinteger(L, logging_is_active());
     return 1;
 }
 
-int Lua_SetLED(lua_State *L)
+static int lua_set_led(lua_State *ls)
```
brentpicasso: ditto on naming change. don't see a pattern of renaming some and not others?

brentpicasso:
the weird changes rustles my jimmies but slack convo assuaged my concerns.

:shipit:

src/lua/luaLoggerBinding.c
stieg: So would you prefer I change it to the no longer acceptable naming standard (Camel Case) or should I alter all the method names?  I didn't do the latter because I felt it would cause unnecessary churn/noise.  But I would be happy to do it and get it out of the way.

brentpicasso: Ah hell, just get it out of the way. Happy friday!

stieg: Ok.  Will do.

stieg:
Closing this because I decided to overhaul Lua bindings while I was here.  PR to follow very soon.

Issue #403, opened 2016-01-25
If Lua script is too busy, cannot update script


brentpicasso:
If Lua script is too busy (e.g. never returning from the onTick() function) the script cannot be updated due to not being able to get a lock on the Lua runtime. 

Work-arounds: always provide a way for the onTick() to exit after a reasonable amount of time.

stieg:
We can create a simple Lua_Watchdog task for this.  Using a semaphore if the LUA task takes too long then we can kill the execution of the task and re-start it (or something along those lines).

kosenko-max:
Is it possible to keep ability for script to never go out of onTick()? It's the only way how I can consistently pull CAN data without gaps. All other approaches put huge timing gaps in the log.

Maybe kill Lua once script update is required? Instead of killing it by watchdog?

stieg:
> Is it possible to keep ability for script to never go out of onTick()? It's the only way how I can consistently pull CAN data without gaps. All other approaches put huge timing gaps in the log.

Well starting with 2.9.0 you will have support for setting the onTick invocation rate to be as high as 1000Hz.  So really you will never need to worry about leaving `onTick` starting with 2.9.0.

> Maybe kill Lua once script update is required? Instead of killing it by watchdog?

Yeah that is the plan of attack in general.  Right now the Lua task isn't well suited for destruction and re-init so some work would need to happen to make it better suited for this use case.

stieg:
In 2.9.0.

Issue #471, opened 2016-03-12
Clear virtual channels upon Lua Script Reload


kosenko-max:
From time to time change of script would not reset existing virtual channels list, so it hits channel count limit and requires power restart of hardware.

stieg:
Yep.  Virtual Channels are only designed for additions.  I'm assuming you are hitting this because you are testing Lua Scripts?

kosenko-max:
I assume that firmware and app should clean-up table of virtual channels
after script update. Am I right?

stieg:
Not quite.  We expect you not to change the channel name.  hence the name check.  Now if that is broken... Then there is a problem

kosenko-max:
Well, since I don't fit in 30 channels I have to jump between scripts and
power restart hardware every time. I just don't really see a point in
storing old channels if script has updated. I can understand a reason to
store app settings of channels even if they don't exist anymore, but not
channels themselves in a firmware or app channel list.

brentpicasso:
Good engineering practices encourage separation of concerns, and that would suggest that updating the script would not reach into a different part of the system to do something - touching virtual channels, for example. That may end up having an unintended side effect someone else with a different, equally justifiable use case.

Perhaps a lua function to clear existing channels might help, then it can be at the top of every script, if necessary?

kosenko-max:
That would work also, I just can't imagine any scenario where channels must
be persistent after script updates. And since power cycling would clean up
this table anyway, whatever this scenario is - it would not work
consistently anyway.

stieg:
in 2.9.0

PR #508, opened 2016-04-11
P/stieg/lua improvements

stieg:
Implements the ability for our system to stop an unresponsive Lua task, reprogram it, and restart it, allowing users to get around bad scripting errors that would never return control to the underlying task.  Also implements clearing of virtual_channels upon script load.  And as a cherry on top, expands the size of the stack to 8K from 6K.  Hopefully that will be enough.

This fixes issues #403 and #471

### DIFF
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -409,6 +409,8 @@ bool lua_task_stop()
         lua_close(state.lua_runtime);
         state.lua_runtime = NULL;
 
+        led_disable(LED_ERROR);
+
         /*
          * If we didn't get the semaphore, then the task we just
          * killed was the likely owner of it (better be).  Since
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -391,12 +391,16 @@ bool lua_task_stop()
 
         const size_t ticks = msToTicks(LUA_LOCK_WAIT_MS);
         const bool got_lock = get_lock_wait(ticks);
-        if(!got_lock)
-                pr_warning(_LOG_PFX "Killing Lua task b/c unresponsive\r\n");
 
         /* Its possible to have a runtime but no task */
         if (state.task_handle) {
-                pr_info(_LOG_PFX "Stopping Lua Task\r\n");
+                if (!got_lock) {
+                        pr_warning(_LOG_PFX "Killing Lua task because "
+                                   "it is unresponsive\r\n");
+                } else {
+                        pr_info(_LOG_PFX "Gracefully stopping Lua Task\r\n");
+                }
+
                 vTaskDelete(state.task_handle);
                 state.task_handle = NULL;
         }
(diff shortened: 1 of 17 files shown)


src/lua/luaTask.c
```diff
@@ -38,258 +38,297 @@
...
+};
+
+static struct _state {
+        int priority;
+        xSemaphoreHandle semaphore;
+        xTaskHandle task_handle;
+        lua_State *lua_runtime;
+        size_t callback_interval;
+        size_t lua_mem_size;
+} state;
```
Jeff-Ciesielski: 👍  me gusta.

src/lua/luaTask.c
```diff
@@ -38,258 +38,297 @@
...
-        pr_info("lua: Loading lua script (len = ");
-        pr_info_int(len);
-        pr_info("): ");
+static void lua_failure_state(const int cause)
+{
+        const char *msg = get_failure_msg(cause);
+        pr_warning_str_msg(_LOG_PFX "Failure: ", msg);
 
-        lua_gc(g_lua, LUA_GCCOLLECT,0);
+        while(true) {
```
brentpicasso: does this run forever inside the lua task?

brentpicasso: IOW, we can kill the task which then exits this loop?

brentpicasso:
Brief look, but looks good. Would love to test this directly to see it in action. will try to do so.

src/lua/luaTask.c
```diff
@@ -300,64 +339,176 @@ void run_lua_interactive_cmd(Serial *serial, const char* cmd)
...
+                pr_error(_LOG_PFX "Already initialized\r\n");
+                return false;
+        }
+
+        /*
+         * Always set initial state here.  Needed because we
+         * may kill this task later.  Setting the lock is what we use
+         * to determine that we are initialized.
+         */
+        vSemaphoreCreateBinary(state.semaphore);
```
Jeff-Ciesielski: Your first commit was to replace this with a mutex to allow for priority inheritance.  Was it intentional that you changed it back?

src/lua/luaTask.c
stieg: Correct.

src/lua/luaTask.c
stieg: Yes.  I forgot to nuke that commit.  Reason was that Mutex includes the priority mechanism and expects all callers to both take and give the mutex back.  We can't do that if we kill the process, so had to move back to a binary semaphore.

stieg:
Killed that bad commit message, toyed with the code a tiny bit.

stieg:
Per Slack Convo in `#firmware`:
```
Look at the RAM usage between when we load Max's script the second time and when its restored back to the default Lua script.  ​_Exact_​ same free RAM value.

[17:19] 
stieg @brent:
[17:20] 
ryan :thumbsup:
new messages
[17:52] 
brent Shipit
```

PR #518, opened 2016-04-14
[RFC] Feedback on Wifi Code

stieg:
So since this WiFi code has gotten quite large, a simple code review process will not suffice here.  So to try and make this a bit easier I am opening this RFC PR to get review feedback on the code.  Take a look at it and give me your feedback so I may incorporate it.  Thanks.


include/serial/ram_serial.h
```diff
@@ -0,0 +1,48 @@
+/*
```
stieg: Please ignore the ram_serial code.  It will be deleted soon.

include/gps/dateTime.h
```diff
@@ -107,6 +107,10 @@ int getUptimeAsInt();
  */
 void getDateTimeFromEpochMillis(DateTime *dateTime, millis_t millis);
 
+tiny_millis_t getUptimeFromNow(const tiny_millis_t to_add);
```
brentpicasso: naming convention

include/gps/dateTime.h
```diff
@@ -107,6 +107,10 @@ int getUptimeAsInt();
...
+
+bool isTimePassed(const tiny_millis_t time);
```
brentpicasso: ditto

include/modem/at.h
```diff
@@ -0,0 +1,164 @@
...
+
+#define AT_CMD_MAX_CMDS	3
+#define AT_CMD_MAX_LEN	64
+#define AT_DEV_CVG_DELIM_MAX_LEN	3
+#define AT_RSP_MAX_MSGS	8
+#define AT_URC_MAX_LEN	16
+#define AT_URC_MAX_URCS	16
+#define AT_URC_TIMEOUT_MS	5
+
+enum at_rx_state {
```
brentpicasso: some comments describing these enums would be helpful

src/devices/esp8266.c
```diff
@@ -0,0 +1,848 @@
...
+        }
+}
+
+/**
+ * Sets up the internal state of the driver.  Must be called before init.
+ */
+static bool _setup(struct Serial *s, const size_t max_cmd_len)
+{
+        /* Check if already initialized.  If so, do nothing */
+        if (state.ati)
```
brentpicasso: Do we have anything firm about naming of module level static variables? When I saw this, I immediately searched 
for the declaration of a local variable.

src/devices/esp8266.c
```diff
@@ -0,0 +1,848 @@
...
+        size_t len;
+        void (*cb)(int);
+};
+
+/**
+ * This call back is designed to handle two types of response (since that is the
+ * way this wonkey device works).  So based on the state and reply we figure out
+ * how to handle things.
+ */
+static bool send_data_cb(struct at_rsp *rsp, void *up)
```
brentpicasso: Despite the comments, not sure what this function is supposed to do exactly. maybe clarify documentation?

src/devices/esp8266.c
```diff
@@ -0,0 +1,848 @@
...
+                cmd_failure(cmd_name, "Can't send empty messages");
+                return false;
+        }
+
+        /*
+         * Set the state before beginning the command.  This is a 2 part
+         * command with a fair bit of data, thus we need to use some
+         * internal state to handle it all.
+         */
+        struct tx_info *ti = portMalloc(sizeof(struct tx_info));
```
brentpicasso: when does this get free'd?

src/drivers/esp8266_drv.c
```diff
@@ -0,0 +1,455 @@
...
+#include "queue.h"
+#include "task.h"
+#include "wifi.h"
+#include <stdbool.h>
+#include <string.h>
+
+#define _AT_TASK_TIMEOUT_MS	3
+#define _BAD_STATE_BACKOFF_MS	3000
+#define _CLIENT_BACKOFF_MS	30000
+#define _DAEMON_SERVER_PORT	7223
```
brentpicasso: maybe put this in a higher level config file

src/logger/loggerApi.c
```diff
@@ -1659,10 +1670,45 @@ int api_setScript(Serial *serial, const jsmntok_t *json)
...
-int api_runScript(Serial *serial, const jsmntok_t *json)
+int api_runScript(struct Serial *serial, const jsmntok_t *json)
 {
         lua_task_stop();
         lua_task_start();
         return API_SUCCESS;
 }
 #endif /* LUA_SUPPORT */
+
+int api_set_wifi_client_cfg(struct Serial *s, const jsmntok_t *json)
```
brentpicasso: do we enable server config somewhere yet?

include/gps/dateTime.h
stieg: Ok.  Will fix.

include/modem/at.h
stieg: Sure.

src/devices/esp8266.c
stieg: See Chapter 4 in https://www.kernel.org/doc/Documentation/CodingStyle

src/devices/esp8266.c
stieg: Its a callback method that gets called twice.  If you look at the switch you will see one fork does a `return true` at the end.  If during a callback you return `true`, then it implies there are more commands to be handled.  So it will wait for more replies to come from the modem and when it gets them and goes down one of the other paths, it will `return false`.  I'll clarify the docs.

src/devices/esp8266.c
stieg: In the callback.

src/drivers/esp8266_drv.c
stieg: Capabilities.h?  Could do.

src/logger/loggerApi.c
stieg: Right now its hacked to always be enabled.

stieg: Not yet.  I have yet to write the code for the server.  Need to implement the device commands, then tie them into the driver.

Issue #214, opened 2015-10-26
Build in support for LUA unit tests.


ryandoherty:
Need to find a way to do unit tests for our LUA scripting.  This will ensure a stronger API that we don't need to test manually.

PR #519, opened 2016-04-15
Lua Bindings Overhaul

stieg:
This change effectively overhauls our Lua bindings by doing the following:

* Ensuring all files are in the /src/lua/ directory where they should be
* Removes unused methods
* Adds strict and verbose parameter count checking
* Adds strict and verbose parameter type checking
* Cleans up method names to our modern spec (mostly)
* Re-tools a bit of the flow to make binding calls consistent throughout our Lua subsystem.

The result is a Lua environment where, if a user makes a bad call, they now get a reason and line number in our log file instead of failing silently (like many methods used to do).  This will drastically
help improve user confidence in Lua since if their script is able to run initially without generating an error, chances are that it will continue to run smoothly while on the track.


Jeff-Ciesielski:
I like it.  Makes the lua subsystem much more robust and as you put it, will help to inspire confidence when people deploy this on the track.  Looks like a nice code size reduction as well.  

One thing this does highlight for me though is the need to deploy a rack of test hardware and test harnesses  that can do things like run regression tests for stuff like this.

stieg:
For sure.  Though we can do unit testing with lua locally on the code
base.  Just need to build it.

stieg:
FYI the Lua unit tests are tracked by issue #214 @Jeff-Ciesielski / @brentpicasso is this good to ship?

Jeff-Ciesielski:
If it doesn't break any known scripts, I vote :shipit:.

stieg:
I'm sure it will break people's scripts.  But that is just because they were using the method incorrectly.  This will no longer fail silently.

PR #521, opened 2016-04-19
Remove dependencies on non-existant _sbrk

stieg:
newlib was looking for this symbol because of the use of malloc,
which was a requirement of vsnprintf and friends.  This is not correct
for RCP family since our malloc methods are handled by FreeRTOS.  This
patch addresses this issue by createing memory.c.  These symbols override
the malloc, free, realloc and calloc methods defined in libc, allowing us
to use our own malloc methods instead of those provided by newlib.  This
in turn allows us to use a single malloc system as well as allows us to
start using stdlib malloc, free, realloc and calloc like a proper program
should instead of having to use portMalloc and portFree etc.


stieg:
Of note: I removed the silly line terminators on heap_4.c, so that is why the diff is so dang big.  In reality the only thing that changed there was that I copied in the realloc methodsfrom mk2's heap allocator into this one.

PR #537, opened 2016-04-26
Add ccache support in our firmware builds

stieg:
Because I like my builds to be quicker than Flash himself, I have
added support for ccache.  This drastically speeds up our build
process, turning builds that can take 30 - 40 seconds to builds that
will take 5 - 7 seconds depending on how much has changed.

By default it is disabled.  To use it simply define the following in
your .bashrc file after installing ccache:

export CCACHE=ccache

Once the CCACHE environment variable is defined, it will be used
during the build process.  That's it.

### DIFF
--- a/lib/lua/src/Makefile
+++ b/lib/lua/src/Makefile
@@ -16,7 +16,7 @@ PLAT= none
 ifeq ($(PLAT), stm32)
 	MCU   = cortex-m4
 	THUMB = -mthumb
-	CC = arm-none-eabi-gcc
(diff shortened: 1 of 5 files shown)


stieg:
@Jeff-Ciesielski or @brentpicasso I'm sure you will both appreciate this.  Builds are amazingly fast now.

Jeff-Ciesielski:
Sweet.  @brentpicasso seems particularly sensitive to build speed, so I'm sure he'll appreciate it ;) :shipit:

PR #539, opened 2016-04-27
RCT Stabilization Fixes

stieg:
This series of patches are a group of changes that have made RCT appear to be stable and fully functional with the BT chip.  The unit can work with the WiFi chip as well, but it will appear to watchdog.  This is normal as its only the connectivityTask re-initializing the device because its seeing AT commands from the WiFi unit and thinks these commands are invalid JSON.

**TODO:***
* Let this thing run all night and ensure it does not get into the Default Handler
* Have someone with a BT enabled RCT unit test that the BT stack works
* Have someone with the hardware USB patch test that the USB stack works.

Once all those ToDo's are complete, then we will be able to merge this into r/2.10.0  I'm assigning @Jeff-Ciesielski to review because @brentpicasso is busy with family bits but if you have time to look at it @brentpicasso then more power to ya.  @Jeff-Ciesielski Please test what you can in the ToDo when you have the time.  Thanks.

### DIFF
--- a/platform/rct/hal/usart_stm32/usart_device_stm32.c
+++ b/platform/rct/hal/usart_stm32/usart_device_stm32.c
@@ -169,7 +169,7 @@ static void enableRxDMA(uint32_t RCC_AHBPeriph,
      NVIC_InitTypeDef NVIC_InitStruct;
      NVIC_InitStruct.NVIC_IRQChannel = NVIC_IRQ_channel;
      NVIC_InitStruct.NVIC_IRQChannelCmd = ENABLE;
-     NVIC_InitStruct.NVIC_IRQChannelPreemptionPriority = 1; //IRQ_priority;
+     NVIC_InitStruct.NVIC_IRQChannelPreemptionPriority = UART_GPS_IRQ_PRIORITY; //1; //IRQ_priority;
      NVIC_InitStruct.NVIC_IRQChannelSubPriority = 0;
 
      NVIC_Init(&NVIC_InitStruct);
@@ -180,6 +180,14 @@ static void enableRxTxIrq(USART_TypeDef * USARTx, uint8_t usartIrq,
                           uint8_t IRQ_priority, uart_irq_type_t irqType)
 {
     NVIC_InitTypeDef NVIC_InitStructure;
+
+    /* Test to clear the Error Flags */
+    USART_ITConfig(USARTx, USART_IT_CTS, DISABLE);
+    USART_ITConfig(USARTx, USART_IT_LBD, DISABLE);
+    USART_ITConfig(USARTx, USART_IT_TC, DISABLE);
+    USART_ITConfig(USARTx, USART_IT_IDLE, DISABLE);
+    USART_ITConfig(USARTx, USART_IT_PE, DISABLE);
+    USART_ITConfig(USARTx, USART_IT_ERR, DISABLE);
 
     /* Configure the NVIC Preemption Priority Bits */
     NVIC_PriorityGroupConfig(NVIC_PriorityGroup_4);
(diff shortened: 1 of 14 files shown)


stieg:
Well unit has been running for 14+ hours at this point and seems to be stable enough.  That is good.  @Jeff-Ciesielski / @brentpicasso  LMK when you have a chance to check out the other two remaining tasks and then I'll merge this code

Jeff-Ciesielski:
USB appears to work just fine, ran through several commands and just sat there with my finger on the return key to stress the driver and it took it just fine.  Blue LED behavior is a little nutty, but it'll have a diffuser :)

platform/rct/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -181,22 +179,34 @@ static void enableRxDMA(uint32_t RCC_AHBPeriph,
...
+        USART_ITConfig(USARTx, USART_IT_WU, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_CM, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_EOB, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_RTO, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_CTS, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_LBD, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_TC, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_IDLE, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_PE, DISABLE);
+        USART_ITConfig(USARTx, USART_IT_ERR, DISABLE);
```
Jeff-Ciesielski: IIRC, these interrupt enablers are set/cleared as a mask.  Might not hurt to define something like UART_IT_ALL with these guys bitwise or'd together and have a single call.

platform/rct/hal/usart_stm32/usart_device_stm32.c
```diff
@@ -415,12 +424,23 @@ static void usart_generic_irq_handler(volatile struct usart_info *ui)
...
+         * STm32F3XX manual for a logic diagram.  If ORE triggers it won't
+         * necessarily set the USART_IT_ORE interrupt flag because we
+         * disable the USART_IT_ERR interrupt.  This breaks the check
+         * provided by USART_GetITStatus(usart, USART_IT_ORE).  This bug has
+         * been fixed in the STM32F4XX libs, but not this series :(.  Thus
+         * we must check the ISR directly to see if this is what caused the
+         * interrupt and clear it appropriately.
+         */
+        if (SET == USART_GetFlagStatus(usart, USART_FLAG_ORE))
+                USART_ClearFlag(usart, USART_FLAG_ORE);
```
Jeff-Ciesielski: :+1:  Good catch

Jeff-Ciesielski:
Minor nit, otherwise :shipit:

platform/rct/hal/usart_stm32/usart_device_stm32.c
stieg: Can't do a `USART_IT_ALL` because that would cause the `USART_ITConfig` function to pee itself.  Put differently, because there are 3 different registers internally, the call would fail with these logically anded together.  But they can probably be eliminated since you are correct about the interrupts being disabled from the get go.  This was a change made on a tired brain trying to throw the kitchen sink at an issue until it worked.  I'll remove them and test quick to ensure things work as expected.

Jeff-Ciesielski: Right on, didn't realize there were 3 interrupt mask registers for this block, ignore.

PR #540, opened 2016-04-29
Adjust USB and I2C Priorities

Jeff-Ciesielski:
Adjust I2C/USB interrupt priorities so they don't clash with the FreeRTOS minimum


stieg:
Tested this locally... solves world hunger and all disease.  :shipit:

PR #543, opened 2016-04-30
Add in basic WiFi Beacon Support

stieg:
This change adds in basic beacon support for WiFi.  It does this by creating the very minimal set of calls needed to create a connection, then using them to create a Socket.  We then graph a Serial device onto that socket and send the beacon JSON code once every 3 seconds.  Listen on port 7223 and you will hear it.  Here is a netcat command I use: `nc -lu 0.0.0.0 7223`.  While the code is stable enough to run it is not robust by any means.  In its current form it does not handle disconnects.  The only reason I am putting this for review now is because this will effectively unblock our app development for socket based communication.  And as that is developed I will continue making this code more robust so that it can handle disconnects etc.

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -23,6 +23,7 @@
 #include "esp8266.h"
 #include "macros.h"
 #include "mem_mang.h"
+#include "net/protocol.h"
 #include "printk.h"
 #include "serial_buffer.h"
 #include "taskUtil.h"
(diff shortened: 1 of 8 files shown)


src/devices/esp8266.c
```diff
@@ -637,34 +638,36 @@ static bool connect_cb(struct at_rsp *rsp, void *up)
...
-                         chan_id, ip_addr, dest_port, udp_port, udp_mode);
+        case PROTOCOL_UDP:
+                proto_str = "UDP";
                 break;
+        default:
+                proto_str = NULL;
         }
 
+        if (!proto_str) {
+                cmd_failure("esp8266_connect", "Invalid protocol");
```
brentpicasso: could we simply put the failure / return right under default: ?

src/drivers/esp8266_drv.c
```diff
@@ -119,44 +121,92 @@ static const char* get_channel_name(const size_t chan_id)
...
+{
+        if (i >= _MAX_CHANNELS)
+                return NULL;
 
-        ch->serial = s;
+        struct _channel* ch =state.channels + i;
+        if (!ch->serial) {
+                ch->serial = setup_channel_serial(i);
+                if (!ch->serial)
+                        return NULL;
```
brentpicasso: any reason to log these error conditions, or maybe they're logged higher up?

src/drivers/esp8266_drv.c
```diff
@@ -119,44 +121,92 @@ static const char* get_channel_name(const size_t chan_id)
...
 {
         if (chan_id >= ARRAY_LEN(state.channels)) {
                 pr_error_int_msg(_LOG_PFX "Channel id to big: ", chan_id);
                 return;
         }
 
-        struct _channel *ch = get_channel(chan_id);
+        struct _channel *ch = get_channel_for_use((unsigned int) chan_id);
         if (NULL == ch) {
                 pr_error(_LOG_PFX "No channel available.  Dropping\r\n");
```
brentpicasso: I see it here. :)

brentpicasso:
Looks pretty good. :shipit:

src/devices/esp8266.c
stieg: We could.  I will do that :).

src/drivers/esp8266_drv.c
stieg: Logged higher up.

Issue #549, opened 2016-05-02
MK2 CAN Doesn't Properly Handle No Mailbox Status with 0 Length Timeout


stieg:
It appears that when I implemented the CAN TX Timeout logic I failed to properly check that an outgoing mailbox was acquired in sending a message.  This can lead to a scenario where we report that we have successfully sent a CAN message when in fact we were unable to send it since no mailbox was available.  This only applies to situations where the timeout value is 0, otherwise the code will operate correctly.

Need to fix this logic so the return status informs the caller on whether a mailbox was successfully acquired and used for the outgoing message.

kosenko-max:
That would be great, thanks for finding it. Timeout 0 is what I plan to use now.

kosenko-max:
If timeout is 1 - it will cancel it silently and return OK. That is already not right. But if timeout is 0 - with a new patch it will not even cancel it, so it will stuck in the queue and not going to be canceled ever. 

I have a suggestion to expose mailbox numbers, ability to cancel messages, check their statuses and error codes in Lua. I know Lua is supposed to be simple, but this only leads to errors eating and CAN messaging system failure.

kosenko-max:
At least consider ability to deal with oldest message in the queue... Like read error code, retry, cancel it, read total number in the queue...

stieg:
> If timeout is 1 - it will cancel it silently and return OK.

Oh, yeah.  Disregard previous comment about that then.

> But if timeout is 0 - with a new patch it will not even cancel it, so it will stuck in the queue and not going to be canceled ever. 

Why would it be stuck in the queue forever exactly?  The only way I could see that happening is 

1.  No CAN bus connected.
2.  CAN bus is so saturated with high priority messages that it can't send this one.  At which point there is an issue with the CAN bus.

> I have a suggestion to expose mailbox numbers, ability to cancel messages, check their statuses and error codes in Lua.  I know Lua is supposed to be simple, but this only leads to errors eating and CAN messaging system failure.

Hrmm... I could adjust the return value on Lua to be a tuple so that if you pass it in a non-blocking timeout I could return the mailbox number.

kosenko-max:
CAN bus is easily not connected or remote party is not answering - it's a standard issue. And in such case there are nobody to cancel request. This CAN_CancelTransmit only called from a single place which you have skipped now, so nobody checks for this mailbox anymore and it's stuck occupied. And there are only 3 of them.

[...]

PR #551, opened 2016-05-02
Properly check for mailbox on no timeout CAN tx

stieg:
If a user called CAN Tx and specified a timeout value of 0, then we
would always return true.  This was bad beacuse its possible that
the message never made it into a mailbox, and thus will never be sent.
This trivial patch checks the return value of the CAN tx value so that
we return true if a mailbox was obtained, and false if none were available.
This will enable stronger confidence in our CAN Tx abilities.

Fixes Issue #549


brentpicasso:
:shipit:

PR #554, opened 2016-05-02
Add hooks for the poorly designed ESP8266 URC AT msgs

stieg:
There are a few URC messages that we need to process to maintain internal stability that aren't presented in a proper AT URC fashion.  They are more along the lines like the modem blurting out what happening when it happens instead of being civil.  In example the modem will blurt out `3,CONNECTED` when a new connection is opened on port 3.  I can't easily handle these with the URC handler because that depends on a prefix (like `+IPD:`, which is the prefix seen for incoming data URCs).  So this creates a generic URC handler and we define a method to process these URCs and call the appropriate hook methods to ensure things are set correctly.

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -26,6 +26,7 @@
 #include "net/protocol.h"
 #include "printk.h"
 #include "serial_buffer.h"
+#include "str_util.h"
 #include "taskUtil.h"
 
 #include <stdbool.h>
@@ -57,6 +58,7 @@ static struct {
         struct at_info *ati;
         void (*init_cb)(enum dev_init_state); /* STIEG: Put in own struct? */
         enum dev_init_state init_state;
+        struct esp8266_event_hooks hooks;
 } state;
 
 static void cmd_failure(const char *cmd_name, const char *msg)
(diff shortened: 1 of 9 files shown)


src/devices/esp8266.c
```diff
@@ -71,16 +73,58 @@ static void cmd_failure(const char *cmd_name, const char *msg)
...
+
+        /* Now look for a message with a comma */
+        char* comma = strchr(msg, ',');
+        if (!comma)
+                return false;
+
+        *comma = '\0';
+        const char* m1 = msg;
+        const char* m2 = ++comma;
+        if (STR_EQ(m2, "CONNECT")) {
```
brentpicasso: are we just assuming the part before the comma is the connection number?

brentpicasso:
So, the idea of the hooks is to provide extra handlers for unusual / unconventional type of responses. Why not modify the core response handler to handle any kind of message response pattern?

src/devices/esp8266.c
stieg: Yep.

stieg:
> Why not modify the core response handler to handle any kind of message response pattern?

Primarily because that modification is effectively what this is.  A well designed URC begins with some form of a command prefix (in example, our data URC begins with `+IPD,`) and I use that to determine if its a URC or not.  But since these URCs begin with numbers or just standard strings, its harder to handle them.  I suppose I could register the `WIFI DISCONNECT` statement as a URC, but the socket versions just are a no go :(.

brentpicasso:
is this PR still relevant given conversations in slack?

stieg:
Yes

src/devices/esp8266.c
brentpicasso: we can probably safely assume it's a number, and if not, it falls back to zero in the atoi. Thanks.

brentpicasso:
Good to go. :shipit:

src/devices/esp8266.c
stieg: Yep.  I have yet to see another one of their messages with a comma in it.  But if you see one...

PR #557, opened 2016-05-03
Fixed LED polarity for production board. remapped LED colors

brentpicasso:
Green = GPS
Blue = Telemetry activity
Red = error

### DIFF
--- a/platform/rct/hal/LED_stm32/led_device_stm32.c
+++ b/platform/rct/hal/LED_stm32/led_device_stm32.c
@@ -55,9 +55,9 @@ static bool led_set_level(struct led_data *ld, const bool on)
 
         ld->level = on;
         if (on) {
-                GPIO_SetBits(GPIO_PORT, ld->mask);
-        } else {
                 GPIO_ResetBits(GPIO_PORT, ld->mask);
+        } else {
+                GPIO_SetBits(GPIO_PORT, ld->mask);
         }
 
         return true;
@@ -105,7 +105,7 @@ bool led_device_init(void)
 
         gpio_conf.GPIO_Speed = GPIO_Speed_50MHz;
         gpio_conf.GPIO_Mode = GPIO_Mode_OUT;
-        gpio_conf.GPIO_OType = GPIO_OType_PP;
+        gpio_conf.GPIO_OType = GPIO_OType_OD;
         gpio_conf.GPIO_Pin = 0;
 
         for (size_t i = 0; i < ARRAY_LEN(leds); ++i) {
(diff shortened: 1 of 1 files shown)


platform/rct/hal/LED_stm32/led_device_stm32.c
```diff
@@ -55,9 +55,9 @@ static bool led_set_level(struct led_data *ld, const bool on)
 
         ld->level = on;
         if (on) {
-                GPIO_SetBits(GPIO_PORT, ld->mask);
-        } else {
                 GPIO_ResetBits(GPIO_PORT, ld->mask);
+        } else {
+                GPIO_SetBits(GPIO_PORT, ld->mask);
```
stieg: Seems like we should just move to an open collector instead of leaving the push pull setup.  Not sure if it makes much (if any) difference internally to the STM chip.

platform/rct/hal/LED_stm32/led_device_stm32.c
```diff
@@ -43,9 +43,9 @@ static struct led_data {
         uint16_t mask;
         bool level;
 } leds[] = {
-        {LED_ERROR,     GPIO_Pin_2, 0},
+        {LED_ERROR,     GPIO_Pin_0, 0},
         {LED_GPS,       GPIO_Pin_1, 0},
-        {LED_TELEMETRY, GPIO_Pin_0, 0},
+        {LED_TELEMETRY, GPIO_Pin_2, 0},
```
stieg: So the new board has different mappings for the LEDs?  Effectively the pins swapped around?

brentpicasso: Yes, the previous boards were actually incorrect. I moved it around and also chose new colors.

platform/rct/hal/LED_stm32/led_device_stm32.c
brentpicasso: Yes, but PP for now so it works with your older board. we can change it later.

stieg: Wise.  Ok lets file an issue so we don't forget it and go this route.

platform/rct/hal/LED_stm32/led_device_stm32.c
stieg: Got it.

stieg:
:shipit: pending a filed issue to move us to OpenCollector once the now incorrect boards are no longer in service.

brentpicasso:
Filed #559 to track followup work.

PR #561, opened 2016-05-04
Fix ORE watchdog issue in MK2

stieg:
Well... we thought we had fixed the ORE issue in MK2.  However there was
a bug where we were not actually reading the status and rdr registers
of the UART that sent the ORE interrupt.  Instead we were always reading
the status and rdr registers of UART1.  Oops.  This fixes that.

This patch also improves how we handle the ORE case a bit by making the
check of the ORE flag part of the ISR prior to reading the data from the
rdr.  This can save a step or two.


brentpicasso:
This is great. :shipit:

Issue #562, opened 2016-05-04
Backport the fix for the ORE handler in MK2 to 2.9.1


stieg:
#561 is a candidate for 2.9.1.  Need to back-port it.

brentpicasso:
:shipit:

stieg:
Heh... this is the ticket.  I assume you mean the PR.  Will merge.

stieg:
Fixed

PR #563, opened 2016-05-05
Backport the MK2 ORE fix to 2.9.1

stieg:
As explained in PR #561 there was a bug with ORE resetting in MK2
where we were only resetting USART1 instead of resetting the handle
that was causing the ORE to occur.  This change fixes that by using
the provided handle to do the steps needed to clear the ORE flag.

Trivial patch.  Addresses issue #562


stieg:
:shipit: per @brentpicasso 's comment in the issue (#562)

Issue #553, opened 2016-05-02
Remove unused CANtx queues from MK2 HAL


stieg:
Because we don't use them.

stieg:
Fixed

PR #564, opened 2016-05-05
Remove unused CANtx queue from MK2 CAN HAL

stieg:
Because it isn't used in any way.  So stop wasting resources and save the
planet (or the heap anyways).

Issue: #553

### DIFF
--- a/platform/mk2/hal/CAN_stm32/CAN_device_stm32.c
+++ b/platform/mk2/hal/CAN_stm32/CAN_device_stm32.c
@@ -33,14 +33,10 @@
 
 #include <stdint.h>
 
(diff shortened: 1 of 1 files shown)


brentpicasso:
In retrospect it would be nice to queue up the messages for transmit by a separate task, but it seems like overkill given the available slots provided by the STM32f4.  Good call.

:shipit:

stieg:
> In retrospect it would be nice to queue up the messages for transmit by a separate task, but it seems like overkill given the available slots provided by the STM32f4. Good call.

Agreed, but until we add that support no point in wasting the resources.  If this is an issue our customers face we should file a ticket to get it fixed.

PR #567, opened 2016-05-05
improve rx buff code

stieg:
Another patch built on top of the esp8266_drv state machine improvements.  This patch improves our rx buffer code so that is it better at quickly draining messages out of serial devices and handling partial, complete, and overflow message scenarios.

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -268,7 +268,7 @@ static void rx_data_cb(int chan_id, size_t len, const char* data)
 
         /*
          * Check that we actually have a call back set to handle the
-         * incomming data.  If not then the socket stays open until it is
+         * incoming data.  If not then the socket stays open until it is
          * closed by our WiFi host.
          */
         if (!state.comm.new_conn_cb) {
@@ -279,7 +279,7 @@ static void rx_data_cb(int chan_id, size_t len, const char* data)
         state.comm.new_conn_cb(ch->serial);
 
         /*
-         * Now that the upper layer has been informed about the incomming
+         * Now that the upper layer has been informed about the incoming
          * data, start sending it said data.  It will unblock and read this
          * data very shortly.
          */
@@ -458,7 +458,7 @@ static void init_wifi()
         cmd_started();
 
         /*
-         * Now register the callback for incomming data.  Safe to do
+         * Now register the callback for incoming data.  Safe to do
          * before init has completed b/c this invokes no modem calls.
          * This should only fail if there was an issue with space or config.
          */
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -324,9 +324,7 @@ static void check_data()
                  * connection is still active.  If not, dump the data
                  */
                 if (!ch->connected) {
-                        char c;
-                        xQueueHandle queue = serial_get_tx_queue(ch->serial);
-                        while(pdTRUE == xQueueReceive(queue, &c, 0));
+                        serial_purge_tx_queue(ch->serial);
                         ch->tx_chars_buffered = 0;
                         continue;
                 }
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -287,6 +287,7 @@ static void rx_data_cb(int chan_id, size_t len, const char* data)
                 xQueueSend(q, data + i, portMAX_DELAY);
         }
         /* pr_info("\"\r\n"); */
+
         cmd_set_check(CHECK_DATA);
 }
 
@@ -309,6 +310,8 @@ static void _send_data_cb(int bytes_sent)
  */
 static void check_data()
 {
+        cmd_check_complete(CHECK_DATA);
+
         for (size_t i = 0; i < ARRAY_LEN(state.comm.channels); ++i) {
                 struct channel *ch = state.comm.channels + i;
                 const size_t size = ch->tx_chars_buffered;
@@ -344,8 +347,6 @@ static void check_data()
                 ch->tx_chars_buffered = 0;
                 return;
         }
-
-        cmd_check_complete(CHECK_DATA);
 }
 
 /**
@@ -475,11 +476,12 @@ static void init_wifi()
  */
 static void check_init()
 {
+        cmd_check_complete(CHECK_INIT);
+
         pr_info(LOG_PFX "Checking Init\r\n");
         switch(state.device.init_state) {
         case DEV_INIT_STATE_READY:
                 /* Then we are where we want to be and are done */
-                cmd_check_complete(CHECK_INIT);
                 return;
         default:
                 /* STIEG: TODO Handle reset here when implemented. */
@@ -505,7 +507,6 @@ static void get_client_ap_cb(bool status, const struct esp8266_client_info *ci)
 {
         cmd_completed();
         cmd_set_check(CHECK_CLIENT);
-        cmd_set_check(CHECK_DAEMON);
 
         memcpy(&state.client.info, ci, sizeof(struct esp8266_client_info));
         *state.client.info.ip = 0;
@@ -597,7 +598,7 @@ static void client_try_join_ap()
                 /* Then we need to backoff */
                 const tiny_millis_t sleep_len =
                         state.client.next_join_attempt - getUptime();
-                cmd_sleep_until(CHECK_CLIENT, sleep_len);
+                cmd_sleep(CHECK_CLIENT, sleep_len);
         } else {
                 /* If here, then its time to do work */
                 state.client.next_join_attempt =
@@ -613,6 +614,8 @@ static void client_try_join_ap()
  */
 static void check_client()
 {
+        cmd_check_complete(CHECK_CLIENT);
+
         pr_info(LOG_PFX "Checking Client\r\n");
 
         /* First check that we are initialized */
@@ -654,7 +657,6 @@ static void check_client()
                 }
 
                 /* If here, we are done */
-                cmd_check_complete(CHECK_CLIENT);
                 return;
         } else {
                 /* Client should be inactive.  Disable as needed */
@@ -664,12 +666,10 @@ static void check_client()
                          * STIEG: TODO: Implement Leave AP call. We are done
                          * for now since code not implemented.
                          */
-                        cmd_check_complete(CHECK_INIT);
                         return;
                 }
 
                 /* Then config and reality align.  Done */
-                cmd_check_complete(CHECK_CLIENT);
                 return;
         }
 }
@@ -700,11 +700,13 @@ static void setup_daemon()
  */
 static void check_daemon()
 {
+        cmd_check_complete(CHECK_DAEMON);
+
         pr_info(LOG_PFX "Checking Daemon\r\n");
 
         /* First check that we are initialized */
         if (!device_initialized()) {
-                cmd_sleep(CHECK_CLIENT, CLIENT_BACKOFF_MS);
+                cmd_sleep(CHECK_DAEMON, CLIENT_BACKOFF_MS);
                 return;
         }
 
@@ -715,7 +717,7 @@ static void check_daemon()
         }
 
         /* Then its listening.  We are done */
-        cmd_check_complete(CHECK_DAEMON);
+        return;
 }
 
 
@@ -796,9 +798,8 @@ bool esp8266_drv_init(struct Serial *s, const int priority,
 
         state.comm.new_conn_cb = new_conn_cb;
 
-        /* Set all check flags so we check the whole system */
-        for (size_t i = 0; i < __NUM_CHECKS; ++i)
-                cmd_set_check(i);
+        /* Set the task loop to check init first */
+        cmd_set_check(CHECK_INIT);
 
         static const signed char task_name[] = TASK_THREAD_NAME;
         const size_t stack_size = TASK_STACK_SIZE;
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -91,13 +91,13 @@ struct comm {
  *
  * Note that each check represents a tiny state machine.  These
  * state machines have interdependeinces upon one another. In
- * example the client wifi (CHECK_CLIENT) requires that the wifi
+ * example the client wifi (CHECK_WIFI_CLIENT) requires that the wifi
  * subsystem (CHECK_INIT) be initialized. Thus we need a priority
  * with the lowest value being the highest priority.
  */
 enum check {
         CHECK_INIT,
-        CHECK_CLIENT,
+        CHECK_WIFI_CLIENT,
         CHECK_SERVER,
         CHECK_DATA,
         __NUM_CHECKS, /* Always the last */
@@ -353,7 +353,7 @@ static void check_data()
 static void client_wifi_disconnect_cb()
 {
         /* Need to check our client now that a change has occured */
-        cmd_set_check(CHECK_CLIENT);
+        cmd_set_check(CHECK_WIFI_CLIENT);
         cmd_set_check(CHECK_DATA);
 
         memset(&state.client.info, 0, sizeof(state.client.info));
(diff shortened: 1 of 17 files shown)


src/drivers/esp8266_drv.c
```diff
@@ -31,82 +31,145 @@
...
-        struct _daemon daemon;
-        bool cmd_ip;
-        enum _cmd cmd;
-        enum dev_init_state dev_state;
-        tiny_millis_t cmd_sleep_until;
+        struct channel channels[MAX_CHANNELS];
+};
+
+/**
+ * These enums define the various thecks we can perform.
```
brentpicasso: spelling.

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +190,163 @@ static const char* get_channel_name(const size_t chan_id)
...
         if (!is_valid_socket_channel_id(i))
                 return NULL;
 
-        struct _channel* ch =state.channels + i;
-        if (!ch->serial) {
-                ch->serial = setup_channel_serial(i);
-                if (!ch->serial)
-                        return NULL;
-        }
+        struct channel* ch =state.comm.channels + i;
```
brentpicasso: whitespace after =

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +190,163 @@ static const char* get_channel_name(const size_t chan_id)
...
-        pr_info_str_msg(_LOG_PFX "Message: ", data);
-        /* STIEG: Make this not portMAX_DELAY to prevent stalling? */
+        state.comm.new_conn_cb(ch->serial);
+
+        /*
+         * Now that the upper layer has been informed about the incomming
+         * data, start sending it said data.  It will unblock and read this
+         * data very shortly.
+         */
+        /* pr_info_int_msg(LOG_PFX "rx_data_cb data len: ", len); */
```
brentpicasso: commented out code.

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +190,163 @@ static const char* get_channel_name(const size_t chan_id)
...
+        /* pr_info(LOG_PFX "rx_data_cb data: \""); */
         xQueueHandle q = serial_get_rx_queue(ch->serial);
-        for (size_t i = 0; i < len; ++i)
+        for (size_t i = 0; i < len; ++i) {
+                /* pr_info_int(data[i]); */
+                /* pr_info_char('|'); */
                 xQueueSend(q, data + i, portMAX_DELAY);
+        }
+        /* pr_info("\"\r\n"); */
```
brentpicasso: ditto.

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +190,163 @@ static const char* get_channel_name(const size_t chan_id)
...
+         * Now that the upper layer has been informed about the incomming
+         * data, start sending it said data.  It will unblock and read this
+         * data very shortly.
+         */
+        /* pr_info_int_msg(LOG_PFX "rx_data_cb data len: ", len); */
+        /* pr_info(LOG_PFX "rx_data_cb data: \""); */
         xQueueHandle q = serial_get_rx_queue(ch->serial);
-        for (size_t i = 0; i < len; ++i)
+        for (size_t i = 0; i < len; ++i) {
+                /* pr_info_int(data[i]); */
```
brentpicasso: applies throughout.

src/drivers/esp8266_drv.c
stieg: Me bad english?  That unpossible.

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +190,163 @@ static const char* get_channel_name(const size_t chan_id)
...
+                /* If the size is 0, nothing to send */
+                if (0 == size)
+                        continue;
+
+                /*
+                 * If here, then we have data to send.  Check that the
+                 * connection is still active.  If not, dump the data
+                 */
+                if (!ch->connected) {
+                        char c;
```
brentpicasso: I'm uncertain about this part of the code. We receive data from the queue, store it in c, but seemingly do nothing with the data?

src/drivers/esp8266_drv.c
```diff
@@ -309,211 +466,294 @@ static void init_wifi()
...
 
+/**
+ * Callback that gets invoked when the get_client_ip command completes.
+ */
 static void get_client_ip_cb(bool status, const char* ip)
 {
+        cmd_completed();
+        cmd_set_check(CHECK_CLIENT);
+
+        /* STIEG: This is a HACK.  IP Info should be its own struct */
```
brentpicasso: Do we have a ticket to track this?

src/serial/rx_buff.c
```diff
@@ -39,85 +46,133 @@ void rx_buff_clear(struct rx_buff *rxb)
...
-
         xQueueHandle h = serial_get_rx_queue(s);
-
-        for(; rxb->idx < rxb->cap; ++rxb->idx) {
-                if (!xQueueReceive(h, rxb->buff + rxb->idx, ticks_to_wait)) {
-                        /* If here, we timed out on receiving a message */
-                        rx_buff_clear(rxb);
-                        return NULL;
+        char c = 1;
+        /* pr_info(LOG_PFX "RX Chars: \""); */
```
brentpicasso: commented out code.

src/serial/rx_buff.c
```diff
@@ -39,85 +46,133 @@ void rx_buff_clear(struct rx_buff *rxb)
...
 
-        /* If there is a \n after the \r, remove it */
-        char c;
-        if (xQueuePeek(h, &c, 0) && '\n' == c)
-                xQueueReceive(h, &c, 0);
-
-        if (rxb->idx >= rxb->cap) {
-                /* Overflow scenario */
-                pr_warning(_LOG_PFX "Overflow!");
+        /*
```
brentpicasso: do we have tests for these scenarios?

src/serial/serial.c
```diff
@@ -186,16 +186,31 @@ char serial_get_c(struct Serial *s)
         return serial_get_c_wait(s, &c, portMAX_DELAY) ? c : 0;
 }
 
-int serial_get_line_wait(struct Serial *s, char *l, const size_t len,
+/**
+ * Reads in a line from a serial device delimeted by \n.  The data is
+ * written to buff BUT MAY NOT BE NULL TERMINATED.  NULL termination is the
```
brentpicasso: what's the rationale for not null terminating the returned line of data?

src/drivers/esp8266_drv.c
stieg: So we emulate a Serial device over a socket in this code. Since serial devices don't check open/closed state, we need to handle the case where a Serial device is still open but has no backing connection.  This is that case.  If the connection is broken for whatever reason and we wait for it to come back (which it won't), we will block indefinitely, causing a permanent stall and eventually a watchdog.  This code says that if we have lost our connection, do exactly what a serial device would do with a broken connection: send data to a non-existent endpoint, which is effectively the same as dropping the data.  This prevents stalls and keeps the overall system in good health.  Once we use proper sockets on top of Serial instead of Serial on top of sockets this can die... but until then this is a necessary evil to keep our changes needed small.

src/drivers/esp8266_drv.c
stieg: I do internally.  I have lots of stuff on an internal doc for what needs to get done.  I can file these in GH tickets if you think it will help.

brentpicasso: Yes, if individual things ends up turning into PRs. thanks!

src/drivers/esp8266_drv.c
brentpicasso: I see that now. it might be helpful to extract this into a function named similar to "purge queue" to clarify what's going on, even if it only gets used once.

src/drivers/esp8266_drv.c
```diff
@@ -31,82 +31,150 @@
...
-        struct esp8266_client_info info;
+struct client {
         const struct wifi_client_cfg *config;
-        tiny_millis_t next_check_ms;
+        struct esp8266_client_info info;
+        tiny_millis_t next_join_attempt;
 };
 
-struct _daemon {
+struct daemon {
```
brentpicasso: I know the 'daemon' naming was introduced previously, but want to suggest it could be named better. Is it truly a background process as the name suggests? https://en.wikipedia.org/wiki/Daemon_(computing)

If you agree we can file a separate issue.

src/drivers/esp8266_drv.c
stieg: Done.

src/serial/rx_buff.c
stieg: Filed #591 to track that work.

src/serial/serial.c
stieg: Because I need to sometimes know what the NULL terminator was.  I need to handle it differently if its a `\r\n` versus a `\0` (like you can see on some netcat commands).  By not modifying the serial here, and instead handling this higher up, I gain more flexibility in my decision making process.  Plus this simplifies the serial code more by making it do less (which is usually a good practice where reasonable).

src/drivers/esp8266_drv.c
stieg: It is indeed a background process on the ESP8266 device.  What name would you prefer and why?

src/drivers/esp8266_drv.c
stieg: Will do.

src/drivers/esp8266_drv.c
brentpicasso: The ESP8266 calls it a 'server' so... perhaps that might be a good match to be consistent with their own naming.

src/drivers/esp8266_drv.c
```diff
@@ -31,82 +31,150 @@
...
+        tiny_millis_t check[__NUM_CHECKS];
+};
+
+static struct {
+        struct device device;
+        struct client client;
+        struct daemon daemon;
+        struct comm comm;
+        struct cmd cmd;
 } state;
```
brentpicasso: state name could be a little more explicit. like esp8266_state might help.

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +195,157 @@ static const char* get_channel_name(const size_t chan_id)
...
+                return ch;
 
-        /* Prep the channel for use */
-        ch->tx_chars_buffered = 0;
-        ch->in_use = true;
-        ch->connected = false; /* Not connected until we say so */
-
-        return ch;
+        /* If here, channel has no serial.  Try to set one up */
+        ch->serial = setup_channel_serial(i);
```
brentpicasso: could these serial assignments be done in advance when everything is set up upon system init?

src/drivers/esp8266_drv.c
```diff
@@ -127,102 +195,157 @@ static const char* get_channel_name(const size_t chan_id)
...
+         * we set the created_externally flag here.  This way
+         * we know to reap the channel when the connection disappears.
+         */
+        ch->created_externally = true;
+        ch->in_use = true;
+        ch->connected = true;
+
+        /*
+         * Check that we actually have a call back set to handle the
+         * incomming data.  If not then the socket stays open until it is
```
brentpicasso: spelling; "incoming "

src/drivers/esp8266_drv.c
```diff
@@ -309,211 +465,294 @@ static void init_wifi()
...
 
 /**
- * Determines what the next command should be given that we have none that
- * need to explicitly be done.
+ * This is the block of logic that manages our WiFi client state.
+ * Its responsible for keeping the client status as close to the
+ * configuration as is reasonably possible.
  */
-static enum _cmd determine_cmd()
+static void check_client()
```
brentpicasso: suggest making more explicit...

check_client_mode
check_station_mode

src/drivers/esp8266_drv.c
```diff
@@ -309,211 +465,294 @@ static void init_wifi()
...
+        }
+
+        /*
+         * First check if we have reasonably fresh client info. Have to
+         * have it before we can make any decisions.
+         */
+        const struct esp8266_client_info *ci = &state.client.info;
+        if (0 == ci->snapshot_time) {
+                get_client_ap();
+                return;
```
brentpicasso: will check_client be triggered again when client_ap data is retrieved?

brentpicasso: Yes, i see that it will in the callback.

src/serial/serial.c
brentpicasso: ok.

brentpicasso:
No further comments.

src/drivers/esp8266_drv.c
stieg: I shortened it because it was a static variable and because C is a spartan language.

src/drivers/esp8266_drv.c
stieg: You mean pre-allocation.  Sure... but we are pretty tight on RAM... what's the point of allocating what might not be used.?

brentpicasso: ok, acceptable.

src/drivers/esp8266_drv.c
brentpicasso: Chap 4 of the Kernel style guide suggests that C is a spartan language, but global variables should be descriptive. While state isn't a global variable, it's scope is in that global direction which I think supports the more explicit naming of the variable and help new people understand it better.

src/drivers/esp8266_drv.c
stieg: I instead opted to create a `serial_purge_{tx,rx}_queue` set of methods and use that.  Should be clearer now.  Do you think it needs more clarification?

src/serial/rx_buff.c
brentpicasso: Ok

src/drivers/esp8266_drv.c
brentpicasso: Looks good.

On Mon, May 9, 2016 at 11:06 AM, Andrew Stiegmann <notifications@github.com>
wrote:

> In src/drivers/esp8266_drv.c
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/567#discussion_r62544584>
> :
>
> > +        cmd_check_complete(CHECK_DATA);
> > +
> > +        for (size_t i = 0; i < ARRAY_LEN(state.comm.channels); ++i) {
> > +                struct channel *ch = state.comm.channels + i;
> > +                const size_t size = ch->tx_chars_buffered;
> > +
> > +                /* If the size is 0, nothing to send */
> > +                if (0 == size)
> > +                        continue;
> > +
> > +                /*
> > +                 * If here, then we have data to send.  Check that the
> > +                 * connection is still active.  If not, dump the data
> > +                 */
> > +                if (!ch->connected) {
> > +                        char c;
>
> I instead opted to create a serial_purge_{tx,rx}_queue set of methods and
> use that. Should be clearer now. Do you think it needs more clarification?
>
> —
> You are receiving this because you were assigned.
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/567/files/5168360ab731d9cd125f95e04299ba28ef17cc6d#r62544584>
>



-- 
Brent Picasso

src/drivers/esp8266_drv.c
stieg: Fixed

stieg:
Good to go @brentpicasso ?

brentpicasso:
still outstanding comments on outdated diffs

src/drivers/esp8266_drv.c
brentpicasso: unresolved

src/drivers/esp8266_drv.c
brentpicasso: unresolved

src/drivers/esp8266_drv.c
brentpicasso: looks to be resolved

brentpicasso:
:shipit:

stieg:
Thank you for working through that monster review @brentpicasso

On Mon, May 9, 2016 at 1:50 PM, Brent Picasso <notifications@github.com>
wrote:

> [image: :shipit:]
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/567#issuecomment-217954351>
>



-- 
Autosport Labs
*Open Source Motor Sports*

PR #570, opened 2016-05-05
Move serial_device into the HAL; Improve our serial_device enum

stieg:
Two changes here.  First we move serial_device into the HAL from the
common src region.  This is because the serial mappings differ between
devices.  Fancy that.

This change also improves upon our serial device naming convention.
We add a SERIAL_WIFI to represent the wifi device and renames the
SERIAL_WIRELESS to SERIAL_BLUETOOTH since that is more accurate.


brentpicasso:
:shipit:

Issue #556, opened 2016-05-03
RCT: Make USB and CAN co-exist


brentpicasso:
Two items needed for USB and CAN simultaneous operation:

1. The STM device driver appears to be using some GPIOs for pullups that are needed for CAN operation.
hw_config.c  -  line 59 - Set_System() is where it happens

2. Interrupt handling. it seems that USB and CAN share the same interrupt.  Appears this is already handled in the common isr handler under common_irq_stm32.

PR #594, opened 2016-05-10
RCT: USB interferes wih CAN

brentpicasso:
unused USB pullup interferes with CAN as USB pullup activation squats on one of the CAN lines.


Jeff-Ciesielski:
:shipit: ST just has a long and sordid history with CAN and USB not playing nicely...

brentpicasso:
resolved #556

Issue #522, opened 2016-04-20
Add lua scripting, Bt, cellular and wifi capabilities to the getCapabilities command


ryandoherty:
Our standardized method of determining if a RC/RCP device can support certain commands or configurations is to query via `getCapabilities`. Currently RC does not support lua scripting or BT and a RCP device may or may not support WiFi/cellular configuration. The RC app needs to know this to determine what configuration screens to show.

The new sections to add to the response to `getCapabilities` should be:

* `lua`
* `bt`
* `wifi` 
* `cellular`

stieg:
For `lua` look for the `"script"` variable under `"db"`.  If its there then we support Lua.  Else we do not.  I will add a communications section for you.

stieg:
This is now done.

PR #601, opened 2016-05-11
Links in Status

stieg:
This change adds to our getStatus API by appending a "links" object.  This object tells the system what links are available for use on the system.

Fixes Issue #522


brentpicasso:
Looks like only this changeset is relevant:

https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/601/commits/e2fccb82ff1692c10fdb15228752a0e409b6db54

If so, looks good. :shipit:

stieg:
More like the last 3 changes are important... but they are small and reasonable.

PR #602, opened 2016-05-11
Murder RCT Watchdogging

stieg:
This PR builds upon #601 , so review that first before you review this.  Else you will be reviewing both!

This patch series fixes the watchdogging that Brent saw when trying to test RCT with WiFi.  Here is what was happening:

* We were running out of memory.  I fixed this by:
** Reducing stack sizes.  There is still plenty to reduce here... I'm just being conservative since it works
** Reduced the length of the predictive time buffer... because it saved us 1100+ bytes of RAM in the .bss
* We had a priority inversion in our USB stack.  No priority may be higher than that of FreeRTOS.  So I fixed that.
* We had a illegal de-reference in the serial_create_buffer code.  Because I put a command call on the wrong line like a dummy.  This is fixed.


stieg:
Oops... opened this against the wrong branch.  Will reopen a new one

PR #600, opened 2016-05-10
Add support for the ESP8266 SoftAP

stieg:
Includes the patches needed to enable support for the softAP functionality on the esp8266.  This will allow us to control the main aspects of the AP (name, password, channel, encryption) as well as allows us to configure it through our APIs.  Also added unit tests for the API calls as well as for the default reset behavior.  Fixes #592

PR #603, opened 2016-05-11
P/stieg/murder rct watchdogging

stieg:
#This PR builds upon PR #600  , so review that first before you review this. Else you will be reviewing both!

This patch series fixes the watch-dogging that Brent saw when trying to test RCT with WiFi. Here is what was happening:

* We were running out of memory. I fixed this by:
 * Reducing stack sizes. There is still plenty to reduce here... I'm just being conservative since it works.
 * Reduced the length of the predictive time buffer... because it saved us 1100+ bytes of RAM in the .bss
* We had a priority inversion in our USB stack. No priority may be higher than that of FreeRTOS. So I fixed that.
* We had a illegal de-reference in the serial_create_buffer code. Because I put a command call on the wrong line like a dummy. This is fixed.


platform/rct/hal/usb_stm32/hw_config.c
```diff
@@ -213,6 +214,7 @@ void USB_Interrupts_Config(void)
   NVIC_Init(&NVIC_InitStructure);
 
 #elif defined(STM32F37X)
+#error "Why are we here?"
   /* Enable the USB interrupt */
```
Jeff-Ciesielski: If we're going to just throw an error here, we need to gut these.  The linenoise doesn't add anything if we're not going to port to these platforms in the future.

include/devices/esp8266.h
```diff
@@ -79,9 +84,9 @@ bool esp8266_get_op_mode(void (*cb)(bool, enum esp8266_op_mode));
 struct esp8266_client_info {
         tiny_millis_t snapshot_time;
         bool has_ap;
-        char ssid[24];
-        char mac[18];
-        char ip[16];
+        char ssid[ESP8266_SSID_LEN_MAX];
+        char mac[ESP8266_MAC_LEN_MAX];
+        char ip[ESP8266_IPV4_LEN_MAX];
 };
```
Jeff-Ciesielski: Always good to ditch magic numbers 👍

platform/mk2/capabilities.h
```diff
@@ -8,18 +8,24 @@
...
-#define VIRTUAL_CHANNEL_SUPPORT
-#define SDCARD_SUPPORT
-#define CELLULAR_SUPPORT
-#define WIFI_SUPPORT
+#define BLUETOOTH_SUPPORT	1
+#define CELLULAR_SUPPORT	1
+#define LUA_SUPPORT		1
+#define SDCARD_SUPPORT		1
+#define USB_SERIAL_SUPPORT	1
+#define VIRTUAL_CHANNEL_SUPPORT	1
```
Jeff-Ciesielski: Might want to double check that VIRTUAL_CHANNEL_SUPPORT is actually being defined here and not VIRTUAL_CHANNEL_SUPPORT1

platform/rct/capabilities.h
```diff
@@ -7,12 +7,26 @@
...
-#define USB_SERIAL_SUPPORT
 
-//configuration
+/* Support Flags */
+#define BLUETOOTH_SUPPORT	0
+#define CELLULAR_SUPPORT	0
+#define LUA_SUPPORT		0
+#define SDCARD_SUPPORT		0
+#define USB_SERIAL_SUPPORT	1
+#define VIRTUAL_CHANNEL_SUPPORT	0
```
Jeff-Ciesielski: Same as above

src/logger/loggerApi.c
```diff
@@ -1712,3 +1724,59 @@ int api_get_wifi_client_cfg(struct Serial *serial, const jsmntok_t *json)
...
+
+        json_bool(serial, "active", cfg->active, 1);
+        json_string(serial, "ssid", cfg->ssid,1);
+        json_string(serial, "passwd", cfg->password,1);
+        json_int(serial, "channel", cfg->channel,1);
+        json_string(serial, "encryption",
+                    wifi_api_get_encryption_str_val(cfg->encryption), 0);
+
+        json_objEnd(serial, 0);
+        json_objEnd(serial, 0);
```
Jeff-Ciesielski: Why the double object end here?

stieg: I think I may have a copy/pasta error on line 1768 and I added that to fix that issue.  That was bugging me late at night. Will fix.

src/logger/loggerTaskEx.c
```diff
@@ -222,7 +222,7 @@ void loggerTaskEx(void *params)
                 if (!g_loggingShouldRun && is_logging) {
                         logging_stopped();
                         const LoggerMessage logStopMsg = getLogStopMessage();
-#if defined(SDCARD_SUPPORT)
+#if SDCARD_SUPPORT
                         queue_logfile_record(&logStopMsg);
 #endif
                         queueTelemetryRecord(&logStopMsg);
```
Jeff-Ciesielski: We're starting to get a sea of ifdefs... Should discuss the long term implications and ways to mitigate.

platform/mk2/capabilities.h
stieg: This is GitHub rendering poorly on a browser.  There is a \t character between there.

stieg: The compile would fail if the VIRTUAL_CHANNEL_SUPPORT variable were not defined.

platform/rct/hal/usb_stm32/hw_config.c
stieg: Sure.  Just did that to confirm my sanity.  Will gut.

src/tasks/wifi.c
```diff
@@ -52,6 +55,8 @@
 #define STACK_SIZE		256
 /* Make all task names 16 chars including NULL char */
 #define THREAD_NAME		"WiFi Task      "
+/* The highest channel WiFi can use (USA) */
+#define WIFI_MAX_CHANNEL	11
```
Jeff-Ciesielski: If we're doing channel clamping, we're going to need to build in some more sophisticated region management.  If we're selling into ETSI regions, we'll need to obey their rules.

platform/mk2/capabilities.h
Jeff-Ciesielski: 👍

src/logger/loggerApi.c
brentpicasso: The double is likely to close the json object so it is balanced. I would
double check .

Brent Picasso
Autosport Labs

brentpicasso:
No additional comments above what was on #600. Looking forward to testing this!

src/logger/loggerTaskEx.c
stieg: Yeah... there is a ticket out to murder those.  But its a low priority for the moment as delivery is key.

src/tasks/wifi.c
stieg: Yep.  It also depends on what the esp8266 supports, which currently is only 1 - 11 AFAICT.

src/logger/loggerApi.c
stieg: This is sane.  The first closes the inner wifi_ap_cfg object.  The latter closes the external object.

Issue #585, opened 2016-05-09
Move definition of WiFi hook methods to one large struct


stieg:
Currently we have two different points where we set the callback hooks for the WiFi driver.  We should consolidate this into a single struct and call point to make it clean and easy to understand and update.  This ticket tracks doing just that.

PR #605, opened 2016-05-12
Retool how we handle callbacks on esp8266 driver and add functionality

stieg:
This change re-tools how we handle registering callbacks against the
esp8266 driver.  We now register all callbacks using a single call and
pass a struct in with the pointers.  This is both flexible and less
invasive on the API.  This also centralizes where this action happens
instead of happening in two different spots like we previously did.

This change also reduces the number of callbacks by making the
callbacks more general.  Instead of doing specific actions, the
callbacks are invoked when general state change occurs, and relevant
information is passed in as part of the parameters.  This reduces the
number of callbacks and reduces a little bit of code.

Finally this change also handles some previously unhandled callbacks
that are only seen during driver init.  Now we act upon them to ensure
our state is always up to date.

Issues: #599, #585

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -85,9 +85,12 @@ static bool sparse_urc_cb(char* msg)
 {
         msg = strip_inline(msg);
 
-        if (STR_EQ(msg, "WIFI DISCONNECT")) {
-                if (state.hooks.client_wifi_disconnect_cb)
-                        state.hooks.client_wifi_disconnect_cb();
+        /* Handle and ignore NO-OP URCs */
+        if (STR_EQ(msg, "WIFI CONNECTED") ||
+            STR_EQ(msg, "WIFI GOT IP") ||
+            STR_EQ(msg, "WIFI DISCONNECT")) {
+                if (state.hooks.client_state_changed_cb)
+                        state.hooks.client_state_changed_cb(msg);
                 return true;
         }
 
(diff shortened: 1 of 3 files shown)


src/drivers/esp8266_drv.c
```diff
@@ -462,8 +451,14 @@ static void init_wifi()
...
-        if (!esp8266_register_ipd_cb(rx_data_cb))
-                pr_error(LOG_PFX "Failed to register IPD callback\r\n");
+        const struct esp8266_event_hooks hooks = {
+                .client_state_changed_cb = client_state_changed_cb,
+                .socket_state_changed_cb = socket_state_changed_cb,
+                .data_received_cb = rx_data_cb,
+        };
+
+        if (!esp8266_register_callbacks(&hooks))
+                pr_warning(LOG_PFX "Failed to register callbacks\r\n");
```
brentpicasso: This is a pretty critical error, right? If this fails basically nothing would work.  Thinking of a system reset, but then it could get stuck in an endless reset loop.

src/devices/esp8266.c
```diff
@@ -85,9 +85,12 @@ static bool sparse_urc_cb(char* msg)
 {
         msg = strip_inline(msg);
 
-        if (STR_EQ(msg, "WIFI DISCONNECT")) {
-                if (state.hooks.client_wifi_disconnect_cb)
-                        state.hooks.client_wifi_disconnect_cb();
+        /* Handle and ignore NO-OP URCs */
```
brentpicasso: NO-OP b/c we're just gonna poll for this information anyway?

brentpicasso:
no more comments on review so far. looks pretty solid.

src/drivers/esp8266_drv.c
stieg: If this fails then there may be some undefined behavior, but in all I would not set it at error level.  The system is not necessarily unstable.  But its a warning we should see.

src/devices/esp8266.c
stieg: I should remove that comment.  Originally I was going to ignore them but then I settled on the "state changed" event callbacks to make things consistent and went with that model.

src/drivers/esp8266_drv.c
brentpicasso: I'd definitely call it an error, but not concerned about it. It will show up in the log for sure.

PR #607, opened 2016-05-12
Soft AP Beacon Support

stieg:
This change adds support that allows our soft AP to broadcast its beacon info to both our guests and clients.  Now you can always connect to the unit, regardless of your approach if you can see the beacon.

This change also moves our IP info into its own structure since we needed to do that.


brentpicasso:
:shipit:

Issue #598, opened 2016-05-10
Improve some of the WiFi Client logic


stieg:
Need to fix and separate a couple of bits of information for our WiFi client system.  This includes:

* Remove the snapshot time info from the esp8266 device and move that to driver because that is where that data belongs
* Separate out the IPv4 info from the wifi client structure since that is not part of its configuration
* Use the `struct wifi_client_info` structure to pass in parameters for client settings.  This keeps things consistent and easier on the APIs

stieg:
> Use the struct wifi_client_info structure to pass in parameters for client settings. This keeps things consistent and easier on the APIs

Turns out we can't really do this because the `struct wifi_client_info` does not have a password field nor should it have one.  So this was an invalid action on my part and is thus no longer part of the task.

PR #608, opened 2016-05-12
Move the client info timestamp to its correct location

stieg:
Previously this was mushed into the client info structure.
This was bad form on my part.  This patch fixes that by instead
movint it to within the client structure where it belongs.

Issue #598


brentpicasso:
:shipit:

PR #610, opened 2016-05-13
Make all non-tagged builds DEBUG builds.

stieg:
We tag all RC and Release build for RaceCapture.  To ensure that we
find bugs more quickly we should always build with the watchdog
disabled so that bugs that cause crashes are more noticable and thus
get squashed sooner.  This change does that by enabling the DEBUG flag
on all non-tagged builds.  Tested a tagged build to ensure that it had
no DEBUG flag set.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,13 @@ ifneq ($(OFFICIAL_TAG),)
 endif
 export RELEASE_TYPE
 
+#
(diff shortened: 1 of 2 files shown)


stieg:
Fixes Issue #598

brentpicasso:
looks good. :shipit:

Issue #595, opened 2016-05-10
RCT: ESP8266 boot configuration 


brentpicasso:
Ensure in the ESP8266 device driver that the ESP8266 boot configuration is handled.

port mappings on the F303 processor:

ESP_GPIO0 -> PA3
ESP_RESET -> PA5

PA3 must be driven high for ESP8266 to boot from it's flash. (low if it is to boot from USART)

Issue #583, opened 2016-05-09
Implement WiFi soft & hard reset support


stieg:
Needed for when we have to reset the WiFi device (which will happen).  Reset code also needs to be able to deal with system fallout and reset all proper state internally.

brentpicasso:
partial dupe of #507  (hardware reset aspect)

brentpicasso:
RCT will have control of the reset line for the wifi module. It is wired on PA5.

To reset, pull PA5 to active low. for normal operation, set to active high.

PR #614, opened 2016-05-14
RCT: ESP8266 I/O configuration

brentpicasso:
Enables flash boot configuration and reset control of ESP8266.


brentpicasso:
Resolves #595

include/devices/wifi_device.h
```diff
@@ -0,0 +1,35 @@
...
+#ifndef WIFI_DEVICE_H_
+#define WIFI_DEVICE_H_
+
+#include "cpp_guard.h"
+#include <stdbool.h>
+
+CPP_GUARD_BEGIN
+
+bool wifi_device_init();
+void wifi_device_reset();
```
stieg: Perhaps we should call this `wifi_device_hard_reset` since that is more specific?

src/devices/esp8266.c
```diff
@@ -329,6 +330,9 @@ bool esp8266_init(struct Serial *s, const size_t max_cmd_len,
         if (!_setup(s, max_cmd_len))
                 return false;
 
+        /* Hard reset our ESP8266*/
```
stieg: Probably shouldn't put this here since MK2 would still have issues if this is a re-init scenario.  I would keep the method separate and I will make use of it when I complete #583

stieg:
Minor comments.  If you agree, then fix and :shipit:.  Else lets discuss...

src/devices/esp8266.c
brentpicasso: In MK2, there's no IO configuration, and the hard device reset is a NOOP, so I think this should be OK?

include/devices/wifi_device.h
brentpicasso: well, since it's at the HAL layer it's pretty much the only thing that could happen. Thoughts?

src/devices/esp8266.c
stieg: The device reset on MK2 will be AT command generated.  We will try to issue the command first, then if that fails fall back to the hard method.  So there will be some form of AT reset on MK2

include/devices/wifi_device.h
stieg: Fair.  Since the name has the prefix `wifi_device` I could agree with that.

src/devices/esp8266.c
brentpicasso: gotcha. want to leave it here until #583 is implemented? That way we at least have proper reset control with RCT.

stieg: I suppose it won't hurt.

stieg:
Go for it

PR #615, opened 2016-05-14
account for variable AT+CWAP? response

### LINKED ISSUE #611: handle extended AT+CWAP? AT command response values
Latest AT command set as seen in the ESPWROOM-02 modules used in RCT has extended response values provided when AT+CWAP? is issued.
https://espressif.com/en/file/721/download?token=5D4jCASk

need to handle max conn and ssid hidden responses:
          +CWSAP: ssid, pwd, chl, ecn, max conn, ssid hidden

Older version of AT command set do not include this value. 

max conn - could keep the hard-coded max 4 connections for now
ssid hidden - we can ignore handling this parameter for now

according to the datasheet these additional parameters are optional.

brentpicasso:
Resolves #611

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -663,12 +663,14 @@ static bool get_ap_info_cb(struct at_rsp *rsp, void *up)
          * pwd  -> string
          * chl  -> number
          * ecn  -> number
+         * [max conn] -> number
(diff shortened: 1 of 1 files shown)


stieg:
Jenkins test this please

stieg:
:shipit:  Good patch.

Issue #619, opened 2016-05-14
Enable CHAP support for 3G devices


stieg:
When a user provides a password for their APN, we should enable authentication support on our 3G module since supplying a password implies some form of actual authentication is happening.  Thus when a password is provided, we should use CHAP.  Else if no password is given (field is empty) we should disable auth.  I wish we could use the automatic method, but the SARA series does not support this :(.

stieg:
Merged

PR #620, opened 2016-05-14
Enable CHAP support for APNs that use PPP auth

stieg:
For networks that make use of secure auth (PPP based) we should
support them.  To best handle this we enable CHAP support whenever
a user supplies a password in the APN field.  Supplying a password
implies that there is actually some form of secure authentication
happening, thus we should enable an authentication mechanism on our
APN settings.

For most providers here in the states, there will be no change since
none of them give a password for the APN.  But for some of our friends
overseas, this will be useful.

Note this only applies to our 3G chips.  2G procedures will not change.

Issue #619

### DIFF
--- a/src/devices/sara_u280.c
+++ b/src/devices/sara_u280.c
@@ -182,6 +182,28 @@ static bool sara_u280_put_dns_config(struct serial_buffer *sb,
         return is_rsp_ok(msgs, count);
 }
 
+enum sara_apn_auth {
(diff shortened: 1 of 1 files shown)


brentpicasso:
:shipit:

zkdzegede:
Thanks for the quick support!

stieg:
@zkdzegede 
  For sure.  Let us know if it works when you can.

Issue #597, opened 2016-05-10
Implement ESP8266 Device Mode State Machine


stieg:
So in order to control whether or not our WiFi client or SoftAP is active or inactive, we need to set the device mode.  To do this sanely we should implement this as its own state machine whose priority is above the client and soft AP configs.  This way the mode is always set correctly prior to any attempt to configure the client or the softAP.

stieg:
Done.

PR #629, opened 2016-05-18
Implement WiFi Mode Control

stieg:
The ESP8266 module has various modes that are used to enable/disable
various functionality.  We need to be able to set these modes based
on the user configs so the device acts correctly.

This patch adds support for controlling the mode and responding to user
changes as they come in.

Issue: #597

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -29,7 +29,6 @@
 #include "str_util.h"
 #include "taskUtil.h"
 #include "wifi_device.h"
-
 #include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


src/devices/esp8266.c
```diff
@@ -470,11 +468,13 @@ bool esp8266_join_ap(const char* ssid, const char* pass, void (*cb)(bool))
                 return false;
 
         char cmd[64];
-        if (pass && *pass)
-                snprintf(cmd, ARRAY_LEN(cmd), "AT+CWJAP=\"%s\",\"%s\"",
-                         ssid, pass);
-        else
-                snprintf(cmd, ARRAY_LEN(cmd), "AT+CWJAP=\"%s\",", ssid);
+        const char pfx[] = "AT+CWJAP_DEF";
```
brentpicasso: we can do _CUR or _DEF - I agree with using _DEF so it comes up that way upon power up. :+1:

brentpicasso:
looks good. :shipit:

src/devices/esp8266.c
stieg: Yeah _DEF because its a super quick boot.  Almost too quick :D

PR #631, opened 2016-05-19
Common C Flags

stieg:
This patch solves a long standing frustration of mine by merging the common C Flags into a single location.  This way all projects that are supposed to have common flags actually get them, instead of having to ensure the same flags are set in each project.

This patch also fixes a compilation issue found by GCC 6 (unused variable).

@brentpicasso and @Jeff-Ciesielski I would love your feedback on how to make this better if you have ideas.


main.c
```diff
@@ -115,9 +115,9 @@ int main( void )
         cpu_init();
         pr_info("*** Start! ***\r\n");
 
-#if !defined(_DEBUG)
-        watchdog_init(WATCHDOG_TIMEOUT_MS);
-#endif /* _DEBUG */
+        /* Defined as part of our compilation process */
+        if (true == ASL_WATCHDOG)
+                watchdog_init(WATCHDOG_TIMEOUT_MS);
```
Jeff-Ciesielski: Are you positive that this generates equivalent object code in all cases?

brentpicasso:
Looks ok.  no additional comments from me.

main.c
stieg:
```txt
$ git describe --exact-match
fatal: no tag exactly matches '71c74778500eeba090780e8074541f434f238ec5'
$ make V=1 mk2
...
  CC      libs/STM32F4xx_DSP_StdPeriph_Lib_V1.0.1/Libraries/STM32F4xx_StdPeriph_Driver/src/stm32f4xx_wwdg.o
ccache arm-none-eabi-gcc -c -MD  -ggdb -Os -std=gnu99 -Wall -Werror -Wempty-body -Wclobbered -Wignored-qualifiers -Wmissing-parameter-type -Wold-style-declaration -Wunused-but-set-parameter  -DASL_WATCHDOG=false ...
---
$ git tag -a 'foo' -m 'foo'
$ git describe --exact-match
foo
$ make V=1 mk2
...
  CC      libs/STM32F4xx_DSP_StdPeriph_Lib_V1.0.1/Libraries/STM32F4xx_StdPeriph_Driver/src/stm32f4xx_exti.o
ccache arm-none-eabi-gcc -c -MD  -ggdb -Os -std=gnu99 -Wall -Werror -Wempty-body -Wclobbered -Wignored-qualifiers -Wmissing-parameter-type -Wold-style-declaration -Wunused-but-set-parameter  -DASL_WATCHDOG=true ...
```

Appears to work based on the compiler verbose output :D

PR #633, opened 2016-05-19
Fix RCT Mass Erase Issue, Improve Version Logging

stieg:
This change does 3 things:

* Improves version check so we can better see what is happening when a version check fails
* Reduces firmware image footprint by removing an object that we don't need frequently.  I instead dynamically allocate it, set appropriate values, then free it.
* Remove the `mass_erase` method from our OpenOCD script.  Because with it we will nuke all our configs.  And that makes me a sad panda.


brentpicasso:
Good memory cleanup. :shipit:

PR #636, opened 2016-05-26
Fix Bitwise Logic Bug, Remove Unused Variable

stieg:
Fixes a logic bug in our USART.  Also removes an unused variable that causes GCC 6 to barf.


brentpicasso:
:shipit:

Issue #647, opened 2016-06-03
Fix IRQ levels on SDCard system for MK2


stieg:
Looking at platform/mk2/hal/fat_sd_stm32/fatfs/lo_level_ub/stm32_ub_sdcard.c I see IRQ levels of 1 and 0.  This violates the FreeRTOS settings where no non-FreeRTOS IRQ can be of higher priority than level 5.  But adjusting them down caused the subsystem to break.  This task tracks adjusting them down and isolating why the subsystem is not working.

Issue #541, opened 2016-04-29
Audit all IRQ values


stieg:
Need to audit all of our IRQ values internally to ensure that they are within the scope needed to be sane.  Things to consider:

* Simplifying our IRQ values so that there are only 1 - 3 levels of interrupt depending on need.
* Ensure that no IRQs are higher priority than what is set in FreeRTOSConfig.h.  Doing so will cause potential hard faults in the scheduler.  They are inconsistent too.
* Make a common IRQ code point so that we don't have varying levels of priority between hardware that cause odd behavior between products.

Must be done before 2.10 to ensure stability in products before the next major release.

stieg:
Merged

PR #648, opened 2016-06-03
Audit and sanitize all IRQ Premptive Priority Values

stieg:
Because of how FreeRTOS works there is a certain maximum IRQ value that
non-FreeRTOS interrutps may use.  Using anything higher can compromise
the stability of the system.

This change evaluates all of the IRQ settings across MK2 and RCT and
adjusts the IRQ settings appropriately so that all non-FreeRTOS priority
are lower priority (lower value) than all FreeRTOS IRQ. In other words,
this change ensures that IRQ values >= 5 are set on all non-FreeRTOS
subsystems that generate IRQs.

There is one exception here: The SDCard subsystem.  In tryint to adjust
those IRQ levels the system was no longer responsive.  That work is
tracked by issue #647

Issue: #541

### DIFF
--- a/platform/mk2/hal/i2c_stm32/i2c_device_stm32.c
+++ b/platform/mk2/hal/i2c_stm32/i2c_device_stm32.c
@@ -333,7 +333,7 @@ static void i2c_nvic_setup(struct i2c_priv *p)
 
     /* Enable the I2Cx Interrupt */
     NVIC_InitStructure.NVIC_IRQChannel = p->ev_irqn;
-    NVIC_InitStructure.NVIC_IRQChannelPreemptionPriority = 1;
+    NVIC_InitStructure.NVIC_IRQChannelPreemptionPriority = 5;
     NVIC_InitStructure.NVIC_IRQChannelSubPriority = 0;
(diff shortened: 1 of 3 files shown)


brentpicasso:
looks reasonable. Do we want to fully and collectively bang on this before approving this PR?

brentpicasso:
I'll smoke test in the morning too and then likely we can merge.

Jeff-Ciesielski:
This is more or less equivalent to what we did on RCT yes?

stieg:
> This is more or less equivalent to what we did on RCT yes?

More or less.  Difference is that this affects more devices than on RCT.  SD Card breaking was no good, so I may need to spend some time in there as its violations are egregious (IRQ levels 0 and 1)!

Jeff-Ciesielski:
I'm surprised we're not seeing more scheduler issues if that's the case...

stieg:
> I'm surprised we're not seeing more scheduler issues if that's the case...

So am I frankly.

stieg:
You smoke test this @brentpicasso ?

brentpicasso:
:shipit:

Issue #639, opened 2016-05-26
Support Measuring Lower RPM Rates


stieg:
We need to be able to measure lower RPM rates with accuracy.  As documented by autosportlabs/RaceCapture-Pro_firmware#516 the lowest single pulse RPM rate we support as of 2.9.0 is currently 90 RPM.  I propose we adjust our prescalars so that the following is true:

Speed | Lowest Supported Hz Measurement
-------- | -----------------------------------------------
Fast | 10 Hz
Medium | 1 Hz
Slow | 0.1 Hz

This should give us the breadth of measurement we need to be extremely flexible while only minimally sacrificing precision (which is frankly quite high as is).  This would change the tables to something like this:

Speed | Prescalar | Lowest (Hz) | Lowest (RPM) | Highest (Hz) | Highest (RPM)
-------- | ------------ | --------------- | ----------------- | ---------------- | -------------------------
Fast | 110 |10 | 600 | 655350 | 39321056
Medium | 1101 | 1 | 60 | 65535 | 3932105
Slow | 11010 | 0.1 | 6 | 6553 | 393210

This looks very reasonable since folks can use the fast speed to retain the same high precision RPM measurements they have today.  They can use the medium for slower engines or faster shafts.  And the slow speed is available for anyone measuring something that turns at some very low rate for some unknown reason.

brentpicasso:
looks like a good compromise.

stieg:
I messed up my original prescalar calculations apparently.  I have since updated them.  But the lowest measurable Hz remains unchanged.

stieg:
Merged

PR #651, opened 2016-06-03
Implement new timer prescalar values, fix bugs

stieg:
This change primarily implements new prescalar values that will
allow us to support a wider range of RPM measurements while still
preserving high accuracy.  The default of medium should now work
on all vehicle RPM measurements, including single cylinder vehicles.
The "slow" speed will now allow you to measure values that were
previously unmeasurable.  High speed is still very accurate, but is
now more useful as most standard vehicles can make use of it for
precise RPM measurement, given that their vehicle has 2 or more
sparks per rotation.

This patch also fixes a bug where we could not adjust the prescalar
value due to a switch statement missing breaks (D'oh!).  It also
improves some logging and clears interrupt flags sooner as a best
practice.

Issue: #639

### DIFF
--- a/platform/mk2/hal/timer_stm32/timer_device_stm32.c
+++ b/platform/mk2/hal/timer_stm32/timer_device_stm32.c
@@ -34,12 +34,13 @@
 
 #define INPUT_CAPTURE_FILTER 	0X0
 #define MK2_TIMER_CHANNELS	3
-#define PRESCALER_FAST		84
-#define PRESCALER_MEDIUM	168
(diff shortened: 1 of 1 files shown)


brentpicasso:
Changes look simple. were you able to verify this using a signal generator? If you have that 'scope handy you can program it to output a frequency, which can be fed into the respective timer inputs. :+1:

stieg:
Yes I verified this using a Signal generator I made using Lua Script.  Worked quite well.

Issue #495, opened 2016-03-24
Adjust build process to include RC, Dev tags in official build package names


stieg:
If a build is at an RC or is a dev build, we should adjust the file names accordingly.  This info should be available already within the make system, just need to apply it to the build process.

stieg:
Merged

PR #652, opened 2016-06-03
Improve release package naming, move RCT bin files to top level

stieg:
Improves the release naming of our packages by using the output of
`git describe --dirty` to name the package.  This always gives us
a unique version that we can give to customers _AT ANY TIME_.  This
way we know EXACTLY what version of the product they are using because
the version now has all of this information.  And when it comes to
official and RC releases, the action of tagging will make our version
appear exactly like it used to.

This patch also enables us to start building .zip packages for the
RCT project in anticipation of delivery.  It also improves our
clean target and improves and sorts the contents of .gitignore so
we can quickly find what is in there (and removes duplicate entries).

Issue: #495


brentpicasso:
tested this out and looks good.  However, I noticed the README.md seems out of date with the latest on how to build various firmware artifacts. Consider updating it here, or make a separate PR.

psallandre:
I just came across this and thought I could help: I usually use these options for git describe:
git describe --tags --always --dirty

tags allows to use non annotated tags
always prevent from failing when no ref is found: use abbreviated commit as fallback.

stieg:
Good call @psallandre.  I'll add those options.  I may not use the `--tags` options because I want to ensure we always use annotated/signed tags going forward (and using `--tags` may hide that error).  But the always is a smart idea.

stieg:
Forgot to note that the README has already been updated separately.  Check master and r/2.10.0 for the latest.

Issue #530, opened 2016-04-24
Improve BT baud to 230400


stieg:
See if we can sanely increase our BT unit baud rate to 230400.  Will allow us to spend less time sending BT data.

stieg:
Merging

PR #653, opened 2016-06-04
Up BT baud to 230400, remove unnecessary baud from config

stieg:
This change ups the default BT baud rate to 230400, effectively
doubling the rate at which we can send data over BT.  Combined with
the DMA improvements the result is a snappier BT experience.

This change also removes the baudRate value from the bluetooth
config since that was not being used in any significant way.

Issue #530

### DIFF
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -396,12 +396,10 @@ typedef struct _TrackConfig {
 
 #define BT_DEVICE_NAME_LENGTH 21
 #define BT_PASSCODE_LENGTH 5
-#define DEFAULT_BT_BAUD 115200
 #define DEFAULT_BT_ENABLED 1
 
 typedef struct _BluetoothConfig {
         unsigned char btEnabled;
-        unsigned int baudRate;
         char new_name [BT_DEVICE_NAME_LENGTH];
         char new_pin [BT_PASSCODE_LENGTH];
 } BluetoothConfig;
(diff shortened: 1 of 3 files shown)


stieg:
Tested on MK2.  Seemed to work quite well.

brentpicasso:
looks good. :shipit: looking forward to seeing this work.

brentpicasso:
I'm having a moment of pause about the max data rate of our RS232 transceivers.  The max guaranteed is 120KBPS, but that's under a 3K ohm load.   Typical data rate is listed at 250K.  I think RCP is driving a pretty light load since it's going from the output of one transceiver to the input of another, but might be worth putting it on a scope to double check.

See the datasheet here: http://www.mouser.com/ds/2/146/SP232A_100_051308-57895.pdf

stieg:
Interesting.  Looking at the sheet if typical is 240K, then we are probably ok since the resistance load of the RJ-11 cable

>  think RCP is driving a pretty light load since it's going from the output of one transceiver to the input of another, but might be worth putting it on a scope to double check.

Here is some interesting text I found on the resistance of the phone cabling:

```
A  telephone is usually connected to the telephone  exchange 
by  about three miles (4.83 km) of a twisted pair of No.22  (AWG) 
or  0.5  mm  copper wires, known by your phone  company  as  "the 
loop".   Although  copper  is  a good  conductor,  it  does  have 
resistance.   The resistance of No.22 AWG wire is 16.46 Ohms  per 
thousand  feet  at 77 degrees F (25 degrees C).
```

So even in the worst case scenario of a user running this wire from the front to the back of their vehicle, we should incur less than 20 ohms of total resistance.  Lets keep it in there for now.  Its trivial to move it back to 120K before we officially release it if needed.

PR #656, opened 2016-06-06
Implement WiFi Soft and Hard Reset in Wifi Driver

stieg:
Both hardware and software code necessary to reset the device. Doesn't yet address multiple errors to cause a reset; that is tracked by a later issue. This just gets us support now so that if we fail to init for any reason, we are able to reset the device.


stieg:
Safe to ignore the failure.  This is due to asl_f4_loader being out of date on build machine.

stieg:
Jenkins build this please

stieg:
Jenkins Test This Please

stieg:
Declaring Review Timeout on this one.  Merging.

PR #662, opened 2016-06-07
Synchronous WiFi Connect Support

stieg:
Makes the connect call for the WiFi driver synchronous.  This is because I don't want us to return a Serial device who may or may not have an open connection since Serial has no knowledge of any underlying connection state.  Having the connect method be synchronous makes this easy.  If we want to write an async model later because one is needed then we can with relative ease.


stieg:
Ignore the failure.  This is because I need to update the version of asl_f4_postprocess on our build machine.  I have built it locally and can verify that it works.

stieg:
And crap... this is not quite ready.  SOOOON

PR #664, opened 2016-06-08
Implement socket disconnect, make conn and disconn synchronous

stieg:
This change completes the majority of the wifi socket support by making the connect and disconnect calls synchronous.  It achieves this by using a couple of synchronization primitives and adjusting the callbacks accordingly.  This is inline with POSIX calls, so their behavior is roughly the same.

This change also adds the ability to correct for any issues where we may not be synchronized with the device.  This happens after a soft reset of the device.

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -29,6 +29,7 @@
 #include "panic.h"
 #include "printk.h"
 #include "queue.h"
+#include "semphr.h"
 #include "str_util.h"
 #include "task.h"
 #include "taskUtil.h"
@@ -92,6 +93,8 @@ struct channel {
 struct comm {
         new_conn_func_t *new_conn_cb;
         struct channel channels[MAX_CHANNELS];
+        xSemaphoreHandle connect_semaphore;
+        xSemaphoreHandle connect_cb_semaphore;
 };
 
 /**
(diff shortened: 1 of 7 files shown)


src/drivers/esp8266_drv.c
```diff
@@ -1128,38 +1176,163 @@ bool esp8266_drv_init(struct Serial *s, const int priority,
...
+                break;
+        }
+        if (!connect_result) {
                 pr_warning(LOG_PFX "Failed to issue connect command\r\n");
-                return NULL;
+                goto done;
+        }
+
+        /* Now wait for our callback to come */
+        if (!xSemaphoreTake(cso->cb_semaphore, portMAX_DELAY)) {
```
brentpicasso: do we really want to wait forever here?

src/drivers/esp8266_drv.c
```diff
@@ -1128,38 +1176,163 @@ bool esp8266_drv_init(struct Serial *s, const int priority,
...
-                                   const char* dst_ip,
-                                   const unsigned int dst_port)
+                                   const char* addr,
+                                   const unsigned int port)
 {
-        const int chan_id = get_next_free_channel_num();
-        if (chan_id < 0) {
+        /* This is synchronous code for now */
+        struct channel_sync_op *cso = &esp8266_state.comm.connect_op;
+        xSemaphoreTake(cso->op_semaphore, portMAX_DELAY);
```
brentpicasso: ditto on waiting forever.  We may wedge the driver if something weird happens.

src/drivers/esp8266_drv.c
stieg: Yep.  The AT state machine will never fail to generate a callback (it has a timeout mechanism) so this is sane.

src/drivers/esp8266_drv.c
```diff
@@ -1128,38 +1176,163 @@ bool esp8266_drv_init(struct Serial *s, const int priority,
...
+        if (!ch->connected)
+                pr_warning(LOG_PFX "Channel supposedly not connected.\r\n");
+
+        if (!esp8266_close(cso->chan_id, close_cb)) {
+                pr_warning(LOG_PFX "Failed to issue close command\r\n");
+                goto done;
+        }
+
+        /* Now wait for our callback to come */
+        if (!xSemaphoreTake(cso->cb_semaphore, portMAX_DELAY)) {
```
stieg: Same reasoning as above.

src/drivers/esp8266_drv.c
stieg: This is a mutual exclusion piece, kinda has to be this way.

src/drivers/esp8266_drv.c
brentpicasso: Got it.

brentpicasso:
Ok, no other comments. :shipit:

PR #666, opened 2016-06-08
Rude AT Command Support

stieg:
This change is built on top of PR #664, so please review that first prior to reviewing this.

This change adds the ability to support "Rude AT commands".  Rude at commands are AT commands that show up during the middle of another command instead of being queued.  They are considered rude because they interrupt the other command (and can even ruin it).  But sometimes you have to work with the cards you're dealt and this is one of those times.

In brief testing it seems to help the WiFi driver cope with a flood of messages, but there is still the chance that if you flood it with messages it simply will not be able to handle them or may even freeze.  I'm doing my best to prevent that as is reasonable.

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -99,6 +99,7 @@ struct channel_sync_op {
 
 struct comm {
         new_conn_func_t *new_conn_cb;
+        struct channel* tx_chan;
         struct channel channels[MAX_CHANNELS];
         struct channel_sync_op connect_op;
         struct channel_sync_op close_op;
@@ -328,6 +329,17 @@ static void rx_data_cb(int chan_id, size_t len, const char* data)
 }
 
 /**
+ * Drops all the data in a queue because there is no connection to send
+ * the data over.
+ * @param ch The channel containing the data to drop.
+ */
+static void drop_data(struct channel *ch)
+{
+        serial_purge_tx_queue(ch->serial);
+        ch->tx_chars_buffered = 0;
+}
+
+/**
  * Callback that is invoked when the send_data method completes.
  */
 static void _send_data_cb(int bytes_sent)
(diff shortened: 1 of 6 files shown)


src/drivers/esp8266_drv.c
```diff
@@ -1128,38 +1206,163 @@ bool esp8266_drv_init(struct Serial *s, const int priority,
...
-                                   const char* dst_ip,
-                                   const unsigned int dst_port)
+                                   const char* addr,
+                                   const unsigned int port)
 {
-        const int chan_id = get_next_free_channel_num();
-        if (chan_id < 0) {
+        /* This is synchronous code for now */
+        struct channel_sync_op *cso = &esp8266_state.comm.connect_op;
+        xSemaphoreTake(cso->op_semaphore, portMAX_DELAY);
```
brentpicasso: what if we can never get the semaphore? We should bail out or fail somehow. Otherwise the thing could get wedged.

brentpicasso: I realized the PR notes and this builds on the previous one I reviewed. holding off comments.

stieg:
Ok @brentpicasso this is ready for your eyes :)

src/modem/at.c
```diff
@@ -253,7 +253,20 @@ static void at_task_run_bytes_read(struct at_info *ati, char *msg)
          * Our rx state will dictate how we process this message beacuse that
          * allows us to know previous messages and what message type to expect.
          */
-        switch (ati->rx_state) {
+        //enum at_rx_state state = ati->rx_state;
```
brentpicasso: should we remove commented code?

brentpicasso:
looks great, aside from one question, optionally address. otherwise :ship: !

src/modem/at.c
stieg: Can do.

stieg: And hell you just saved my butt.  That was some code that I left in there for testing that should have been removed.  Thanks :)

stieg:
Ok... turns out I got ahead of myself with this one.  Need to do more.  So closing this until its really ready :\

Issue #668, opened 2016-06-09
Reduce DMA Buffer Size for GPS USART Rx to prevent delay in reception of data


stieg:
Our previous setting was 132 bytes.  Based on that it would take 66 bytes to trigger any type of transfer to the rx queue.  If there is a message that isn't 66 bytes long, then it wouldn't show up until 66 bytes had been sent.  That is not good.  This change helps that out by reducing the DMA receive size down to 32 bytes.  This means we only need to receive 16 bytes before a transfer happens.  This is much more reasonable.

All this gets fixed in 2.10.0 with the timers to flush the DMA buffers, but until then this is a simple change that will help reduce the chances of us not responding to incoming messages.

stieg:
Merged

PR #669, opened 2016-06-09
Reduce the GPS DMA Rx buffer to 32 bytes from 132

stieg:
This means that it will only take 16 characters to trigger a transfer
of data from the buffer instead of 66.  This is a solid improvement since
it will help us deal with low traffic situations (like when we detect a
timeout).

Issue #668


brentpicasso:
if GPS appears stable then :shipit: !

Issue #670, opened 2016-06-09
WiFi Client Fails to Join on MK2 Unit


stieg:
Seeing this message:

```
[serial (Aux/Wifi)] RX: OK(\r)(\n)[ESP8266 Driver] Successfully joined network
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Retrieving Wifi Client Info

[serial (Aux/Wifi)] TX: AT+CWJAP_DEF?(\r)(\n)
[serial (Aux/Wifi)] RX: +CWJAP_DEF:"LAN Downunder","ff:ff:ff:ff:ff:ff"(\r)(\n)
[serial (Aux/Wifi)] RX: (\r)(\n)
[serial (Aux/Wifi)] RX: OK(\r)(\n)[esp8266] get_ap_cb failed with msg: Failed to parse AP info
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Setting Wifi Client Info
[ESP8266 Driver] Joining network: LAN Downunder

[serial (Aux/Wifi)] TX: AT+CWJAP_DEF="LAN Downunder","<REDACTED>"(\r)(\n)
[serial (Aux/Wifi)] RX: WIFI DISCONNECT(\r)(\n)
[serial (Aux/Wifi)] RX: WIFI CONNECTED(\r)(\n)
[serial (Aux/Wifi)] RX: WIFI GOT IP(\r)(\n)
[serial (Aux/Wifi)] RX: (\r)(\n)
[serial (Aux/Wifi)] RX: OK(\r)(\n)[ESP8266 Driver] Successfully joined network
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Retrieving Wifi Client Info

[serial (Aux/Wifi)] TX: AT+CWJAP_DEF?(\r)(\n)
[serial (Aux/Wifi)] RX: +CWJAP_DEF:"LAN Downunder","ff:ff:ff:ff:ff:ff"(\r)(\n)
[serial (Aux/Wifi)] RX: (\r)(\n)
[serial (Aux/Wifi)] RX: OK(\r)(\n)[esp8266] get_ap_cb failed with msg: Failed to parse AP info
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Setting Wifi Client Info
[ESP8266 Driver] Joining network: LAN Downunder
```

Due to an incomplete handling of WiFi data on older firmwares.

PR #671, opened 2016-06-09
Fix Wifi Client Join

stieg:
We were failing to join clients on older WiFi hardware because the
older baseband software doesn't return as many arguments as newer
does.  This patch works around that issue so we can support both.

Issue #670


ryandoherty:
LGTM!

PR #672, opened 2016-06-10
Fix Wifi Deadlock

stieg:
This patch eliminates the WiFi deadlock that can happen when you flood the device with more data than it can process.  It achieves this by reducing the xQueueSend timeout to one tick (previously at the maximum value).  This gives the system a chance to context switch back and receive the data, but if it can't then the data will be dropped at the expense of one tick.

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -47,6 +47,7 @@
 #define INIT_FAIL_SLEEP_MS	10000
 #define LOG_PFX			"[ESP8266 Driver] "
 #define MAX_CHANNELS		5
+#define RX_DATA_TIMEOUT_TICKS	1
 #define SERIAL_BAUD		115200
 #define SERIAL_BITS		8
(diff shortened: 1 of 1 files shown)


Jeff-Ciesielski:
Is a single tick enough?  I'm not familiar enough with the codepath to speculate whether there could be situations where you might not process it in time.

stieg:
> Is a single tick enough? I'm not familiar enough with the codepath to speculate whether there could be situations where you might not process it in time.

There definitely are situations where we can't process this in time.  Think of this incoming data callback as a soft IRQ.  If I don't service this quickly, it will stall the rest of the WiFi pipeline, hence why it needs to be quick.  If I set it to 0 then I give the system absolutely no chance to try and receive more data than what the serial buffer can take.  Hence why I chose 1.  Small enough to really not notice the stall, but enough for the system to be able to catch up if it can.

Jeff-Ciesielski:
Fair enough.  Change looks sane to me.  🚢

Issue #515, opened 2016-04-12
Add support for controlling streaming via our API calls


stieg:
Currently if you want to use our API to get samples, the caller must poll.  This is crap because it adds unnecessary CPU load to our system and has to fake a lot of information (such as sample tick) since that is not relevant.  This ticket tracks implementing an API endpoint that will allow us to enable sample streaming.  Ideally the caller can specify the maximum sample rate they wish to receive, and the provider will provide up to that rate if configured so.  We would then deprecate the polling API and remove it in the next major API version.

stieg:
Adding the WiFi tag here since with polling you can safely get about 10Hz.  However we would like to get more and to do so we would need this feature.

stieg:
Ok.  Come up with a plan on how this will work:

* Implement a form of ioctl handler on Serial devices.  This will allow us to send arbitrary commands to the subsystem that manages the particular Serial device
* Implement the actual JSON commands to enable/disable streaming.  These will call the ioctl command in Serial with the defined actions and arguments
* Add a command to the logger task where other tasks may register for callbacks when logging samples are ready.  They will provide a sample rate and an endpoint to call.
* Re-tool the WiFi Task to be 100% event driven (it mostly is now, just some slight hackiness will change)
* Add an event handler for the WiFi Serial devices that handles all serial ioctl invocations
* Tie the sample logic into the ioctl commands.

That should do it and still enable us to later migrate over towards using sockets to replace Serial devices (since sockets implement ioctl).

stieg:
Merged

PR #682, opened 2016-06-15
Telemetry Stream API Support

stieg:
This PR adds telemetry streaming support to both USB and WiFi endpoints.  It is controlled by the following API commands:

* Start the telemetry stream
 * rate: The maximum sample rate to supply.
```json
{"setTelemetryStart":{"rate":10}}
```
* Stop the telemetry stream: 
```json
{"setTelemetryStop":{}}
```

I implemented this by creating an ioctl method on our Serial device that gets handled by the managing tasks.  This is ideal since it is extensible and will allow us to add custom features and support per each Serial backing device going forward as well as play nice with Sockets when we do move over to them.

Samples need always be sent by the managing task of the Serial device to ensure that no two streams interfere with one another and to ensure that the loggingTask is never blocked on any one Serial device.

My apologies that this PR is a bit of a beast.  Worth it.

Issue #515


src/usb_comm/usb_comm.c
```diff
@@ -19,38 +19,236 @@
...
+
+/**
+ * Event struct used for all events that come into our wifi task
+ */
+struct usb_event {
+        enum task {
+                TASK_RX_DATA,
+                TASK_SAMPLE,
+        } task;
+        union {
```
Jeff-Ciesielski: Any reason this is a union? Future expansion planned?

stieg: Yeah.  Leaving the door open to future expansion as needed without the need to go back and change it all.

stieg:
Good enough.  Shipping.

Issue #621, opened 2016-05-15
Toggle LED upon WiFi Activity


brentpicasso:
Need to toggle green LED upon WiFi activity.  TCP message, AT command, or whatever.

stieg:
Merged

PR #686, opened 2016-06-17
RCT WiFi LED support

stieg:
Adds support so that our LED on the RCT unit will blink whenever Tx or Rx traffic is detected.  This is only for socket traffic, not AT traffic.

Issue #621

### DIFF
--- a/include/LED/led.h
+++ b/include/LED/led.h
@@ -43,6 +43,7 @@ enum led {
         LED_LOGGER    =  1,
         LED_GPS       =  2,
         LED_TELEMETRY =  3,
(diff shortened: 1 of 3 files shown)


Jeff-Ciesielski:
Looks sane to me.  I think a very thin case could be made to not have a constant timer and to instead set a single-shot timer to disable the LED, but that's a premature optimization that doesn't really buy a ton.
:shipit:

stieg:
I like that idea.  I can implement it quick before I merge.

stieg:
Apply feedback from @Jeff-Ciesielski , make blinking based on AT commands.

Issue #425, opened 2016-02-11
Ensure new RaceCapture IMU conforms to SAE J670E standard


stieg:
As was addressed in #360, we need to ensure that the new RC unit is also setup appropriately.  I would have done it as part of the last ticket except I am still waiting for the hardware to show up.

stieg:
Merged

PR #687, opened 2016-06-19
Map the RCT IMU to SAE J670E spec

stieg:
We standardized Podium on this spec and so our units should default
map to this setting based on default orientation.

Issue #425

### DIFF
--- a/platform/rct/hal/imu_stm32/imu_device_stm32.c
+++ b/platform/rct/hal/imu_stm32/imu_device_stm32.c
@@ -88,22 +88,22 @@ int imu_device_read(enum imu_channel channel)
 {
         /*
          * Mapped to SAE J670E standard. Front of unit is RJ-45 and USB
-         * plug facing rear of vehicle when mounted on windshield.
+         * plug facing front of vehicle when mounted on windshield.
          * Issue #425
          */
         switch(channel) {
         case IMU_CHANNEL_X:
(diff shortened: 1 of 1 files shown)


platform/rct/hal/imu_stm32/imu_device_stm32.c
```diff
@@ -86,33 +86,27 @@ void imu_device_init()
...
-    default:
-        break;
-
-    }
-
-    return ret;
+        /*
+         * Mapped to SAE J670E standard. Front of unit is RJ-45 and USB
+         * plug facing rear of vehicle when mounted on windshield.
+         * Issue #425
```
brentpicasso: Actually, to match the IMU's natural mounting orientation of MK2 it should match MK2.

MK2:  mounted flat with green terminal block facing forward
RCT: mounted flat with USB / RJ45 facing forward

stieg: So this is backwards from when we last spoke about it (which was some time ago).  I don't mind changing it, but then we need to be sure that the cover for the unit with our logo on top reflects this since this is what customers will use for orientation.  IIRC it does not reflect this now (at least with the units we were demoing at Lucky Dog).  Will you confirm this will be fixed before we ship?

brentpicasso:
:shipit:

Issue #178, opened 2015-10-26
Change distance to be calculated by integrating speed over time.


ryandoherty:
Customer feedback: http://www.autosportlabs.org/viewtopic.php?p=22581#22581

stieg:
Will need to test if this is more accurate than what we currently have.

ryandoherty:
I think a user commented in the forums or via email a long time ago with data that showed it was (used RCP log to derive distance based on speed). We still need to test, but seems promising.

JJMcDonough:
I'm not sure what method can be used to verify accuracy, but spot checking one of my files I do see about a 3% difference in distance over 1 mile comparing the raw distance channel (GPS set to 10hz) to my speed integration distance calc. Like I said on the message board, when I was running too high of a GPS sampling rate, the raw distance was way off because the position data jumped around a bunch. Happy to provide file/feedback/testing.

brentpicasso:
Hi, great feedback.  Do you have a sense which one is more accurate? IOW, did you also compare this against a known distance?

JJMcDonough:
Did not compare against a known distance. I have DL1 data on the same course, but it's a bit tricky to identify the finish line from the data.

stieg:
merged with commit 18b423aea03d98f69c4a54f4d0aa341fe51688c7

PR #688, opened 2016-06-19
Move our distance calculation to one that is speed/time based

stieg:
***1 Upvote*** As we have seen in the field measuring distance by taking the
pythag distance between our current point and the last is not good
in many situations due to GPS inaccuracies.  Speed over time may
be a better approach since we have noticed that speed calculations
tend to be very accurate relative to the sometimes erratic location
calculations.  Hence this change.

Issue #178

### DIFF
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -142,7 +142,7 @@ static void update_distance(const GpsSnapshot *gps_ss)
          * Delta ms: ms
          * KM/H * delta ms / 3600 = delta distance.
          */
-        g_distance += speed_avg * gps_ss->delta_last_sample / 3600;
+        g_distance += speed_avg * gps_ss->delta_last_sample / 3600000.0;
 }
 
 static void set_distance(const float distance)
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -19,7 +19,6 @@
  * this code. If not, see <http://www.gnu.org/licenses/>.
  */
 
-
 #include "auto_track.h"
 #include "dateTime.h"
 #include "geoCircle.h"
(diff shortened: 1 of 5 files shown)


src/lap_stats/lap_stats.c
```diff
@@ -129,15 +129,22 @@ static void end_lap_timing(const GpsSnapshot *gpsSnapshot)
...
+static void update_distance(const GpsSnapshot *gps_ss)
 {
-    const GeoPoint prev = gpsSnapshot->previousPoint;
-    const GeoPoint curr = gpsSnapshot->sample.point;
+        const float speed_avg =
+                (gps_ss->sample.speed + gps_ss->previous_speed) / 2;
 
-    if (!isValidPoint(&prev) || !isValidPoint(&curr))
-        return;
+        if (speed_avg < 1)
```
brentpicasso: so if speed_avg is 0.99 we discount it? did you really mean <= 0?

brentpicasso:
Looks straightforward, only one comment.  We should take this branch for a spin in real world! :)

stieg:
Agreed.  This is why I have made it part of an RC blocker.  My intention is to release an alpha to some customers on our beta list to see if they want to give it a go.  Thoughts?

brentpicasso:
That wouldn't hurt shortly after we smoke test it internally!

Jeff-Ciesielski:
This looks like a really good change to me.  I think the only thing better would be to use speed/time/distance measurements as inputs to a kalman filter (since, you know,  that's exactly what they were designed for...except in space :D ).  Very curious to see the results.

stieg:
@Jeff-Ciesielski Yeah I imagine we will need to do that at some point anyways.  As we have seen we can't count on this GPS unit to be consistent so this is at least a step in the correct direction.  But yeah... it would be good to play with Kalman filters at some point.

src/lap_stats/lap_stats.c
stieg: I wanted a low baseline here so people moving the car don't trip it up slightly.  Anything under 2mph is slower than walking.  I could be convinced into changing this to 0.5.  0 is out of the question IMHO since it would be trivial to bump the unit and get a non-zero value.  And if we ever see a speed less than 0 then I want words with our GPS chip maker.

brentpicasso: Understood. I would recommend documenting this, and removing the magic number so the intention is obvious.  Thanks!

brentpicasso:
Awesome. Let's road test this and once it looks good, ship it.

brentpicasso:
:+1:   I should've caught that off by 1000 issue :)

stieg:
I'll merge this once testing passes.  Currently GPS is giving me fits.

stieg:
Closing this for now.  Will merge independently.

brentpicasso:
I think we should continue testing this and merge if it looks good. willing to do so tomorrow on the road with Mk2.

stieg:
Are you saying you just want to merge it and test or keep it isolated until confirmed operational?

brentpicasso:
Test first then merge is what I had in mind.

Brent Picasso
Autosport Labs

Issue #612, opened 2016-05-14
Switch from deprecated AT commands to _CUR variants


brentpicasso:
need to switch to the _CUR variants of configuration AT commands for ESP8266.
https://espressif.com/en/file/721/download?token=5D4jCASk
e.g.

AT+CWMODE
AT+CWJAP
AT+CWSAP
AT+CWDHCP

stieg:
Merged

stieg:
Merged

PR #690, opened 2016-06-20
Migrate away from all ESP8266 deprecated AT Commands

stieg:
This patch migrates us away from using any/all deprecated AT commands
on the chip.  Instead I opted to use _DEF extensions on the commands
that were deprecated since it would give us the quickest startup time
(sub 1 second) for the wifi device if we went down this road.

Issue #612

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -378,8 +378,8 @@ bool esp8266_set_op_mode(const enum esp8266_op_mode mode,
         if (!check_initialized("set_op_mode"))
                 return false;
 
-        char cmd_str[16];
-        snprintf(cmd_str, ARRAY_LEN(cmd_str), "AT+CWMODE=%d", (int) mode);
+        char cmd_str[24];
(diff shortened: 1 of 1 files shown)


Jeff-Ciesielski:
I can't exactly tell from their (not exactly crystal clear) documentation, but I *think* you might want to use the _CUR suffix for 'gets'.  

I could be misunderstanding, but I would assume based on the set behavior, that _DEF would give you the default value stored in flash, not necessarily the current operating parameters.  While I realize that the default stored in flash SHOULD be the current operating mode in our case, it seems a safer bet to try getting the current state of affairs rather than the default setting.

stieg:
The way I read their docs the only difference between _CUR and _DEF commands is whether or not things are persisted (making the getters of _CUR and _DEF supposedly the same) but yours is not a bad idea.  I will update it.

brentpicasso:
I think it would be fine with either _CUR or _DEF on the get calls.   other than that, :shipit:

PR #691, opened 2016-06-21
Adds handling of corner case for DMA, and serial overflow check

stieg:
This useful code adds a check for DMA in our rx buffer code for a case that could happen but never should (DMA counter == 0).  It also adds a check that only runs on non release builds that will inform us if we are overflowing any of our buffers.  This helped catch an overflow issue on RCT.  Tha is fixed as part of this patch.


stieg:
Review Timeout.  Merging

PR #701, opened 2016-06-24
Consistent Builds for MK2 and RCT

stieg:
Does all the work necessary to get us consistent binaries on MK2 and RCT.  Unit test directory does not matter IMHO.  Here is what changed:

* Got rid of -g and -ggdb since these introduce entropy based on the system building the binary.  You can re-enable them by specifying ASL_DEV_CFLAGS='-ggdb' in your environment.
* Fixed ar library building to be deterministic by adding the necessary flags and ensuring the order of arguments into ar is consistent.
* Ported all changes to RCT
* Migrated RCT linker through C instead of C++
* Sorted .o files for C linker as order matters.
* Removed duplication of .elf build for RCT binary
* Added `-static` to both builds
* Added in support for FULL_ASSERT, then disabled it since MK2 has bugs that need fixing.
* Removed mod_string since it was causing problems and we should be using newlib anyways.


make/asl_cflags.mk
```diff
@@ -19,17 +19,26 @@
...
+# Basic flags
+ASL_CFLAGS += -Os -std=gnu99
 
+#
+# Only specify ASL_DEBUG on devel builds. Reason being that when
+# we release RC builds to the field, we want them to be as close
+# to the official build as possible without being labeled as official.
+# That means no checks that would occur when DEBUG is active.
+#
+# STIEG: BUG: -DUSE_FULL_ASSERT - Not working on MK2.
```
Jeff-Ciesielski: Missing symbol, or just blowing up?

platform/mk2/Makefile
```diff
@@ -67,18 +67,19 @@ LUA_LIB_DIR=$(APP_PATH)/lib/lua/src
 
 INCLUDES += $(CPU_INCLUDES) $(BOARD_INCLUDES) $(LIB_INCLUDES) $(APP_INCLUDES) $(LUA_INCLUDE)
 CFLAGS += $(INCLUDES)
-ASFLAGS += -mcpu=$(CPU_TYPE) $(FPU) -g -Wa,--warn
+# STIEG: Investigate size opt flags on assembler.
```
Jeff-Ciesielski: IIRC, they matter with LTO

platform/mk2/hal/ADC_stm32/ADC_device_stm32.c
```diff
@@ -21,15 +21,15 @@
...
+#include <string.h>
 #include "stm32f4xx.h"
 #include "stm32f4xx_adc.h"
 #include "stm32f4xx_dma.h"
 #include "stm32f4xx_gpio.h"
 #include "stm32f4xx_rcc.h"
 #include <stdbool.h>
 #include <stdint.h>
-
+#include <string.h>
```
Jeff-Ciesielski: OMG DOUBLE INCLUDE!!! WHAT DOES IT MEAN!?!?!?

platform/rct/Makefile
```diff
@@ -252,33 +251,37 @@ inc-y += $(RC_PLATFORM_DIR)/mk2/hal/fat_sd_stm32/fatfs
 ################################################################################
 obj-y := $(addprefix $(OUTDIR)/,$(patsubst %.s,%.o,$(patsubst %.c,%.o,$(src-y))))
 
+#
+# Our chip is actually a cortex-m4 with FPU.  We have not included the correct
+# FreeRTOS libs as well as compiler args to make it work properly with this arch
+# so for now this stays as an M3.
```
Jeff-Ciesielski: O.o

Jeff-Ciesielski:
Minor nits, but overall I dig it.  Moving to the same libraries will (hopefully) keep weird edge cases to a minimum and should make migrating to a higher end platform easier in the future.  

That CM3 part is disturbing, but not surprising.  The whole point behind the ARM CM-x series is being able to seamlessly move up to higher end cores, and apparently this is a delivery on that promise.  Probably wouldn't hurt performance to get that fixed though.

:shipit:

platform/rct/Makefile
stieg: Yeah... I know.  I was a bit shocked to learn this.  The F1 and F2 series are cortex M3, F3 and F4 are cortex M4 chips.

platform/mk2/Makefile
stieg: Ahh... that may explain some of the size jumps I saw.  Will dig into this.

make/asl_cflags.mk
stieg: Blowing up.  We are using some improper values during init and the asserts are catching them since we never used them before now.  But I really wanted to get this out the door so i did this.

Jeff-Ciesielski: Hehe, gotcha.  So in reality, it actually *is* working, but we have bugs...

platform/mk2/hal/ADC_stm32/ADC_device_stm32.c
stieg: say what?  So confused

Jeff-Ciesielski: string.h was already included on line 24.

Issue #683, opened 2016-06-15
Add "setLiveTrack" API


stieg:
This feature will set the live volatile track in our lap stats.  This is necessary to that the app can implement track updating without adjusting the user's config.

stieg:
Merged

Issue #681, opened 2016-06-15
lapstats_config_changed Does Not Reset Internal Fields of LapStats


stieg:
Seems that if we call this method we don't reset internal pieces.  This is probably logical since doing so could cause a segfault as part of a race condition.  To avoid this we need to reset everything right before we configure everything to ensure that we start fresh every time we configure our lap_stats.  So simple fix is to drop a call to `lapstats_init` in `lapstats_setup_internals`.  Should be sane.

stieg:
Merged

PR #705, opened 2016-06-27
SetActiveTrack Support, Valid Track info in status API and a Minor Bug Fix

stieg:
Adds support for `setActiveTrack` API endpoint.  This allows the caller to set this track without saving it to config.  Calls look somehing like this:
```json
{"setActiveTrack":"track"{"id":999999,"type":0,"sf":[1.0,1.0]}}
```
Optional "rad" or "radius" keys are available too for setting radius of geo circles of the track.

This changeset also adds a "valid" status to the track status api to let the reader know if RCP has a valid track (ie one with at least a valid start/finish) defined.

This also fixes a minor bug related to resetting lapstats data during a change in configuration.  

Issues #683 #681


include/lap_stats/lap_stats.h
```diff
@@ -26,16 +26,18 @@
...
 CPP_GUARD_BEGIN
 
 typedef enum {
-    TRACK_STATUS_WAITING_TO_CONFIG = 0,
-    TRACK_STATUS_FIXED_CONFIG,
-    TRACK_STATUS_AUTO_DETECTED,
+        TRACK_STATUS_WAITING_TO_CONFIG = 0,
+        TRACK_STATUS_FIXED_CONFIG,
+        TRACK_STATUS_AUTO_DETECTED,
+        TRACK_STATUS_EXTERNALLY_SET,
```
brentpicasso: good naming here :)

src/lap_stats/lap_stats.c
```diff
@@ -82,15 +82,210 @@ static float g_distance;
...
+ */
+float lapstats_degrees_to_meters(const float degrees)
 {
-    // There are 110574.27 meters per degree of latitude at the equator.
-    return degrees * 110574.27;
+        /*
+         * LEGACY HACK. Our original Track object had the radius of
+         * geoCircle measured in degrees!!! This method exists to convert
+         * from that measurement to one that is measured in meters, a more
+         * civilized unit of measure. I would love to murder the need to use
```
brentpicasso: murder?

test/json_api_files/set_active_track_radius_meters.json
```diff
@@ -0,0 +1 @@
+{"setActiveTrack":{"radius":7,"track":{"id":8888,"type":0,"sf":[1.0,1.0]}}}
```
brentpicasso: suggest a test case to verify setting an active track that also includes sector points.

src/lap_stats/lap_stats.c
```diff
@@ -82,15 +82,210 @@ static float g_distance;
...
+/**
+ * Resets all the track information. This will cause us to become
+ * unconfigured which may prompt an automatic reconfiguration.
+ */
+static void reset_track()
+{
+        /*
+         * Reset items related to the track so that if the track supplied
+         * is invalid we can still do basic stuff like keep track of
+         * distance.
```
brentpicasso: :+1:  I like this safety net.

src/lap_stats/lap_stats.c
```diff
@@ -82,15 +82,210 @@ static float g_distance;
...
+
+/**
+ * Sets the active track that lap_stats uses for determining its various
+ * statistics.
+ * @param track Pointer to the track where we are at.
+ * @param radius Radius of the geo circles at this track in meters.
+ * @param track_status Status indicating who set the track value.
+ * @return true if setup was successful, false otherwise.
+ */
+static bool _set_active_track(const Track *track, const float radius,
```
brentpicasso: I see very few cases of function names starting with _ in our code, so it stood out for me.

the static keyword indicates the function is private to this module. is this to visually identify and further reinforce that fact?

src/lap_stats/lap_stats.c
```diff
@@ -82,15 +82,210 @@ static float g_distance;
...
+}
+
+/**
+ * Informs the caller as to whether or not the currenctly configured
+ * track (if one is configured at all) is valid. By valid we mean that
+ * the track is capable of recording laps (valid start/finish points
+ * are defined).
+ * @return true if valid start/finish points exist, false otherwise.
+ */
+bool lapstats_is_track_valid()
```
brentpicasso: since we also have a settable radius, should we check for this as well?

src/lap_stats/lap_stats.c
```diff
@@ -47,7 +47,7 @@
 
 #define TIME_NULL -1
 
-static const Track *g_active_track;
+static Track g_active_track;
```
brentpicasso: I like this too, always having an active track as a static memory block

src/lap_stats/lap_stats.c
```diff
@@ -82,15 +82,210 @@ static float g_distance;
...
+/**
+ * Tells us if we have sectors defined for this track.
+ * @return true if we do, false otherwise.
+ */
+static bool isSectorTrackingEnabled(const Track *track)
+{
+        if (!isStartFinishEnabled(track))
+                return 0;
+
+        /*
```
brentpicasso: comment formatting.  s/b 
/**

src/lap_stats/lap_stats.c
brentpicasso: also comment formatting /**

src/lap_stats/lap_stats.c
```diff
@@ -82,15 +82,210 @@ static float g_distance;
...
+static void setup_geo_triggers(const Track *track, const float radius)
+{
+        GeoPoint gp;
+        struct GeoCircle gc;
+
+        gp = getStartPoint(track);
+        gc = gc_createGeoCircle(gp, radius);
+        g_start_geo_trigger = createGeoTrigger(&gc);
+
+        /*
```
brentpicasso: applies throughout

src/lap_stats/lap_stats.c
stieg: Negative.  Not meant for documentation like the method comments are.

src/logger/loggerApi.c
```diff
@@ -1837,3 +1838,35 @@ int api_set_telemetry_stop(struct Serial *serial, const jsmntok_t *json)
...
+        const jsmntok_t *json_track = jsmn_find_node(json, "track");
+        /* If no track is given, then fail */
+        if (json_track == NULL)
+                return API_ERROR_PARAMETER;
+
+        memset(&track, 0, sizeof(track));
+        setTrack(json_track + 1, &track);
+
+        const bool rad_set = setFloatValueIfExists(json, "rad", &tmp);
+        const bool radius_set = setFloatValueIfExists(json, "radius", &radius_m);
```
brentpicasso: are we first introducing the concept of radius in meters with this API call?

src/lap_stats/lap_stats.c
brentpicasso: ok.

On Mon, Jun 27, 2016 at 8:14 PM, Andrew Stiegmann <notifications@github.com>
wrote:

> In src/lap_stats/lap_stats.c
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/705#discussion_r68692018>
> :
>
> > + */
> > +static bool isStartFinishEnabled(const Track *track)
> > +{
> > +        return isFinishPointValid(track) && isStartPointValid(track);
> > +}
> > +
> > +/**
> > + * Tells us if we have sectors defined for this track.
> > + * @return true if we do, false otherwise.
> > + */
> > +static bool isSectorTrackingEnabled(const Track *track)
> > +{
> > +        if (!isStartFinishEnabled(track))
> > +                return 0;
> > +
> > +        /*
>
> Negative. Not meant for documentation like the method comments are.
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly, view it on GitHub
> <https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/705/files/61367b12e274e8856b5582ce0b17db5337599935#r68692018>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ADHbzVYjK53rGafFRBEHxIxRP9E08H8uks5qQJGBgaJpZM4I-v7P>
> .
>



-- 
Brent Picasso

src/lap_stats/lap_stats.c
stieg: I tend to use the _<method name> prefix to indicate that the caller should really understand what they are doing when they invoke this method.  Its a practice I learned when doing Linux work and is one that I am still refining admittedly.  In this case its because both `lapstats_set_active_track` and `lapstats_setup` both invoke this method as it is common code to them and to no-one else.  I suppose I could drop the leading `_` and just make it set_active_track, but then that might get confusing with `lapstats_set_active_track`

src/lap_stats/lap_stats.c
stieg: Could substitute with `kill`, `ax`, `eradicate`, `exterminate`, `wipe out`,  :trollface:

src/lap_stats/lap_stats.c
stieg: I like your style.   Will make it so.

test/json_api_files/set_active_track_radius_meters.json
stieg: Sure.

src/logger/loggerApi.c
stieg: Yes, but with full backwards compatibility.  In traditional track setting method the "rad" value holds the value of radius in degrees of latitude.  This is still true here too.  The "radius" variable is new and takes the radius value in meters.  Its value trumps that of "rad" if set.

Internally we have been converting that value to meters since 2.8.0 days (whenever I did my first major refactor on lap stats).  The only reason it has stayed around is because the API supports it that way (but I imagine you never set/use it in the App).

src/lap_stats/lap_stats.c
stieg: I removed the leading `_`

src/lap_stats/lap_stats.c
stieg: We already do check for it when the `set_active_track` method is invoked.  I upgraded it so that it enforces a minimum radius of 1 meter.

Issue #520, opened 2016-04-17
RCT: Battery voltage channel only samples once


brentpicasso:
It seems the battery voltage in RCT only samples once and never updates. ADC channel for battery needs to be updated as needed.

PR #706, opened 2016-06-28
Fix and Improve ADC setup on RCT

stieg:
ADC wasn't firing more than once after init. We had properly setup
the init structure for the device but had missed setting the
ADCDmaConfig to circular (defaults to one shot).  This is what was
causing the failure.

In the process of doing this I read most of the reference manual on
this device and implemented some other bits we will benefit from
(like calibration).  I also cleaned out a lot of unnecessary fluff
since most of it was default settings.

Issue #520

### DIFF
--- a/platform/rct/hal/ADC_stm32/ADC_device_stm32.c
+++ b/platform/rct/hal/ADC_stm32/ADC_device_stm32.c
@@ -95,7 +95,6 @@ int ADC_device_init(void)
         ADC_SelectCalibrationMode(ADC3, ADC_CalibrationMode_Single);
         ADC_StartCalibration(ADC3);
         while(SET == ADC_GetCalibrationStatus(ADC3));
-        /* calibration_value = ADC_GetCalibrationValue(ADC3); */
 
         ADC_InitTypeDef ADC_InitStructure;
         ADC_StructInit(&ADC_InitStructure);
--- a/platform/rct/hal/ADC_stm32/ADC_device_stm32.c
+++ b/platform/rct/hal/ADC_stm32/ADC_device_stm32.c
@@ -25,6 +25,7 @@
 #include "stm32f30x_dma.h"
 #include "stm32f30x_gpio.h"
 #include "stm32f30x_rcc.h"
+#include "taskUtil.h"
 #include <stdbool.h>
 #include <stdint.h>
 
@@ -34,8 +35,8 @@
  */
 
 #define ADC_SYSTEM_VOLTAGE_RANGE	20.0f
-#define SCALING_BATTERYV	0.00465f
-#define TOTAL_ADC_CHANNELS	1
+#define SCALING_BATTERYV		0.00465f
+#define TOTAL_ADC_CHANNELS		1
 
 volatile unsigned short ADC_Val[TOTAL_ADC_CHANNELS];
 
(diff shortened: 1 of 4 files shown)


platform/rct/hal/ADC_stm32/ADC_device_stm32.c
```diff
@@ -76,54 +79,44 @@ int ADC_device_init(void)
...
-
-        ADC_InitTypeDef         ADC_InitStructure;
+        /* Calibration procedure */
+        ADC_VoltageRegulatorCmd(ADC3, ENABLE);
+        /*
+         * Wait at least 10us before starting calibration or enabling.
+         * This is a total hack since I don't want to put a timer here
+         * and there is no interrupt to tell me this is ready :(.
+         */
+        for (size_t i = 0; i < 10000; ++i);
```
brentpicasso: why don't we want a delayMs here? are we not running in the context of a freeRTOS task?

platform/rct/hal/ADC_stm32/ADC_device_stm32.c
```diff
@@ -76,54 +79,44 @@ int ADC_device_init(void)
...
+
+        /* We compare against Vref only */
+        ADC_SelectDifferentialMode(ADC3, ADC_Channel_1, DISABLE);
+
+        /* Now calibrate our ADC */
+        ADC_SelectCalibrationMode(ADC3, ADC_CalibrationMode_Single);
+        ADC_StartCalibration(ADC3);
+        while(SET == ADC_GetCalibrationStatus(ADC3));
```
brentpicasso: No chance of this never returning, I take it? It's down at the processor level so it's be a processor bug if it never returned. Those never happen, right? :)

brentpicasso:
only a couple of comments here.

platform/rct/hal/ADC_stm32/ADC_device_stm32.c
stieg: Good call.  I was still running under the assumption that it would be possible to this to be invoked before the scheduler was called.  If that is still possible then no, else yes.

platform/rct/hal/ADC_stm32/ADC_device_stm32.c
stieg: Heh... there is no chance that I can think of off-hand.

platform/rct/hal/ADC_stm32/ADC_device_stm32.c
```diff
@@ -76,54 +80,46 @@ int ADC_device_init(void)
...
-        ADC_CommonInitStructure.ADC_Mode = ADC_Mode_Independent;
-        ADC_CommonInitStructure.ADC_Clock = ADC_Clock_AsynClkMode;
-        ADC_CommonInitStructure.ADC_DMAAccessMode = ADC_DMAAccessMode_Disabled;
-        ADC_CommonInitStructure.ADC_DMAMode = ADC_DMAMode_Circular;
-        ADC_CommonInitStructure.ADC_TwoSamplingDelay = 0;
-        ADC_CommonInit(ADC3, &ADC_CommonInitStructure);
+        /* Now calibrate our ADC */
+        ADC_SelectCalibrationMode(ADC3, ADC_CalibrationMode_Single);
+        ADC_StartCalibration(ADC3);
+        while(SET == ADC_GetCalibrationStatus(ADC3));
```
Jeff-Ciesielski: Perhaps it wouldn't be a terrible idea to put a timeout here to prevent a deadlock?

stieg: If its broken then the chip is fundamentally broken.  I think we will be ok.

brentpicasso: agreed

Issue #703, opened 2016-06-24
Get -DUSE_FULL_ASSERT working


stieg:
This flag works well on RCT but hits some bugs with MK2.  Lets figure out why and get it back in so we can build with more confidence.

PR #710, opened 2016-06-29
Full Assert Support

stieg:
This change enables us to use the `assert_param` methods in firmware, allowing us to have assert checks in debug (non-tagged) versions of the firmware.  This better allows us to validate the inputs we pass into hardware and to catch potentially tricky bugs during development.

It did expose several bugs in MK2, half of which were semi-legit to full legit (casting issues with USART).  Those are all fixed as part of this release.

### DIFF
--- a/platform/mk2/hal/cpu_stm32/cpu_device_stm32.c
+++ b/platform/mk2/hal/cpu_stm32/cpu_device_stm32.c
@@ -1,5 +1,6 @@
 #include "cpu_device.h"
 #include "portmacro.h"
+#include "printk.h"
 #include <stm32f4xx_misc.h>
 #include <stm32f4xx_rcc.h>
 #include <core_cm4.h>
@@ -12,7 +13,13 @@
 #define SERIAL_ID_BITS		96
 #define SERIAL_ID_BUFFER_LEN	(((SERIAL_ID_BITS / 8) * 2) + 1)
 
-extern uint32_t _flash_start;
+/*
+ * Set by f407_mem.ld linker script.  Somehow this is getting
+ * altered to a value of 0x20020000.  Don't know why yet.
+ * But we handle this below by masking off the top 12 bits.
+ */
+extern const uint32_t _flash_start;
+
 static char cpu_id[SERIAL_ID_BUFFER_LEN];
 
 static void init_cpu_id()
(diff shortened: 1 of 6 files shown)


stieg:
Addresses issue #703

platform/mk2/hal/cpu_stm32/cpu_device_stm32.c
```diff
@@ -12,7 +13,13 @@
 #define SERIAL_ID_BITS		96
 #define SERIAL_ID_BUFFER_LEN	(((SERIAL_ID_BITS / 8) * 2) + 1)
 
-extern uint32_t _flash_start;
+/*
+ * Set by f407_mem.ld linker script.  Somehow this is getting
+ * altered to a value of 0x20020000.  Don't know why yet.
+ * But we handle this below by masking off the top 12 bits.
+ */
```
Jeff-Ciesielski: I'm so confused about this....  Did a quick grep to see if it gets touched anywhere and didn't come up with anything...

platform/mk2/libs/STM32F4xx_DSP_StdPeriph_Lib_V1.0.1/Libraries/STM32F4xx_StdPeriph_Driver/inc/stm32f4xx_dma.h
```diff
@@ -356,8 +356,8 @@ typedef struct {
 /** @defgroup DMA_flags_definition
   * @{
   */
-#define DMA_FLAG_FEIF0                    ((uint32_t)0x10800001)
-#define DMA_FLAG_DMEIF0                   ((uint32_t)0x10800004)
+#define DMA_FLAG_FEIF0                    ((uint32_t)0x10000001)
+#define DMA_FLAG_DMEIF0                   ((uint32_t)0x10000004)
```
Jeff-Ciesielski: wat?

stieg: Yep.  The 8 in those first 2 lines seems to be a bug since they mask it out later.  It was causing the assert to fire when I had it enabled.

Jeff-Ciesielski:
I think there should probably be a background story to figure out WTF is going on with the flash origin, but otherwise :shipit:

Issue #698, opened 2016-06-23
USB comms issue after writing a script (2.10.0)


brentpicasso:
After writing a script, USB comms degrades and then fails.  Details:

STR
* r/2.10.0 firmware on MK2
* launch app 
* observe auto-read config is fast.
* make a trivial script change. 
* write script back to MK2
* read config.  observe config is read very slowly.
* close app. 
* re-start app. observe app cannot connect. 
* unplug / replug RCP,  comms is restored.

does not reproduce on 2.9.2

stieg:
Dug into this.  Attached are the log files of a read after the app first starts up and a read later down the line.  Firmware is simply being flooded by requests for samples from the app and is taking time to catch up to commands in its backlog.  This can be seen clearly in the logs.  My guess is our probing overflows the event queue in usb and we just get delayed.  I have a simple fix for this.

[First_load_RCP.txt](https://github.com/autosportlabs/RaceCapture-Pro_firmware/files/340437/First_load_RCP.txt)
[Read_after_write.txt](https://github.com/autosportlabs/RaceCapture-Pro_firmware/files/340438/Read_after_write.txt)

PR #711, opened 2016-06-29
Usb Throughput Fix

stieg:
Fixes the throughput issues that @brentpicasso reported in #698 .  Also fixes a warning that I am seeing locally.


brentpicasso:
:shipit:

Issue #697, opened 2016-06-23
Script fails to load on 2.10.0


brentpicasso:
This GoPro script fails to load on 2.10.0
https://wiki.autosportlabs.com/RaceCapturePro2_GoProWiFi

[lua] Loading script. Length: 2647
[lua] Successfully loaded script.
[lua] Script error: [string "--Specify your GoPro wifi password here..."]:30.0: BUG. Please inform AutosportLabs
[lua] Script error: [string "--Specify your GoPro wifi password here..."]:30.0: BUG. Please inform AutosportLabs
[lua] Script error: [string "--Specify your GoPro wifi password here..."]:30.0: BUG. Please inform AutosportLabs
[lua] Failure: Runtime Error

This script works as expected on 2.9.2

stieg:
Huh... its choking on the first line of the script yet the parser is printing it out as line 30.  That is odd.

brentpicasso:
seems that even a script like this fails:

function onTick()
println("foo")
end

stieg:
Heh... the bug seems to be somewhere in the `println` method

PR #712, opened 2016-06-30
Fix bug in print and println in LUA

stieg:
Missed adding a `break;` statement in a switch.  We were falling
into the Lua Panic code as a result.  Oops.

Issue #697


brentpicasso:
:shipit:

Issue #469, opened 2016-03-10
Get rid of mod_string and mod_* and such since we are using newlib now


stieg:
Using these methods is causing issues with glibc in our test environment.  We shouldn't be using them anymore and should instead use our GLibc bits (newlib).

PR #713, opened 2016-06-30
Remove modp_aton library and replace with stdlib

stieg:
Removes the modp_aton library and replaces it with the much better
tested newlib suite of functions which do the same thing only
have a much better pedigree.

Issue #469


brentpicasso:
Good cleanup, and the right time to do it. :shipit:

stieg:
Thanks for the review @brentpicasso

Issue #714, opened 2016-06-30
Fix SD Logging on MK2


stieg:
Doesn't want to work for some reason.  Need to figure out why.

PR #719, opened 2016-07-04
Get SDCard Logging Functional Again

stieg:
SD Card functionality broke on MK2 in 2.10.0 sometime back.  This PR fixes it by resolving the following issues:

* Corrects a buffer overflow we were having when calculating the CPU serial number.  Now we no longer write outside the bounds of our array.
* Corrects some bad logic in the fileWriter.c where we could get stuck in an infinite loop.
* Fixes the IRQ levels to not have higher priority than the FreeRTOS subsystem.

Resolves issue #714 and #647

### DIFF
--- a/platform/mk2/hal/cpu_stm32/cpu_device_stm32.c
+++ b/platform/mk2/hal/cpu_stm32/cpu_device_stm32.c
@@ -51,7 +51,7 @@ static void init_cpu_id()
 {
         const uint8_t* ids = (const uint8_t*) CPU_ID_REGISTER_START;
 
-        for (int i = 0; i < CPU_ID_BYTE_COUNT; ++i) {
+        for (size_t i = 0; i < CPU_ID_BYTE_COUNT; ++i) {
                 cpu_id[2 * i] = to_hex(ids[i] >> 4);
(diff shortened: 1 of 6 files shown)


platform/mk2/hal/cpu_stm32/cpu_device_stm32.c
```diff
@@ -22,19 +41,20 @@ extern const uint32_t _flash_start;
...
-            cpu_id[7 - i + j] = ASCII(*p >> (i * 4));
-        p++;
-        j += 8;
-    }
+static void init_cpu_id()
+{
+        const uint8_t* ids = (const uint8_t*) CPU_ID_REGISTER_START;
 
-    cpu_id[SERIAL_ID_BUFFER_LEN - 1] = 0;
+        for (int i = 0; i < CPU_ID_BYTE_COUNT; ++i) {
```
brentpicasso: size_t instead of int?

stieg: Sure.

stieg:
Also threw in the fix for the NVIC levels now that I can test the SDcard infrastructure again.  Works like a charm.

brentpicasso:
:shipit:

Issue #720, opened 2016-07-04
Remove BT "Name" and "Pass" from API entirely


stieg:
Because it is not appropriate to return them anymore since they should only be settable.

Issue #715, opened 2016-06-30
Don't send BT Name and Password in getConnConfig query


stieg:
Because its not secure to do so and the meaning of those fields is not intuitive when in a getter.  They should only be in the setter.  This was the agreement per discussion with Ryan

PR #721, opened 2016-07-04
Remove retrieval of BT name and password

stieg:
Since we can only set these parameters anyways it makes no sense to
return them in the getter commands for BT informaiton.  Thus we no
longer return them.  Instead we return empty strings since deleting
these would break our API stabiliity standards.  Filed issue #720 to
track removal of those parameters entirely.

Issue #715


brentpicasso:
:ship:

Issue #707, opened 2016-06-28
If User Programs in an Empty Lua script, Disable Lua Runtime


stieg:
Because empty is simply not valid.  Or we could disable Lua entirely if we detect that.

ryandoherty:
I think disabling is fine, users might be confused if they configure an empty script, but then see the default onTick() function appear when they read the config.

stieg:
Yeah... my thought there as well.

brentpicasso:
Vote to simply disable Lua.

Somewhat related, if there's a missing onTick(), we should not consider it an error condition, but just disable the onTick() caller or make it a noop situation. Recently fielded a case where a user was confused by the red error LED.

stieg:
> Somewhat related, if there's a missing onTick(), we should not consider it an error condition, but just disable the onTick() caller or make it a noop situation. Recently fielded a case where a user was confused by the red error LED.

@brentpicasso This is what prompted me to file this.  But I don't like your suggestion because if something goes wrong when they enter the script, I want an error condition.  This is why I opted for this behavior.  If there is no script, then there is nothing to do and that is the end of the story.  If there is a script, then behave strictly.  It makes sense and enforces strict LUA behavior so users don't have issues in the field if they have a bad script entered (say a misspelled `onnTick` method).

brentpicasso:
I think there are legitimate cases where someone would want a one-time script that didn't include an repetitive task.

If they misspelled onTick() then they would see their repetitive behavior not happening, and diagnose accordingly.

stieg:
Plan here per discussion is to simply not error during situation when `onTick` method is not available.  This allows for one time scripts without the need for that callback method and opens the door for a standard of callbacks to be done.

PR #722, opened 2016-07-04
Don't error on missing `onTick` method in Lua

stieg:
This change alters our Lua subsystem to no longer error on a
missing `onTick` method.  This will allow users to create scripts
that run one time during first load of the script and will also
set us up to be able to fire callbacks in future versions should
we choose to do so.

Issue #707

### DIFF
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -166,9 +166,10 @@ static int lua_invocation(struct lua_run_state *rs)
         /* Now run the callback */
         lua_getglobal(rs->lua_state, function);
         if (lua_isnil(rs->lua_state, -1)) {
-                pr_error_str_msg(_LOG_PFX "Function not found: ", function);
+                /* No longer a failure per Issue #707 */
+                pr_debug_str_msg(_LOG_PFX "Function not found: ", function);
                 lua_pop(rs->lua_state, 1);
-                status = LUA_ERR_CALLBACK_NOT_FOUND;
+                status = 0;
                 goto done;
         }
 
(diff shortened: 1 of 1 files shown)


src/lua/luaTask.c
```diff
@@ -166,9 +166,10 @@ static int lua_invocation(struct lua_run_state *rs)
         /* Now run the callback */
         lua_getglobal(rs->lua_state, function);
         if (lua_isnil(rs->lua_state, -1)) {
-                pr_error_str_msg(_LOG_PFX "Function not found: ", function);
+                /* No longer a failure per Issue #707 */
+                pr_debug_str_msg(_LOG_PFX "Function not found: ", function);
                 lua_pop(rs->lua_state, 1);
-                status = LUA_ERR_CALLBACK_NOT_FOUND;
```
brentpicasso: should we remove this #define since if we're not using it elsewhere?

src/lua/luaTask.c
```diff
@@ -166,9 +166,10 @@ static int lua_invocation(struct lua_run_state *rs)
...
+                status = 0;
```
brentpicasso: should we add a #define for indicating "no lua error"?

src/lua/luaTask.c
stieg: Yeah was debating that myself.  Can't hurt.

src/lua/luaTask.c
stieg: Ok

brentpicasso:
:shipit:

Issue #709, opened 2016-06-29
Add capabilities flags to firmware


stieg:
As part of our discussion today we will add a "flags" section to the `getCapabilities` JSON method in firmware.  This will return a list of strings that denote the various capabilities of our system, similar to how its done in linux kernel in /proc/cpuinfo

stieg:
Ok I know we need the following:

* activeTrack - Denotes ability to set the activeTrack in non-volatile fashion
* wifi - Denotes wifi support
* bluetooth - Denotes Bluetooth support
* telemStream - Denotes api support for telemetry streaming endpoint.

What other flags did we need.  Do you want ones for GPIO and analogue channels etc?  There are some limitations like the fact that we have support for reading the battery, yet don't really have much control over it (other than frequency of updating)

brentpicasso:
if you want to be consistent, you can have flags for all hardware capabilities. The other existing capabilities objects can describe how many of some support might exist. 

adc, timer, pwm, imu, gps, racetiming, gpio, can, obd, lua

any other ones?

PR #723, opened 2016-07-04
Add flags support to "getCapabilities" API

stieg:
Flags tell the caller what features will be supported on the unit.
This allows us to selectively enable/disable windows in the app so
that the user only sees what they need to see.  This also allows us
to have the app be flexible as we add new features going forward while
retaining backwards compatibility as much as possible.

Issue #709

### DIFF
--- a/src/system/flags.c
+++ b/src/system/flags.c
@@ -39,9 +39,9 @@
  * - gps Global Positioning Satellite support
  * - imu Inertia Measurement Unit support
  * - lua Lua scripting support
- * - pwmin Pulse width modulation measurement input support
- * - pwmout Pulsw width modulation generation output support
+ * - pwm Pulsw width modulation generation output support
  * - telemstream Supports telemetry streaming API
+ * - timer Timed pulse frequency measurement support
  * - tracks Track DB support
  * - usb USB connectivity support
  * - wifi WiFi support
@@ -71,13 +71,13 @@ static const char* feature_flags[] = {
 #if CAN_CHANNELS > 0
         FEATURE_FLAG("obd2")
 #endif
-#if TIMER_CHANNELS > 0
-        FEATURE_FLAG("pwmin")
-#endif
 #if PWM_CHANNELS > 0
-        FEATURE_FLAG("pwmout")
+        FEATURE_FLAG("pwm")
 #endif
         FEATURE_FLAG("telemstream")
+#if TIMER_CHANNELS > 0
+        FEATURE_FLAG("timer")
+#endif
 #if MAX_TRACKS > 0
         FEATURE_FLAG("tracks")
 #endif
--- a/src/system/flags.c
+++ b/src/system/flags.c
@@ -1,0 +1,93 @@
+/*
+ * Race Capture Firmware
+ *
+ * Copyright (C) 2016 Autosport Labs
+ *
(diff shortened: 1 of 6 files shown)


src/system/flags.c
```diff
@@ -0,0 +1,93 @@
...
+#if GPIO_CHANNELS > 0
+        FEATURE_FLAG("gpio")
+#endif
+        FEATURE_FLAG("gps")
+        FEATURE_FLAG("imu")
+#if LUA_SUPPORT > 0
+        FEATURE_FLAG("lua")
+#endif
+#if PWM_CHANNELS > 0
+        FEATURE_FLAG("pwmin")
```
brentpicasso: this should be pwmout

src/system/flags.c
```diff
@@ -0,0 +1,93 @@
...
+#endif
+#if TIMER_CHANNELS > 0
+        FEATURE_FLAG("pwmout")
```
brentpicasso: I get the symmetry of the pwm in/out,  yet at the same time they canonical name in the system is 'timer'.  Thinking we should be consistent with that naming.

src/system/flags.c
```diff
@@ -0,0 +1,93 @@
...
+static const char* feature_flags[] = {
+        FEATURE_FLAG("activetrack")
+#if ANALOG_CHANNELS > 0
+        FEATURE_FLAG("adc")
+#endif
+#if BLUETOOTH_SUPPORT > 0
+        FEATURE_FLAG("bt")
+#endif
+#if CAN_CHANNELS > 0
+        FEATURE_FLAG("can")
```
brentpicasso: should add obd2 support.  unless we say, with CAN we automatically get OBD2 and later direct mapping of CAN data.

src/system/flags.c
stieg: Do I have it backwards then?  Timer is pwm input measurement and `pwm` is output generation

src/system/flags.c
stieg: I thought about that too.  Problem is that the name `timer` is a poor name for what it represents.  Timer implies a stopwatch or some form of timing measurement.  While PWM measurement technically is this, it is a poor use of that name of a flag IMHO.  This is why I was opting for something more concrete like `pwmin` and `pwmout`

src/system/flags.c
stieg: I'll add it.

src/system/flags.c
brentpicasso: So what we're doing with the timer inputs is measuring frequency (or RPM) by measuring duration between incoming pulses - aka timing the pulses the pulses, so 'timer' is an appropriate name that covers all of these use cases.

stieg: Yeah... I get why you chose that name but it really only makes sense to you.  Anyone coming in from an outsider perspective would assume `timer` to mean something entirely different and likely not have it be associated with PWM input measurement.

brentpicasso: PWM = Pulse Width Modulation. PWM is a different concept than frequency. This means the width of the pulse, not the time between the full cycle of repeated pulses. The time between pulse intervals is what's used to measure frequency and RPM.  Therefore, timer is the accurate term for these inputs.

brentpicasso:
:shipit:

PR #724, opened 2016-07-05
Fix minor irq handler bug in RCT USB stack

stieg:
There was a case where we may not properly wake up the FreeRTOS
scheduler in our send logic in USB.  This patch resolves that little bug.


Jeff-Ciesielski:
Looks sane.  :shipit:

stieg:
thanks @Jeff-Ciesielski

Issue #676, opened 2016-06-13
Add WiFi status to system status API


brentpicasso:
To facilitate production testing, and to provide useful info to customers, we should provide WiFi status in the system status API. 

at minimum for production testing we would need an indicator that the WiFi module can be communicated with - an 'initialized' status would cover this.

PR #726, opened 2016-07-06
Add WiFi section to status endpoint

stieg:
Adds support for retrieving WiFi information in the status endpoint.
It includes info on whether the client or ap modes are active and if
the client is connected to its intended network.

Issue #676

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -897,7 +897,7 @@ static void check_wifi_client()
                  * the correct client network.  If not, try to get on
                  * the correct network.
                  */
-                if (!ci->has_ap || !STR_EQ(ci->ssid, cfg->ssid)) {
+                if (!esp8266_drv_client_connected()) {
                         /* Then we need to try and join the AP */
                         client_try_join_ap();
                         return;
@@ -1203,6 +1203,11 @@ bool esp8266_drv_update_client_cfg(const struct wifi_client_cfg *cc)
         return true;
 }
 
+const struct wifi_client_cfg* esp8266_drv_get_client_config()
+{
+        return esp8266_state.client.config;
+}
+
 bool esp8266_drv_update_ap_cfg(const struct wifi_ap_cfg *wac)
 {
         if (NULL == wac)
@@ -1218,6 +1223,11 @@ bool esp8266_drv_update_ap_cfg(const struct wifi_ap_cfg *wac)
         cmd_set_check(CHECK_WIFI_AP);
 
         return true;
+}
+
+const struct wifi_ap_cfg* esp8266_drv_get_ap_config()
+{
+        return esp8266_state.ap.config;
 }
 
 static bool init_channel_sync_op(struct channel_sync_op* op)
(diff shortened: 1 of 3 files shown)


src/logger/loggerApi.c
```diff
@@ -267,6 +269,29 @@ int api_getCapabilities(struct Serial *serial, const jsmntok_t *json)
...
+        const struct wifi_client_cfg* clt_cfg = esp8266_drv_get_client_config();
+
+        const bool ap_active = ap_cfg && ap_cfg->active;
+        const bool client_active = clt_cfg && clt_cfg->active;
+        const bool client_connected = esp8266_drv_client_connected();
+
+        json_objStartString(serial, "wifi");
+
+        json_objStartString(serial, "ap");
+        json_bool(serial, "active", ap_active, false);
```
ryandoherty: Number of clients connected to the AP would be handy to have, if simply for debugging purposes. Also could be useful if someone is seeing weird activity with their RCP, would be helpful to know if there was another device connected to it.

ryandoherty:
API info looks good to me other than my 1 suggestion, can't comment on the C code :)

src/logger/loggerApi.c
stieg: Let me see what I can do there.

stieg: After discussing with @ryandoherty in Slack we are going to leave things as is until we know more about what is helpful for diagnostic info.  Then we can add new metrics as needed.

src/drivers/esp8266_drv.c
```diff
@@ -1466,3 +1476,14 @@ const struct esp8266_ipv4_info* get_ap_ipv4_info()
...
+
+/**
+ * Tells us if the WiFi client is connected to the correct WiFi network.
+ * @return true if it is, false otherwise.
+ */
+bool esp8266_drv_client_connected()
+{
+        const struct wifi_client_cfg *cfg = esp8266_state.client.config;
+        const struct esp8266_client_info *ci = &esp8266_state.client.info;
+        return cfg && ci->has_ap && STR_EQ(ci->ssid, cfg->ssid);
```
brentpicasso: curious, do we strip whitespace from the cfg->ssid when programmed?

src/logger/loggerApi.c
```diff
@@ -326,7 +351,9 @@ int api_getStatus(struct Serial *serial, const jsmntok_t *json)
         json_int(serial, "trackId", lapstats_get_selected_track_id(), 1);
         json_int(serial, "inLap", (int)lapstats_lap_in_progress(), 1);
         json_int(serial, "armed", lc_is_armed(), 0);
-        json_objEnd(serial, 0);
+        json_objEnd(serial, true);
+
+        get_wifi_status(serial, false);
```
brentpicasso: suggest renaming this function to send_wifi_status to be consistent with other functions with similar purposes. Also, more closely matches what the function does.

src/drivers/esp8266_drv.c
stieg: Currently we do not.  Would be easy to do.

brentpicasso: K. up to you if you think it's a separate PR.

src/logger/loggerApi.c
stieg: `get_wifi_status` is consistent with other names like `getStatus` or `get_wifi_client`.  I like the get/put methodologies

brentpicasso: I'm ok with that. recommend filing an issue to track refactoring the other functions so the naming is consistent throughout.

brentpicasso:
:shipit:  with comments above.

Issue #675, opened 2016-06-13
Add IMU status to system status info


brentpicasso:
To facilitate production testing, we could return IMU initialization status in our status API

PR #727, opened 2016-07-06
IMU information in getStatus API

stieg:
Returns some very basic information about our IMU in the status API.
Currently only returns whether or not it is initialized.  This can
be expanded later as needed.

Issue #675

### DIFF
--- a/src/logger/loggerApi.c
+++ b/src/logger/loggerApi.c
@@ -268,7 +268,7 @@ int api_getCapabilities(struct Serial *serial, const jsmntok_t *json)
         return API_SUCCESS_NO_RETURN;
 }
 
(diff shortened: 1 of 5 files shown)


src/logger/loggerApi.c
```diff
@@ -267,6 +268,16 @@ int api_getCapabilities(struct Serial *serial, const jsmntok_t *json)
         return API_SUCCESS_NO_RETURN;
 }
 
+static void imu_status(struct Serial *serial, const bool more)
```
brentpicasso: suggest naming 'send_imu_status' - consistent with other similar functions. Similar comment in other PR.

brentpicasso:
:shipit:

Issue #169, opened 2015-10-26
Speed based auto-start/stop logging


ryandoherty:
Rough spec:

- User can enable/disable this feature
- User can configure the speed to start/stop logging
- Some simple cooldown period is needed to prevent flapping. Speed => N for 1s == start. Speed < N for 5s == stop. Behavior is it should start logging 'asap' and be reluctant at stopping to prevent missing data.
- Default for the firmware should be 5mph and enabled. 
- App will have a new section or feature to allow users to configure easily
- Lua start/stop logging can still override this behavior. RCP should log to the system log how SD card logging was enabled and disabled so it can be determined what started and stopped the SD card logging.

brentpicasso:
Also - 

"is engine running" auto logging option.  Could be based on battery voltage, or RPM channel

stieg:
Lets try and sneak this in as part of 2.10

brentpicasso:
agreed. it's time.

stieg:
So for this revision of the feature I am going to do speed based auto start/stop only.  RPM adds complications because I don't know which channel to monitor or which channel is the correct RPM channel to monitor (since there is a chance there could be more than one).

The main question I have now are what our default values should be.  Should this be enabled by default since its a new feature and users are not used to having it?  I can see both sides of the fence here really.  On one hand if they don't have an SD card installed we will try to log to something that doesn't exists and that will turn on the red light and may create user issues.  On the other hand we want things to "just work".  I'm typically of the camp of "make the user enable the feature" so they explicitly know they did it.  Tends to create a bit less confusion that way.  Thoughts?

ryandoherty:
+1 for just speed based for now. Simple and effective.

Defaulting to on for current customers could cause problems, but defaulting to off will make our new customers' experience worse. If we didn't have LuaScript, I'd be 100% behind enabling it by default. 

I'm thinking we have the default of it being on and log when speed is > 5MPH. We'd need some very strong messaging for users. 

Having a red light turn on if there's no SD card is fine, by default users will want to log data so RCP should tell people when it can't. If they don't want SD card logging, they can turn it off.

brentpicasso:
I agree making this simple for now, advise on designing the configuration API with extensibility in mind.

To trigger on other sensors we could maintain a list of other sensor inputs to trigger on and make that configurable by sensor index (0-8 for analog) (0-2 for timer) (0-2 for GPIO) etc. Start and stop thresholds would likely be the same.

[...]

PR #731, opened 2016-07-08
Auto Logger

stieg:
Resolves issue #169 by adding speed based logging start and stop support that can be used tuned to their needs.  User must maintain X or above given speed for Y seconds to trigger a start event.  Conversely a user must maintain A or below speed for B seconds to trigger a stop event.

Still need to test but code compiles and all tests pass.


src/logger/auto_logger.c
```diff
@@ -0,0 +1,185 @@
...
+        json_float(serial, "speed", alst->speed, 2, true);
+        json_uint(serial, "time", alst->time, false);
+        json_objEnd(serial, more);
+}
+
+void auto_logger_get_config(struct auto_logger_config* cfg,
+                            struct Serial* serial,
+                            const bool more)
+{
+        json_objStartString(serial, "autoLoggerCfg");
```
brentpicasso: Thoughts on how would this config structure be extensible for triggering on other sensors in the future?

src/gps/gpsTask.c
```diff
@@ -64,6 +67,8 @@ void GPSTask(void *pvParameters)
                         if (result == GPS_MSG_SUCCESS) {
                                 const GpsSnapshot snap = getGpsSnapshot();
                                 lapstats_processUpdate(&snap);
+                                auto_logger_gps_sample_cb(&snap.sample);
```
brentpicasso: I'd imagine in the future when triggering on other samples we'd move this somewhere central. loggerTask?

brentpicasso: err... I meant, when triggering on other sensors we'd move it to a central location, like logger task.

src/logger/auto_logger.c
stieg: Discussed on Hangouts.  No change for now as anything more complex can be done in Lua

src/gps/gpsTask.c
stieg: Yeah... probably that or a timer task or something along those lines.

stieg:
Hey @brentpicasso any further comments here?  I would like to merge is this soon since I have other patches that are built upon it that are waiting.

brentpicasso:
@stieg I don't see any updates based on the conversation we had; where the auto logger state would maintain a separate logger state so that users can manually trigger a logging session and not have the autologger interfere.  Are you still going to add that?

stieg:
@brentpicasso That would be because I forgot to push it.  Done now.

brentpicasso:
Looks sane. I'm not in a place to do a real-world test. Did you @stieg  ?  If you did and works, I say :shipit:

stieg:
Tested it last Friday using my RC Car test rig.  Worked wonderfully.

Issue #628, opened 2016-05-17
LUA setGpio not accepting digits for 2nd parameter.


stieg:
According to our wiki we should support both numbers and boolean values for this field.  Also the wiki needs to be updated to reflect valid values for both parameters.

PR #732, opened 2016-07-10
Add support for flexible boolean in Lua

stieg:
In some of our Lua API calls we are using 0 and 1 values where we
should really be using false and true values.  These include

* setGpio
* setLed

This patch adds some new methods to allow users to either specify a
boolean parameter or use the legacy behavior and specify a digit to be
interpreted as a boolean.  0 is false and everything else is true.

Issue #628


brentpicasso:
:shipit:

Issue #550, opened 2016-05-02
[RCT] CAN Tx has no timeout support or error checking


stieg:
Looking at the code for RCT, there are no checks to ensure that an outgoing CAN message actually received a mailbox to send its message in.  Also we have no support for timeouts, and that is not good.

Need to fix both of these issues.  MK2 code has working code we can easily port here.

PR #733, opened 2016-07-10
Port timeout CAN Tx logic to RCT, minor improvements to MK2 CAN

stieg:
Ports the CAN Tx logic from MK2 to RCT so that we may support timeout
capabilities on RCT for sending messages.  Also makes a minor improvement
in both stacks by making a variable a constant and sorting includes.

Issue #550


brentpicasso:
:shipit:

Issue #470, opened 2016-03-12
Provide ability to set double value for pulses/revolution for RPM inputs


stieg:
As @kosenko-max reported in https://github.com/autosportlabs/RaceCapture_App/issues/900

> In my case I'm using RPM input for EFR Turbo Speed Sensor. It makes 5V pulse every 8th time it sees compressor blade passing the sensor. I have 14 blades compressor wheel. So one revolution generates 14/8=1.75 pulses which I can't input there. If I put 2 pulses per revolution - that's 15% off - up to 22500 RPM error.

> So I input 29 (closest to 14/8*1000/60=29.1(6)) to get KRPM on a Frequency channel. Unfortunately, it rounds up everything to 0 digit precision whatever precision for channel I set. So it's off by 0.6% + 1000 RPM. That is total maximum error of 1900 RPM with half of it as uncertainty.

> Ability to set double value for pulses per revolution (so I can input 1.75) would set maximum error below 1 RPM without uncertainty due to the rounding.

This ticket tracks adding this type of support to the firmware.  It would need to be plugged in first before we could implement the fix in the app.

stieg:
Merged

PR #735, opened 2016-07-10
Float value for pulses per revolution

stieg:
Not all pulse per revolution counts line up on whole numbers.
To account for that I have moved this value over to a float value
so that our users may specify arbitrary numbers for this.

I have also removed the filter check and instead added checks for
0 so that should a user pass in a value of 0 we will simply return
0 values and handle it accordingly.

Issue #470

### DIFF
--- a/src/logger/loggerApi.c
+++ b/src/logger/loggerApi.c
@@ -1188,7 +1188,7 @@ static void sendTimerConfig(struct Serial *serial, size_t startIndex, size_t end
         json_uint(serial, "st", 0, 1);
         json_uint(serial, "mode", cfg->mode, 1);
         json_float(serial, "alpha", cfg->filterAlpha, FILTER_ALPHA_PRECISION, 1);
-        json_float(serial, "ppr", cfg->pulsePerRevolution, 2, 1);
+        json_float(serial, "ppr", cfg->pulsePerRevolution, 6, 1);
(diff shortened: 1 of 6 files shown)


src/logger/loggerApi.c
```diff
@@ -1189,7 +1188,7 @@ static void sendTimerConfig(struct Serial *serial, size_t startIndex, size_t end
         json_uint(serial, "st", 0, 1);
         json_uint(serial, "mode", cfg->mode, 1);
         json_float(serial, "alpha", cfg->filterAlpha, FILTER_ALPHA_PRECISION, 1);
-        json_uint(serial, "ppr", cfg->pulsePerRevolution, 1);
+        json_float(serial, "ppr", cfg->pulsePerRevolution, 2, 1);
```
brentpicasso: is 2 decimal points adequate here? wondering if this should be 4 or 6 

1.333333? :)

brentpicasso:
looks sane. just one question / suggestion.   Resolve if needed and :shipit:

src/logger/loggerApi.c
stieg: Good argument.  I will bump to 6.

stieg:
Thanks for the feedback @brentpicasso .  Merging.

Issue #728, opened 2016-07-06
Strip Whitespace when putting WiFi SSID on API


stieg:
So we don't get a bad value in there since whitespace is not valid in an SSID

PR #736, opened 2016-07-11
Add strip feature to json setter; strip ssid and other values

stieg:
Now we can strip user input of whitespace if needed.  This is wise
since whitespace is not valid in several places.  This change adds
that support along with supporting stripping of SSID values for wifi.

Issue #728

### DIFF
--- a/include/jsmn/jsmn.h
+++ b/include/jsmn/jsmn.h
@@ -135,7 +135,8 @@ bool jsmn_exists_set_val_bool(const jsmntok_t* root, const char* field,
                               void* val);
 
 bool jsmn_exists_set_val_string(const jsmntok_t* root, const char* field,
-                                void* val, const size_t max_len);
+                                void* val, const size_t max_len,
+				const bool strip);
 
 CPP_GUARD_END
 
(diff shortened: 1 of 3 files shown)


include/jsmn/jsmn.h
```diff
@@ -135,7 +135,8 @@ bool jsmn_exists_set_val_bool(const jsmntok_t* root, const char* field,
                               void* val);
 
 bool jsmn_exists_set_val_string(const jsmntok_t* root, const char* field,
-                                void* val, const size_t max_len);
+                                void* val, const size_t max_len,
+				const bool strip);
```
Jeff-Ciesielski: Mixed whitespace here.

Jeff-Ciesielski:
Is this the traditional notion of "strip" i.e. "strip whitespace from ends", or does the strip inline function literally strip out all whitespace?

brentpicasso:
:shipit:

include/jsmn/jsmn.h
stieg: Strange.  It looked fine on my console.  Will double check.

stieg:
> Is this the traditional notion of "strip" i.e. "strip whitespace from ends", or does the strip inline function literally strip out all whitespace?

It strips leading and trailing whitespace.  The reference inline means that we will not allocate new memory to achieve the strip.  Instead we modify the string in place

Jeff-Ciesielski:
🚢

Issue #684, opened 2016-06-16
Add support for changing WiFi Baud Rate


stieg:
Need to be able to talk to the chip at a higher baud rate.  Currently at 115200.  MK2 can go to 230400 and RCT can go to 921600 in theory.  Implement this support.

PR #740, opened 2016-07-13
Support Higher WiFi Baud Rate; Improve Init Process

stieg:
Initially I just wanted to add support for a faster baud rate for the WiFi device.  However after completing a working prototype I ran into many many issues with how the wifi init system worked and frankly determined that it was broken and needed to be fixed before this baud rate change could go in.  Thus why this patch is so very large.

This PR does two things.  One is that it makes our Wifi init process sane.  By sane I mean we guarantee that we will always get back to a useable and readable state after a reset (either soft or hard) and that we can find a Wifi device at all times using a basic serial connection if we know the baud rates it could be at.

Then we added support for incrementing the baud rate to the maximum supported value possible.  In doing so I encountered a bug in the esp8266 subsystem where the uart change message doesn't always return the final `\n` character in its ack reply.  This exposed a need to change how we handle stripping of messages in the AT subsys.

Issue #684

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -25,6 +25,7 @@
 #include "mem_mang.h"
 #include "net/protocol.h"
 #include "printk.h"
+#include "serial.h"
 #include "serial_buffer.h"
 #include "str_util.h"
 #include "taskUtil.h"
@@ -36,13 +37,19 @@
 
 #define _AT_CMD_DELIM		"\r\n"
 #define _AT_DEFAULT_QP_MS	250
-#define _AT_QP_PRE_INIT_MS     	500
-#define _AT_QP_STANDARD_MS	1	/* Can probably be 0 */
+#define _AT_QP_PRE_INIT_MS	500
+#define _AT_QP_STANDARD_MS	1
 #define _AUTOBAUD_TRIES		3
+#define RESET_SLEEP_MS		500
+#define SERIAL_DEF_BAUD		115200
+#define SERIAL_DEF_MSG_BITS	8
+#define SERIAL_DEF_PARITY_BITS	0
+#define SERIAL_DEF_STOP_BITS	1
 #define _TIMEOUT_LONG_MS	5000
 #define _TIMEOUT_MEDIUM_MS	500
 #define _TIMEOUT_SHORT_MS	50
 #define _TIMEOUT_SUPER_MS	30000
+
 
 /* STIEG: Temp until we write *_create methods for serial_buff and at_info */
 struct serial_buffer _serial_cmd_buff;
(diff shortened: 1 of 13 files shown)


src/devices/esp8266.c
```diff
@@ -154,34 +165,42 @@ void esp8266_do_loop(const size_t timeout)
...
-        state.init_cb(true); /* Will always be defined */
+	cmd_failure("Init", msg);
+        init_complete(false);
 }
 
-static bool get_version_cb(struct at_rsp *rsp, void *up)
+static bool init_get_version_cb(struct at_rsp *rsp, void *up)
 {
         if (!at_ok(rsp)) {
                 /* WTF Mate */
```
brentpicasso: lol

src/devices/esp8266.c
```diff
@@ -204,138 +223,141 @@ static bool get_version_cb(struct at_rsp *rsp, void *up)
...
- * @param mux True if we want to, false otherwise.
- * @cb The callback to invoke when complete.
- * @return True if the command was queued, false otherwise.
- */
-static bool set_mux_mode(const bool mux, void (*cb)(bool status))
+static bool init_set_mux_mode()
 {
-        char cmd_str[12];
-        snprintf(cmd_str, ARRAY_LEN(cmd_str), "AT+CIPMUX=%d", mux ? 1 : 0);
+        const char cmd_str[] = "AT+CIPMUX=1";
```
brentpicasso: I guess we always want mux mode?

src/devices/esp8266.c
```diff
@@ -204,138 +223,141 @@ static bool get_version_cb(struct at_rsp *rsp, void *up)
...
-        return true;
+	/* Set normal quiet period */
+        at_configure_device(state.ati, _AT_QP_STANDARD_MS,
+                            _AT_CMD_DELIM);
+
+	/*
+	 * To ensure our device is initialized properly we do the
+	 * following in this order:
+	 *
+	 * + Soft Reset
```
brentpicasso: did we change the hard reset scheme? 

/me keeps looking

brentpicasso:
Tested, performs reliably.

brentpicasso:
You're right, it is a big change - sorry you had to do this to work around the module's weirdnesses. 

I tested it, it seems solid, and nothing major stood out. All clear for me, but @Jeff-Ciesielski  may still have something.

src/devices/esp8266.c
stieg: Yes.  My driver does not support non-mux mode.

src/devices/esp8266.c
stieg: No.  I just soft reset the device every boot since its the only reliable way to get it back to defaults every time and to ensure state is where I expect it to be.

stieg:
Yeah there was a lot here.  Much of it had to do with the fact that the init system was not as reliable as I needed it to be; hence the size of the change increased dramatically.  I'll wait for a some feedback from @Jeff-Ciesielski if he has any

Issue #738, opened 2016-07-12
Change Auto-logger speed value to KPH instead of MPH


stieg:
Because we are trying to standardize on using SI for input.

PR #741, opened 2016-07-13
Change Auto_logger speed parameter to be in KPH

stieg:
We want our API and configs to use SI where possible per a discussion
between Brent, myself and Ryan.  Thus this change makes that a reality
by adjusting the auto logger config to use KPH instead of MPH

Issue #738


brentpicasso:
:shipit:

Issue #742, opened 2016-07-15
Code and test use of CGREG AT command in U-Blox Sara module


stieg:
To test if we can better support Data-only sims this way.

stieg:
Preliminary test passes with AT&T sim on Sara-280.  T-Mobile failed but I believe that is because I have no access to their bands where I'm testing at.

@ryandoherty If you have a data-only SIM from Project Fi I would love your help in testing this.

PR #744, opened 2016-07-15
Move to CGREG command from CREG command

stieg:
After in-depth conversation with our friends at U-Blox we learned
that CREG is a registration command that does more than just GPRS.
It also covers registration of voice services. While this is ok in
many cases it likely will not work with some data-only sims because
they have no voice capabilities and the MNO may configure the SIM to
disallow any registration of voice services (which is logical).

So instead of checking if voice services are registered, this command
will only check if GPRS services are registered.  From there if we
are registered then we can activate a PDP context through the APN.
Hopefully this will add the ability to use more data-only sims as
they become more prevelant.

Issue #742


brentpicasso:
Looks good. Need this tested? If you've already tested it then :shipit:

stieg:
Tested it before I pushed out there review

Issue #536, opened 2016-04-26
Cannot update RCP track maps if non trivial script is loaded


brentpicasso:
If a script is running it is possible that a trackmap cannot be loaded due to insufficient memory. 

STR:
- load a non-trivial script, such as below. 
- add some trackmaps to config.
- observe trackmaps not added upon read-back
- observe error in log console "Failed to allocate memory for track buffer" 

Example script:

``` lua
--This example adopted to Hondata KPro4 CAN output 

--how frequently we poll for CAN messages, 30Hz is the max 
tickRate = 30 
--the CAN baud rate 
CAN_baud = 250000 
--CAN channel to listen on. 0=first CAN channel, 1=second 
CAN_chan = 0 

--add your virtual channels here 
rpmId = addChannel("RPM", 25, 0, 0, 9000, "RPM") 
vltId = addChannel("EcuVolts", 10, 1, 0, 20, "volts") 
iatId = addChannel("IAT", 1, 0, 0, 100, "C") 
ectId = addChannel("EngineTemp", 1, 0, 0, 150, "C") 
tpsId = addChannel("TPS", 10, 0, 0, 100, "%") 
mapId = addChannel("MAP", 10, 1, 0, 15, "PSI") 
injId = addChannel("InjectorPW", 10, 3, 0, 100, "ms") 
ignId = addChannel("Ignition", 10, 0, -20, 20, "D") 
lmdId = addChannel("AFR", 10, 3, 0, 2, "lambda") 
knkId = addChannel("Knock", 1, 0, 0, 15, "count") 
camId = addChannel("CamTiming", 10, 0, -20, 20, "D") 

function toF(value)
 return value * 1.8 + 32
end

function toAFR(value)
 return value * 14.7
end

--customize here for CAN channel mapping 
--offset/length in bytes? 
--format is: [CAN Id] = function(data) map_chan(<channel id>, data, <CAN offset>, <CAN length>, <multiplier>, <adder>) 
CAN_map = { 
--did not bother logging gear speed and target cam angle 
[1632] = function(data) map_chan(rpmId, data, 0, 2, 1, 0) map_chan(vltId, data, 5, 1, 0.1, 0) end, 
[1633] = function(data) map_chan(iatId, data, 0, 2, 1, 0, toF) map_chan(ectId, data, 2, 2, 1, 0, toF) end, 
[...]

PR #745, opened 2016-07-15
Add support for non read modify write operations to flash

stieg:
In cases where it is unreasonable to load the entire DB of a particular
object into RAM we need to be able to write the contents to flash
directly so as to avoid the large/impossible allocation. This change
adds that support so that we may write to flash without the need of
allocating such a ridiculously large buffer.

Issue #536

**NOTE:** We need to discuss if this can actually go into firmware due to existence of https://github.com/autosportlabs/RaceCapture_App/issues/1059 the implications of how this will break backwards compatibility.  If we put it in then at a minimum a user must use the latest app for things to work.

### DIFF
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -27,12 +27,12 @@
 #include "channel_config.h"
 #include "cpp_guard.h"
 #include "geopoint.h"
+#include "macros.h"
 #include "serial_device.h"
 #include "timer_config.h"
 #include "tracks.h"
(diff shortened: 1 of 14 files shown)


brentpicasso:
Looks good on first pass.  Will have to dig in to the #1059 some more.

stieg:
Because we will likely not be able to merge this in with 2.10.0 due to https://github.com/autosportlabs/RaceCapture_App/issues/1059 I am closing this PR.

Issue #748, opened 2016-07-18
Change LapStats API to Enforce Consistent Rate


stieg:
Currently it is possible to set different sample rate values for the various channels that tell us lap information.  This can lead to strange behavior in the system and we can not have it.  As a workaround we are fixing it by enforcing a consistent rate as tracked by issue #503.  However this is merely a workaround.  The long term solution is to do like we did with GPS and change the API so that we only specify one sample rate value that applies to all channels.  We should also include the ability to enable/disable sample rate channels per issue #641 so we can have some form of solution for that.

Issue #503, opened 2016-04-08
Ensure LapStats sample rates are consistent at runtime


brentpicasso:
lapstats channels have different sample rates, but the app hides it via the UI for simplification. 

There's a chance that lapstat sample rates can get out of sync, and can only be corrected if the config is read and lapstats config is written back - leaving the chance of a bad config lingering.

Firmware could compensate at runtime by selecting and normalizing on the highest configured sample rate for all lapstat channels.  

...this is a band-aid until which we modifiy the API to only have a single common samplerate.

stieg:
> There's a chance that lapstat sample rates can get out of sync, and can only be corrected if the config is read and lapstats config is written back - leaving the chance of a bad config lingering.

Do you have any info on how they are getting out of sync?

brentpicasso:
Unsure. likely due to an old configuration state not getting updated from a previous firmware upgrade.

PR #749, opened 2016-07-18
Ensure lapstats channel sample rates are consistent

stieg:
As described by issue #503 some users are sometimes getting inconsistent
sample rate settings in their configs.  This is because older configs
sometimes have these values being separately configurable and thus
can introduce the bug.  While this is technically a bug in the app it
is also a firmware bug since we have yet to change the API to enforce
consistent behavior.  That issue is tracked by issue #748.

Issue #503


brentpicasso:
looks good, man. :shipit:

Issue #737, opened 2016-07-11
Check CAN Filters, Fix if necessary


stieg:
According to https://forum.autosportlabs.com/viewtopic.php?p=24320 the filters on CAN are not functioning correctly.  Need to verify them and fix them if something is indeed broken.

stieg:
Built a loopback device to send messages from CAN 1 to CAN 0 on my MK2.  Wrote the following script to test out the filtering:

```lua
--- SETTINGS ---
-- How frequently we poll for CAN messages
tickRate = 1

-- The CAN baud rate
CAN0_baud = 500000
CAN1_baud = 500000

--CAN channel to listen on. 0=first CAN channel, 1=second
CAN_0 = 0
CAN_1 = 1

-- CAN Filter and Addressess
CAN_REG_ADDR_GOOD_1 = 0xC0
CAN_REG_ADDR_GOOD_2 = 0xCF
CAN_REG_ADDR_FAIL_1 = 0xF0
CAN_REG_FILTER_MASK = 0xF0
CAN_REG_FILTER_VAL  = 0XC0
CAN_REG_FILTER_ID   = 1

CAN_EXT_ADDR_GOOD_1 = 0xC0000
CAN_EXT_ADDR_GOOD_2 = 0xCFFFF
CAN_EXT_ADDR_FAIL_1 = 0xF0000

CAN_EXT_FILTER_MASK = 0xF0000
CAN_EXT_FILTER_VAL  = 0XC0000
CAN_EXT_FILTER_ID   = 2

---- CODE ----
initCAN(CAN_0, CAN0_baud)
initCAN(CAN_1, CAN1_baud)
setCANfilter(CAN_0, CAN_REG_FILTER_ID, 0, CAN_REG_FILTER_VAL, CAN_REG_FILTER_MASK)
setCANfilter(CAN_0, CAN_EXT_FILTER_ID, 1, CAN_EXT_FILTER_VAL, CAN_EXT_FILTER_MASK)
setTickRate(tickRate)

function print_can_msg(chan, id, data)
    print(getUptime())
    print(' [CAN ')
    print(chan)
    print(']: ')
    print(id .. ' => [')

    local comma = false
    for i = 1,#data do
        print(comma and ', ' or '')
        comma = true
        print(data[i])
    end

    print(']\n')
end

function rx_can(chan, wait)
    	local id, ext, data = rxCAN(chan, wait)
	if not id then return false end
[...]

PR #750, opened 2016-07-18
Fix CAN Filtering, CAN Filtering Init Procedure & Add CAN Filter Enable/Disable

stieg:
Does the following for our CAN subsystem:

* Fixes the CAN filtering issue as tracked by #737 
* Fixes the Init routine to reset all filters for a given CAN channel except filter 0 which is set to receive all.
* Adds support to LUA so users may enable/disable filters.

### DIFF
--- a/platform/mk2/hal/CAN_stm32/CAN_device_stm32.c
+++ b/platform/mk2/hal/CAN_stm32/CAN_device_stm32.c
@@ -29,15 +29,17 @@
 #include "stm32f4xx_rcc.h"
 #include "task.h"
 #include "taskUtil.h"
+#include <stdbool.h>
 #include <stdint.h>
 #include <string.h>
 
 static xQueueHandle xCan1Rx = NULL;
 static xQueueHandle xCan2Rx = NULL;
 
+#define CAN_FILTER_COUNT	13
+#define CAN_IRQ_PRIORITY	5
+#define CAN_IRQ_SUB_PRIORITY	0
 #define CAN_QUEUE_LENGTH	10
-#define CAN_IRQ_PRIORITY 	5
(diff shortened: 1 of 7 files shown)


platform/mk2/hal/CAN_stm32/CAN_device_stm32.c
```diff
@@ -189,57 +184,76 @@ static void CAN_device_init_2(int baud)
...
-        }
-        pr_info(" win\r\n");
-        return 1;
-    } else {
-        pr_info(" fail\r\n");
-        return 0;
-    }
+	pr_info("Initializing CAN");
+	pr_info_int(channel);
+	pr_info_int_msg(" with baud rate ", baud);
```
brentpicasso: ah eventually we'll get pr_printf()! :)

stieg: Yep.  Its very close.  Currently on float and double printing fail.  Gotta figure out why.  Beyond that we will have it.

stieg:
@brentpicasso does your lack of other commentary mean this is good to ship?

brentpicasso:
Looks good, but hard to tell what change fixed this.  Is it that we left filter 0 alone for the init routine?

stieg:
The big change here was figuring out the correct alignment in the registers for the CAN filter.  This was never correct to begin with.  The CAN PID needs to be left shifted a certain amount depending on whether or not it is a normal id len or an ext id len.  That is why the driver code changed so much.in the `CAN_device_set_filter` methods for both MK2 and RCT.

brentpicasso:
Got it.  :shipit:

PR #756, opened 2016-07-20
Improve our Panic Code

stieg:
Inspired by the events @ryandoherty is facing on his Mac, I have improved our panic code in 2 ways.

1.  Tie in the assertion system to the panic subsys.  Now if we hit an assertion while the debug code is present on the app, the panic code will display as such instead of just seemingly freezing up.

2.  Fix platform spin loop timing issue.  Since the platforms runs at different frequencies, we need to spin at different rates for 1ms to pass.


brentpicasso:
:shipit:

Issue #757, opened 2016-07-20
Enable auto-logger by default


stieg:
Since the app should now have the ability to control this by this firmware release.

Issue #755, opened 2016-07-19
Disable Auto-Logger by Default in 2.10.0


stieg:
Because the app bits may not be ready to control it by the time that 2.10.0 is released.  So disable it by default until the app that can control it is fully available in the field.

PR #758, opened 2016-07-20
Disable autologger by default

stieg:
Since the app code might not be ready to control this feature we
thought it best to disable it now with the intention of enabling
it later by default.  That work is tracked by issue #757

Issue #755


brentpicasso:
:shipit:

Issue #752, opened 2016-07-19
Increase WiFi Telemetry Rate to 50Hz.  Turn Capabilities down to 50Hz for RCT


stieg:
Because I need to ensure this is done now that we have the increased Baud rate.  May need to keep it slower on MK2.  Alos need to slow down capabilities rates for RCT to 50hz max.

PR #759, opened 2016-07-20
Adjust Max Supported Sample Rates For Channels and WiFi

stieg:
Adjusts RCT max sample rate to 50Hz since that is its limit.  Also
adjusts the WiFi system to limit the WiFi stream rate to 50Hz,
up from 10Hz.

Issue #752


brentpicasso:
:SHIPITTTTTTT:

Issue #751, opened 2016-07-19
Disable Track DB on RCT


stieg:
Since this will not be supported on RCT (because of the `setActiveTrack` functionality) we need to ensure that this is sanely disabled.

PR #760, opened 2016-07-20
Disable the track DB on RCT

stieg:
Since the track will be either directly uploaded by the user or
managed by the app per design I am disabling the track DB here.

Issue #751


brentpicasso:
wow that was easy. :shipit:

stieg:
Right?

Issue #734, opened 2016-07-10
Honor mph / kph and mi / km units in firmware


brentpicasso:
Firmware should honor units value for the following channels:

- distance
- speed

if distance channel's units is 'km' then the telemetry stream / logfile stream should record in km. correspondingly, if 'mi'  then we record in miles.

if speed channel units is 'mph' then we record in miles/hour. correspondingly, if 'kph' then we record in km/hour.

Since we don't have full channel definitions defined in the API for speed and distance channels, we could add members to the GPS config:

'speed_ut'
'dist_ut'
'altitude_ut'

these would map to the internal units values for those corresponding channels.

stieg:
Looks like we also need to cover altitude to measure between ft and meters.

stieg:
Ok.  I intend to use the units outline at https://github.com/autosportlabs/RaceCapture-Pro_firmware/wiki/Units-Supported as a guidline for this.  Here is how it will work:

* Altitude: Valid units will be `ft` and `m` (meters).  All other units default to `ft`
* Distance: Valid units will be `km` and `mi` (miles).  All other units default to `mi`
* Speed: valid units are `mph` and `kph`.  All other units default to `mph`

This gets us two items: It solves the ticket and starts us down the road of strict units.

PR #761, opened 2016-07-21
Expose 3 GPS Channel Unit Params, Add Basic Metric Support

stieg:
This large ish PR solves issue #734 by introducing support for basic metric values in our GPS system.  It does this by doing the following:

* Starting the foundation of a strong typed unit system and defining Length and Speed units
* Adjusting the JSON API to allow for the reading/writing of 3 GPS channels (altitude, distance, speed).
* Adjusting the internal method setter in our logger to respect the units decision as set by the API.


src/logger/loggerApi.c
```diff
@@ -1237,42 +1238,82 @@ int api_getGpsConfig(struct Serial *serial, const jsmntok_t *json)
     json_int(serial, "alt", gpsCfg->altitude.sampleRate != SAMPLE_DISABLED, 1);
     json_int(serial, "sats", gpsCfg->satellites.sampleRate != SAMPLE_DISABLED, 1);
     json_int(serial, "qual", gpsCfg->quality.sampleRate != SAMPLE_DISABLED, 1);
-    json_int(serial, "dop", gpsCfg->DOP.sampleRate != SAMPLE_DISABLED, 0);
+    json_int(serial, "dop", gpsCfg->DOP.sampleRate != SAMPLE_DISABLED, 1);
+
+    json_objStartString(serial, "units");
+    json_string(serial, "altitude", gpsCfg->altitude.units, 1);
```
brentpicasso: recommend making these distance and altitude keys consistent with the other keys in this message.

brentpicasso:
It's readable and cohesive.  Only one suggestion on API naming.

src/logger/loggerApi.c
stieg: You mean use the string 'dist' and 'alt'?  Ok.

brentpicasso:
:shipit

PR #764, opened 2016-07-22
Align firmeware in linker to prevent CRC errors in Bootloader

stieg:
Aligns firmware along word boundaries to prevent issues with the
CRC checksum code in the bootloader.  Without it we may get random
failures that make us sad pandas.


Jeff-Ciesielski:
:shipit:

Issue #767, opened 2016-07-25
should make [wifi] Received CMD messages debug level


brentpicasso:
...or remove

These tend to fill the log:

[wifi] Received CMD: {"getStatus":null}

PR #768, opened 2016-07-26
Demote wifi msg received log entry to trace level

stieg:
Because otherwise it pollutes the log file under active polling.
Issue #767


brentpicasso:
:ship:

PR #777, opened 2016-07-29
Support Serial Close

stieg:
This large-ish commit adds the ability for us to be able to "close" a serial device.  This is necessary since in some cases (wifi) the backing serial transport (a TCP session) can die without warning.  Thus we need a sane way to close a serial device without it adversely affecting the system.  This change implements that.

As part of this change I also took the liberty to rename many of the get/put serial methods into read/write methods.  Why?  For one to find all the spots where they were used so that I could ensure that we interpreted the new return values correctly.  Another reason was to make these calls almost the same as those that would be used for f_read and f_write.  The intention that Serial object is becoming our bridge object until we implement full File Stream support to replace it (and oh what a glorious day that will be).

This change also alters how we handle Serial devices on Wifi.  Now the WiFi task is responsible for closing said devices rather than the driver.  This is logical since the driver need only handle creation of Serial devices when asked or when new connections come in or closing Serial devices (but not destroying them) when said connections disappear.

This also optimizes our system a bit so that we use less memory when there are fewer connections.

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -240,8 +240,7 @@ static bool channel_is_open(struct channel* ch)
 }
 
 /**
- * Called when we close a channel.  If we own the channel, reap it
- * completely to avoid memory leaks.  Else simply close it.
+ * Called when we close a channel.
  */
 static void channel_close(struct channel* ch)
 {
@@ -345,13 +344,17 @@ static void socket_state_changed_cb(const size_t chan_id,
                 pr_info_int_msg(LOG_PFX "Socket connected on channel ",
                                 chan_id);
 
-		channel_setup(chan_id);
-		if (!esp8266_state.comm.new_conn_cb) {
-			pr_error(LOG_PFX "No incomming server callback "
-				 "defined. Data \r\n");
-			return;
-		} else {
+		if (!channel_setup(chan_id)) {
+			/* Close the channel.  Best effort here. */
+			esp8266_close(chan_id, NULL);
+			break;
+		}
+
+		if (esp8266_state.comm.new_conn_cb) {
 			esp8266_state.comm.new_conn_cb(ch->serial);
+		} else {
+			pr_warning(LOG_PFX "No incomming server callback "
+				   "defined. Data \r\n");
 		}
                 break;
         default:
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -96,9 +96,6 @@ struct server {
 struct channel {
         struct Serial *serial;
         size_t tx_chars_buffered;
-        bool in_use;
-        bool connected;
-        bool created_externally;
 };
 
 struct channel_sync_op {
@@ -230,14 +227,32 @@ static bool cmd_ready() {
         return !esp8266_state.cmd.in_progress;
 }
 
-/* Begin code for sending/receiving data */
+/* Begin code for sending/receiving data using our internal channels */
 
-static bool is_valid_socket_channel_id(const size_t chan_id)
+static bool channel_is_valid_id(const size_t chan_id)
 {
         return chan_id < ARRAY_LEN(esp8266_state.comm.channels);
 }
 
-static const char* get_channel_name(const size_t chan_id)
+static bool channel_is_open(struct channel* ch)
+{
+	return !!ch->serial;
+}
+
+/**
+ * Called when we close a channel.  If we own the channel, reap it
+ * completely to avoid memory leaks.  Else simply close it.
+ */
+static void channel_close(struct channel* ch)
+{
+	if (!ch->serial)
+		return;
+
+	serial_close(ch->serial);
+	ch->serial = NULL;
+}
+
+static const char* channel_get_name(const size_t chan_id)
 {
         static const char* _serial_names[] = {
                 "Wifi Chan 0",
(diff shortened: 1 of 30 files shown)


src/drivers/esp8266_drv.c
```diff
@@ -230,14 +227,32 @@ static bool cmd_ready() {
...
 }
 
-static const char* get_channel_name(const size_t chan_id)
+static bool channel_is_open(struct channel* ch)
+{
+	return !!ch->serial;
+}
+
+/**
+ * Called when we close a channel.  If we own the channel, reap it
```
stieg: Need to fix this comment I see

src/drivers/esp8266_drv.c
```diff
@@ -260,112 +275,126 @@ static void _tx_char_cb(xQueueHandle queue, void *post_tx_arg)
...
-                return NULL;
 
-        struct channel* ch = esp8266_state.comm.channels + i;
-        if (ch->serial)
-                return ch;
+	if (!s) {
+		/* Fail state */
+		pr_warning(LOG_PFX "Insufficient resources for new "
+			   "Serial\r\n");
+		return NULL;
```
stieg: Ensure that failure here leads to a channel clousre.

stieg: This has been addressed.

src/drivers/esp8266_drv.c
```diff
@@ -260,112 +274,130 @@ static void _tx_char_cb(xQueueHandle queue, void *post_tx_arg)
...
+		if (!channel_setup(chan_id)) {
+			/* Close the channel.  Best effort here. */
+			esp8266_close(chan_id, NULL);
+			break;
+		}
+
+		if (esp8266_state.comm.new_conn_cb) {
+			esp8266_state.comm.new_conn_cb(ch->serial);
+		} else {
+			pr_warning(LOG_PFX "No incomming server callback "
```
brentpicasso: spelling. should be: incoming

src/drivers/esp8266_drv.c
```diff
@@ -1310,16 +1290,15 @@ static void connect_cb(const bool status, const bool already_connected)
...
 
-        ch->connected = status;
-
-        if (!status && already_connected) {
-                /* Attempt to close the connection to rectify state */
-                pr_info(LOG_PFX "Closing unexpectedly open connection\r\n");
-                esp8266_close(cso->chan_id, NULL);
-        }
+        if (already_connected) {
+		pr_info(LOG_PFX "State Mismatch. Channel already connected\r\n");
```
brentpicasso: Nit. camel caps on State Mismatch. s/b State mismatch.

src/drivers/esp8266_drv.c
```diff
@@ -1310,16 +1290,15 @@ static void connect_cb(const bool status, const bool already_connected)
...
+                channel_close(ch);
+        } else if (!status) {
+		pr_info(LOG_PFX "Channel failed to connect\r\n");
```
brentpicasso: should indicate what channel ID failed to connect maybe, for context?

src/drivers/esp8266_drv.c
```diff
@@ -1425,16 +1405,19 @@ bool esp8266_drv_close(struct Serial* serial)
         xSemaphoreTake(cso->op_semaphore, portMAX_DELAY);
         bool status = false;
 
-        cso->chan_id = find_channel_with_serial(serial);
-        if (0 > cso->chan_id) {
+        cso->chan_id = channel_find_serial(serial);
+        if (cso->chan_id < 0) {
                 pr_warning(LOG_PFX "Unable to find channel with associated "
                            "Serial device.\r\n");
```
brentpicasso: capitalization.  s/b serial device

src/tasks/wifi.c
```diff
@@ -259,15 +271,74 @@ static void process_rx_msgs(struct Serial* s)
...
+		 * b/c this is managing task
+		 */
+		pr_info_int_msg(LOG_PFX "New external connection: ", i);
+		serial_set_ioctl_cb(s, wifi_serial_ioctl);
+		state.connections[i] = s;
+		return;
+	}
+
+	/* If here, no slot found.  Close the Serial port */
+	pr_warning(LOG_PFX "Max external connections reached. Closing\r\n");
```
brentpicasso: is this the meat of the change?  I'm trying to quickly find what I should be looking at amid the other refactoring noise.

src/tasks/wifi.c
```diff
@@ -259,15 +271,74 @@ static void process_rx_msgs(struct Serial* s)
...
+static void check_connections()
+{
+	bool msg_handled = false;
+	for (int i = 0; i < ARRAY_LEN(state.connections); ++i) {
+		struct Serial* serial = state.connections[i];
+		if (!serial)
+			continue;
+
+		/* Check if serial is closed.  If so, handle it */
+		if (!serial_is_connected(serial)) {
```
brentpicasso: Rather, I meant maybe if this is the meat of the change.

src/drivers/esp8266_drv.c
stieg: Wise.  Added.

src/tasks/wifi.c
stieg: The meat of the change is as follows:

* Serial devices can now close.
* All Serial read/write ops return number of characters read/written, or -1 as an error code.
* All read/write ops were evaluated and refactored to ensure their return codes were interpreted correctly
* Esp8266_driver is only responsible for creating Serial objects when a new connection arrives or is requested or closing serial objects when a connection disappears or is closed internally.
* WiFi task is now responsible for managing sockets and reaping their resources when they close.

brentpicasso:
Looks good overall.  I say :shipit: and we'll tease out any issues during extended testing.

Issue #778, opened 2016-07-29
Don't enable WiFi by default on MK2


stieg:
Because customers might not have it or may be using the aux port for something else. But it needs to be enabled by default on RCT since that is a critical path for WiFi there.

PR #779, opened 2016-07-29
Disable Wifi by default on MK2, enable by default everywhere else

stieg:
Because WiFi isn't a native part of MK2 like it is on some other
devices.  Hence we don't want it on unless the customer says so.

Issue #778


stieg:
I'm merging this because its trivial

Issue #780, opened 2016-07-31
Reduce # of Tracks Stored to work Around Track DB CRUD Issue


stieg:
To work around the issue described in #536 until proper fix is in place, we should reduce the total # of tracks that can be stored so that the memory footprint is small enough that malloc failure doesn't happen.

PR #781, opened 2016-07-31
Reduce Track DB size to 2/3 of original size

stieg:
Because we were occationally hitting malloc issues while trying to
get a memory block large enough to hold the entire track DB, I am
reducing its size here because:

* It is more probable we can allocate the memory block we need
* It is likely to not be noticed since most people don't put
  240 tracks into their DB (wonder how many even hit 100).

With this change we go from needing 40K to only needing just over
25K.  Much more managable.

Issue #780 #536


brentpicasso:
:shipit:

PR #788, opened 2016-08-02
Fix Duplicate Telemetry Streams Bug

stieg:
Turns out that my optimization caused a regression.  If a user was to try to adjust the rate of a stream that currently existed, then we would simply create a new one, causing a duplicate stream to appear and also causing the handle to the original stream to be lost.  The end result was the stream failing to close, which would cause the unit to hang :(.

This fixes that bug.  It also adds strong typing to serial ioctl return values so I know the meaning of the code with certainty.

### DIFF
--- a/src/tasks/wifi.c
+++ b/src/tasks/wifi.c
@@ -98,6 +98,15 @@ static struct connection* find_connection(struct Serial* serial)
 	return NULL;
 }
 
+/**
+ * Resets a given connection back to a sane initial state.
+ */
+static void reset_connection(struct connection* c)
+{
+	c->serial = NULL;
+	c->ls_handle = -1;
+}
+
 /* *** All methods that are used to generate events *** */
 
 /**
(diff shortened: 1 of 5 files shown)


src/tasks/wifi.c
```diff
@@ -191,42 +200,49 @@ static void wifi_sample_cb(const struct sample* sample,
...
 
-	/* Stop the stream if the rate is <= 0 */
-	if (rate <= 0) {
+	if (rate <= 0 || conn->ls_handle >= 0) {
 		pr_info_str_msg(LOG_PFX "Stopping telem stream on ",
 				serial_name);
 
 		if (!logger_sample_destroy_callback(conn->ls_handle))
-			return -2;
+			return SERIAL_IOCTL_STATUS_ERR;
```
brentpicasso: yay, strongly typed codes! :)

brentpicasso:
Good catch, and looks sane. :shipit:

stieg:
Thank you for the quick turnaround @brentpicasso

Issue #793, opened 2016-08-10
Fix outdated samples at beginning of Logfile


stieg:
When you start logging on 2.10.0 often you will see the first 10 samples be out of order.  Here is an example:

```
"Interval"|"ms"|0|0|1,"Utc"|"ms"|0|0|1,"Battery"|"Volts"|0.0|20.0|1,"AccelX"|"G"|-3.0|3.0|100,"AccelY"|"G"|-3.0|3.0|100,"AccelZ"|"G"|-3.0|3.0|100,"Yaw"|"Deg/Sec"|-120.0|120.0|100,"Pitch"|"Deg/Sec"|-120.0|120.0|100,"Roll"|"Deg/Sec"|-120.0|120.0|100,"RPM"|"RPM"|0|9000|25,"EcuVolts"|"volts"|0.0|20.0|10,"IAT"|"C"|0|100|1,"EngineTemp"|"C"|0|150|1,"TPS"|"%"|0|100|10,"MAP"|"PSI"|0.0|15.0|10,"InjectorPW"|"ms"|0.0|100.0|10,"Ignition"|"D"|-20|20|10,"AFR"|"lambda"|0.0|2.0|10,"Knock"|"count"|0|15|1,"CamTiming"|"D"|-20|20|10,"Latitude"|"Degrees"|-180.0|180.0|10,"Longitude"|"Degrees"|-180.0|180.0|10,"Speed"|"MPH"|0.0|150.0|10,"Distance"|"Miles"|0.0|0.0|10,"Altitude"|"Feet"|0.0|4000.0|10,"GPSSats"|""|0|20|10,"GPSQual"|""|0|5|10,"GPSDOP"|""|0.0|20.0|10,"LapCount"|""|0|0|10,"LapTime"|"Min"|0.0|0.0|10,"Sector"|""|0|0|10,"SectorTime"|"Min"|0.0|0.0|10,"PredTime"|"Min"|0.0|0.0|10,"ElapsedTime"|"Min"|0.0|0.0|10,"CurrentLap"|""|0|0|10
190163,0,,0.01,-0.0,0.97,-1.39,-0.37,-1.27,0,,,,,,,,,,,,,,,,,,,,,,,,,
[...]

PR #794, opened 2016-08-10
P/stieg/sd logging improvements

stieg:
Sorry for the tremendously long review... but I promise its totally worth it.

This PR fixes 4/5 SD writing issues, with the one remaining issue being benign when it shows up (tracked in #793).  It does this by making the following changes:

* Upgrading our FatFS library to version 0.11
* Upgrading our underlying SDIO library to one provided by http://stm32f4-discovery.net/2014/07/library-21-read-sd-card-fatfs-stm32f4xx-devices/
* Adding in additional checks to ensure that tick values on samples match
* Adding in more proactive measures to aggressively set tick values when new sample buffers are populated.

All of this is done without losing any of our previous functionality.  And the results are quite good so far.  The main issues of corrupted writes and empty lines seem to have been eliminated (tested on 3 different SD cards).  Issue #793 does still show up occasionally, but it is benign for the most part since its data is still technically valid.


src/logger/loggerTaskEx.c
```diff
@@ -232,7 +232,7 @@ void loggerTaskEx(void *params)
                 }
 
                 /* Prepare a Sample */
-                struct sample *sample = &g_sample_buffer[bufferIndex];
+                struct sample *sample = g_sample_buffer + bufferIndex;
```
brentpicasso: what are your thoughts on this being better in terms of readability? I tend to think the array reference is clearer to understand.

brentpicasso:
:shipit:  looks great

src/logger/loggerTaskEx.c
stieg: We can use either honestly.  C pointer arithmetic does get a little odd looking I agree, I just don't personally like the dereference to reference operations together.  I'll make a point not to change these in future patches if they are written that way.  Good?

Issue #514, opened 2016-04-12
Add better unit testing on the version and status API calls


stieg:
In the 2.9.0 RCs, a bug was introduced where `bugfix` was mistakenly labeled as `bufix`.  This caused the app to break in a subtle way, but it was noticed and corrected before GA release.  This should have been caught by our unit tests, but was not for some reason.  Need to figure out why and plug this hole in testing.

stieg:
Merged

PR #796, opened 2016-08-10
Improve the version and status API tests

stieg:
Improves these tests so that we validate all fields are present
and contain expected values.  This also cleans up the code so
that it is more readable and easier to parse.

Issue #514


brentpicasso:
:shipit:

Issue #800, opened 2016-08-11
RPM2 and RPM3 channels have no default name defined


brentpicasso:
observed at the top of 2.10.0 with a fresh resetConfig

![image](https://cloud.githubusercontent.com/assets/3267533/17599125/7a3c4ea6-5fb2-11e6-881e-33a3ce057018.png)

RPM channel defaults should be named:  RPM, RPM2, RPM3

PR #801, opened 2016-08-11
Fix the default label on the timer channels.

stieg:
This adjusts the label so that these channels are always labeled with
"RPM", "RPM2", "RPM3" respectively.  They also have default units set
for all channels.

Issue #800

### DIFF
--- a/src/timer/timer_config.c
+++ b/src/timer/timer_config.c
@@ -21,11 +21,12 @@
 
 #include "api.h"
 #include "loggerConfig.h"
+#include "macros.h"
 #include "timer_config.h"
 #include <stdlib.h>
 #include <string.h>
+#include <stdio.h>
 
-#define RPM_CHAN_LABEL	"RPM"
 #define RPM_CHAN_UNITS	"rpm"
 #define RPM_CHAN_MIN	0
 #define RPM_CHAN_MAX	8000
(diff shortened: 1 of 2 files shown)


src/timer/timer_config.c
```diff
@@ -50,23 +51,24 @@ TimerConfig* get_timer_config(int channel)
...
-                TimerConfig *tc = tim_cfg + i;
-                set_default_channel_config(&tc->cfg);
-                tc->filterAlpha = TIMER_DEFAULT_FILTER_ALPHA;
-                tc->mode = MODE_LOGGING_TIMER_RPM;
-                tc->pulsePerRevolution = TIMER_DEFAULT_PULSE_PER_REV;
-                tc->timerSpeed = TIMER_MEDIUM;
-                tc->filter_period_us = TIMER_DEFAULT_FILTER;
-                tc->edge = TIMER_DEFAULT_EDGE;
-        }
+	for (int i = 0; i < cnt; ++i) {
```
brentpicasso: should use size_t here for array index

brentpicasso:
only one comment here. Looks good otherwise

src/timer/timer_config.c
stieg: Migration to an int type was intentional because of the use of the `snprintf` command.  This way I avoid a cast.  Perhaps you want me to use `unsigned int` instead.  That is really the only reason you want me to use `size_t` right?

brentpicasso: Pretty much the standard reason to use size_t based on conventional wisdom. unsigned int works too.

brentpicasso:
:shipit:

Issue #803, opened 2016-08-12
Improve WiFi Task Managed Connections to be interrupt driven


stieg:
Currently we use a polling mechanism to handle new data on these connections.  While this works it is most certainly not ideal since it chews up extra CPU resources and can cause our system to flood itself with events unless specifically coded to handle that scenario.  An event based solution would be much more elegant.  To do this we need to fix our Serial Device to not destroy its resources upon close, and instead do this upon destroy.  This is so we use the FreeRTOS Queue Set API Functions to wait on incoming data.  In its current state a serial close invocation would cause a SEGFAULT because it would delete a queue before the queue was removed from the queue set.

Issue #802, opened 2016-08-12
Fix WiFi send buff len calculation


stieg:
Every now and then when we are sending data across the WiFi TCP stream I see NULL characters that mysteriously appear.  Here is one such example:

```
[DEBUG             ] [SocketConnection] returning data {"s":{"t":98360,"d":[98362,0,0.13,0.1,0.87,-0.81,-3.59,2.11,507]}}
[DEBUG             ] [SocketConnection] returning data {"s":{"t":98400,"d":[98402,0,0.13,0.1,0.86,-0.76,-3.57,1.93,0.0,0.0,0.0,0.0,0.0,0,0,0.0,0,0.0,-1,0.0,0.0,0.0,0,16777211]}}
[DEBUG             ] [SocketConnection] returning data {"s":{"t":98440,"d":[98442,0,0.13,0.1,0.86,-0.82,-3.27,1.89,507]}}
[DEBUG             ] [SocketConnection] returning data {"s":{"t":98480,"d":[98482,0,0.13,0.1,0.87,-0.73,-2.92,1.56,507]}}
[DEBUG             ] [SocketConnection] returning data {"s":{"t":98500,"d":[98502,0,0.0,0.0,0.0,0.0,0.0,0,0,0.0,0,0.0,-1,0.0,0.0,0,14679555]}}
[WARNING           ] [RCPAPI      ] Message rx worker exception: u'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00{"s":{"t":98480,"d":[98482,0,0.13,0.1,0.87,-0.73,-2.92,1.56,507]}}' | No JSON object could be decoded
[DEBUG             ] [Traceback (most recent call last)]
  File "/home/stieg/Devel/asl/app/autosportlabs/racecapture/api/rcpapi.py", line 200, in msg_rx_worker
    msgJson = json.loads(msg, strict=False)
  File "/usr/lib64/python2.7/json/__init__.py", line 352, in loads
    return cls(encoding=encoding, **kw).decode(s)
  File "/usr/lib64/python2.7/json/decoder.py", line 364, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
[...]

PR #804, opened 2016-08-12
Some Wifi Connectivity Improvements

stieg:
These small tweaks help messaging become more reliable and prevent interrupt floods from overwhelming the wifi task when under load.  Ideally we would use event based wakeups but that is a large task as described in #803.  Things work a bit better now but there are still issues with the App and our unit including issue #802 which seems to cause some bad data to be sent and thus causes problems with the driver.


src/drivers/esp8266_drv.c
```diff
@@ -405,55 +417,63 @@ static void rx_data_cb(int chan_id, size_t len, const char* data)
 /**
  * Callback that is invoked when the send_data method completes.
  */
-static void _send_data_cb(int bytes_sent)
+static void _send_data_cb(const bool status, const size_t bytes,
+			  const int chan)
```
brentpicasso: recommend size_t for chan

src/drivers/esp8266_drv.c
```diff
@@ -269,29 +268,41 @@ static const char* channel_get_name(const size_t chan_id)
...
-static struct channel* channel_setup(const unsigned int i)
+/**
+ * Sets up a channel by allocating a channel and then creating a Serial
+ * device that backs it.
+ * @param index Channel ID/index.
+ * @param rx_size Size of the Rx Buffer
+ * @param tx_size Size of the Tx Buffer
+ * @param pointer to the channel structure that was setup, NULL if failure.
+ */
+static struct channel* channel_setup(const unsigned int index,
```
brentpicasso: recommend size_t for index

src/drivers/esp8266_drv.c
```diff
@@ -1309,11 +1329,15 @@ static void connect_cb(const bool status, const bool already_connected)
  * @param proto The protocol to use.
  * @param addr The destination address (either IP or DNS name).
  * @param port The destination port to connect to.
+ * @param rx_size The size of the Rx buffer. 0 will yeild the default value.
```
brentpicasso: spelling. s/b yield. ditto below

src/serial/rx_buff.c
```diff
@@ -99,7 +99,10 @@ void rx_buff_destroy(struct rx_buff *rxb)
 bool rx_buff_read(struct rx_buff *rxb, struct Serial *s, const bool echo)
 {
         xQueueHandle h = serial_get_rx_queue(s);
-        char c = 1;
+	if (!h)
+		return false;
+
+        char c = 0xEF;
```
brentpicasso: what is 0xEF ?

brentpicasso:
Looks solid; I tested it and wifi socket recovered. Only a few minor suggestions.

src/serial/rx_buff.c
stieg: Invalid UTF8 Unicode character.  Yields the symbol with the question mark in it on most terminals

brentpicasso: cool. I would just put a comment above it for it's purpose :+1:

src/drivers/esp8266_drv.c
stieg: Reasoning? http://stackoverflow.com/questions/1951519/when-should-i-use-stdsize-t

src/serial/rx_buff.c
stieg: Crap... I wrote that backwards.  Should be 0xFE.  I'll just make it 0xFF since that is also invalid.

src/drivers/esp8266_drv.c
brentpicasso: mostly b/c it represents an index of an item, so partially referenced by that article.  Also, you use size_t elsewhere in the code to reference the channel ID, so also to be consistent with that.

src/serial/rx_buff.c
stieg: For reference: https://en.wikipedia.org/wiki/UTF-8#Invalid_byte_sequences

src/drivers/esp8266_drv.c
brentpicasso: unsigned int would work too, consistent with the discussion in the previous PR.

PR #813, opened 2016-08-16
Fix Serial Closing and Close Socket on Send Fail

stieg:
Fixes a critical bug that could cause threads to die on our firmware.  Also fixes bad send ops by terminating the connection (because we lose data).

### DIFF
--- a/src/serial/serial.c
+++ b/src/serial/serial.c
@@ -29,12 +29,15 @@
 #include "projdefs.h"
 #include "serial.h"
 #include "str_util.h"
+#include "queue.h"
 #include "usart.h"
 #include "usb_comm.h"
 #include <stdarg.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <string.h>
+
+static const char invalid_char = INVALID_CHAR;
 
 enum data_dir {
         DATA_DIR_RX,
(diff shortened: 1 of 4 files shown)


src/serial/serial.c
```diff
@@ -59,27 +63,52 @@ struct Serial {
...
+
+/**
+ * Unblocks our Seral rx queue by putting an invalid character on the
+ * front of the queue. This will unblock any task that is waiting on
+ * the queue.  Its up to the rx handler below to ensure that we return
+ * the correct status code.
+ */
+static void unblock_rx_queue(struct Serial *s)
+{
+	xQueueSendToFront(s->rx_queue, &invalid_char, 0);
```
brentpicasso: So we're using the invalid char as a flag  that says "close this queue"? Seems legit.

brentpicasso:
looks good to me. :shipit:

src/serial/serial.c
stieg: Not quite.  I do this to unblock any tasks that are blocked by this queue.  The only way they can read data from this queue is the get_c method below.  In that method I check if the serial device is closed and if it saw the invalid character.  If that happens, then I return an error status indicating that the Serial device is closed instead of returning this invalid character.  I then also re-call this very method there to unblock any other tasks that may be blocked.  This ensures that no task remains blocked (no matter how many are waiting).

PR #816, opened 2016-08-17
change gyro precision to 0 for usability / clarity

### LINKED ISSUE #815: Make precision on gyro channels (yaw / pitch / roll ) zero position
since we're dealing with hundreds of degrees of rotation / sec, having two decimal points of precision seems excessive. Should represent the value in whole numbers.

brentpicasso:
resolves #815


stieg:
:shipit:

stieg:
Merging because trivial

Issue #814, opened 2016-08-16
Track config for devices with no track db have autodetect set to 1


ryandoherty:
RaceCapture with firmware 2.10.0 reports back that autodetect is on:

`{"trackCfg":{"rad":0.0001,"autoDetect":1,"track":{"id":0,"type":0,"sf":[0.0,0.0],"sec":[[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0],[0.0,0.0]]}}}`

It should probably be off since it has no track db.

brentpicasso:
So, the auto detect flag should still be there even if there is no track
db. This is the indicator that the app should select the track
automatically.

ryandoherty:
The bug is that autodetect is set to 1, not that the key exists.

stieg:
Correct.  Need to adjust the default so that if the number of DB entries available is 0, then the auto-detect value is 0.

stieg:
Merged

PR #817, opened 2016-08-17
Disable track auto-detect by default if trackDB size is 0

stieg:
If we have no track DB, then auto-detect is useless.  So we should
disable it by default in that case. Otherwise we should keep the
traditional behavior of enabling it by default if we do have a
track DB available (whether or not it has tracks is another story).

Issue #814


brentpicasso:
:ship:

PR #818, opened 2016-08-17
Try and setup the track, even without GPS

stieg:
To improve the UX we should try and set the track config, even when we
don't have GPS. This works in cases where the user has manually
specified what track they want to use. In the case of automatic track
selection, then we need to make use of the GPS position and thus will
wait for GPS lock before this action happens.


brentpicasso:
:shipit:

PR #819, opened 2016-08-18
fix unit tests

brentpicasso:
change gyro precision to 0, including fixed unit tests


stieg:
Since I reverted your change you should revert that revert and include it as part of this patch.  Once you do that and build it locally before you merge, :shipit:

Issue #775, opened 2016-07-29
If WiFi Enabled but WiFi Device Not Present, WiFi Init Hangs


stieg:
This was the root cause of issue #753 and is getting a workaround as tracked by issue #774.  However the main point still remains: WiFi gets stuck during init and is never able to leave.

Culprit seems to be down in the Serial level in that the `at_basic_wait_for_msg` command never returns.  Here is a stack:

```
Breakpoint 6, at_basic_wait_for_msg (serial=serial@entry=0x200046b8, msg=msg@entry=0x8069db6 "OK", delay_ms=delay_ms@entry=200) at ../../src/modem/at_basic.c:41
41      {
(gdb) bt
#0  at_basic_wait_for_msg (serial=serial@entry=0x200046b8, msg=msg@entry=0x8069db6 "OK", delay_ms=delay_ms@entry=200) at ../../src/modem/at_basic.c:41
#1  0x08033596 in at_basic_ping (serial=serial@entry=0x200046b8, tries=tries@entry=3, delay_ms=delay_ms@entry=200) at ../../src/modem/at_basic.c:69
#2  0x080335e0 in at_basic_probe (serial=0x200046b8, serial@entry=0x38400, bauds=bauds@entry=0x20002498, size=size@entry=2, tries=tries@entry=3, delay_ms=delay_ms@entry=200, msg_bits=msg_bits@entry=8, parity=pa\
rity@entry=0, stop_bits=stop_bits@entry=1) at ../../src/modem/at_basic.c:96
#3  0x0802912c in esp8266_probe_device (serial=serial@entry=0x38400, fast_baud=fast_baud@entry=230400) at ../../src/devices/esp8266.c:1111
#4  0x0802a1f6 in esp8266_drv_init (s=0x38400, priority=priority@entry=3, new_conn_cb=new_conn_cb@entry=0x8034b55 <_new_conn_cb>) at ../../src/drivers/esp8266_drv.c:1254
#5  0x08034bc2 in wifi_init_task (wifi_task_priority=wifi_task_priority@entry=2, wifi_drv_priority=wifi_drv_priority@entry=3) at ../../src/tasks/wifi.c:404
#6  0x080201d2 in setupTask (param=<optimized out>) at ../../main.c:104
#7  0x080393d8 in vPortEnterCritical () at libs/FreeRTOSV7.6.0/FreeRTOS/Source/portable/GCC/ARM_CM4F/port.c:390
[...]

Issue #587, opened 2016-05-09
Add support for "Rude AT" devices


stieg:
So from field testing it would seem that the esp8266 is not very well mannered when it comes to commands and URCs.  Specifically it seems that it will just throw URCs back at my state machine **MID COMMAND**.  This is not very good since the AT command engine was designed on the principle that commands would be free of URCs while a command was in progress.  This is how sane engines like those on U-Blox devices work; they queue URCs and wait for commands to finish before releasing the URC hounds.  This ticket tracks trying to work in a way in which we might better support this interrupting URC system.  I'm calling it "Rude AT support" since the URCs are interrupting the middle of a command.

stieg:
Putting this back into 2.10.0.  Without this we miss many instructions and URCs, and that causes plenty of issues (slow writes, missed connection close events).  Here is one such example were missing the socket close URC causes all sorts of hell:

```
[serial (WiFi)] TX: {"s":{"t":93940,"d":[258495,0,0.0,0.0,0.0,0.0,0.0,0,0,0.0,130563]}}(\r)(\n)
[serial (WiFi)] RX: > (\r)(\n)
[serial (WiFi)] RX: Recv 69 bytes(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: SEND OK(\r)(\n)
[serial (WiFi)] TX: AT+CIPSEND=1,98(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: OK(\r)(\n)
[serial (WiFi)] TX: {"s":{"t":93960,"d":[258515,0,0.22,0.11,0.86,-0.71,-2.69,1.87,0.0,0.0,0.0,0.0,0.0,0,0,0.0,131067]}
[serial (WiFi)] RX: > (\r)(\n)
[serial (WiFi)] RX: Recv 98 bytes(\r)(\n)
[serial (WiFi)] RX: 1,CLOSED(\r)(\n)
[serial (WiFi)] RX: SEND FAIL(\r)(\n)[esp8266] send_data_cb failed with msg: Bad response value
[ESP8266 Driver] Failed to send data on channel 1
[wifi] Event overflow: Sample CB
[wifi] Event overflow: Sample CB
[wifi] Event overflow: Sample CB
[wifi] Event overflow: Sample CB
[wifi] Event overflow: Sample CB
[wifi] Event overflow: Sample CB
[...]

PR #824, opened 2016-08-19
Rude AT Support And More

stieg:
We have all heard my grumblings about how the ESP8266 chip is "rude" in that it sends its events during the middle of a normal command (how rude indeed).  This PR includes the patches needed to be able to deal with that.

In the process of adding this support, I hit several other issues, including fixing a hack to work around a different esp8266 issue (no serial EOM message after baud rate change) and had to deal some fallout of my design change.

Regardless the result is almost no dropped incoming messages now, so the app responds much more quickly.

### DIFF
--- a/src/modem/at.c
+++ b/src/modem/at.c
@@ -229,8 +229,8 @@ static void process_cmd_or_urc_msg(struct at_info *ati, char *msg)
                 return process_urc_msg(ati, msg);
         }
 
-	if (ati->unhandled_urc_cb && ati->unhandled_urc_cb(msg)) {
-		/* It was a generic URC that was handled. */
+	if (ati->sparse_urc_cb && ati->sparse_urc_cb(msg)) {
+		/* It was a sparse URC that was handled. */
 		return;
 	}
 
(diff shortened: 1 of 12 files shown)


stieg:
Addresses #587 and may address #775

src/modem/at.c
```diff
@@ -494,18 +505,20 @@ struct at_urc* at_register_urc(struct at_info *ati, const char *pfx,
...
  * @param delim The command delimeter characters.  Normally this is "\r\n".
+ * @param flags One or more at_dev_cfg_flag items or'd together.
  * @return true if the parameters are acceptable, false otherwise.
  */
 bool at_configure_device(struct at_info *ati, const tiny_millis_t qp_ms,
-                         const char *delim)
+                         const char *delim, const enum at_dev_cfg_flag flags)
 {
         if (!delim || strlen(delim) >= AT_DEV_CVG_DELIM_MAX_LEN) {
-                pr_error("[at] Failed to set delimeter\r\b");
```
brentpicasso: whoops! fail on previous code review :)

brentpicasso:
Big change at a late hour but the reliability / performance is significant after direct testing. Code looks clean. 

I think a write up for all of the ESP8266 inconsistencies would be in order for esspressif. :)

PR #826, opened 2016-08-19
A Few Important Misc Fixes for RCT

stieg:
The following fixes are in this PR:

* Fix ADC calibration crash:  Turns out we need to disable ADC before calibrating it.  Else it is liable to destabilize the system.
* Improve handling of no serial EOF on IPD command
* Use strncpy instead of memcpy on our config to avoid reading data from potentially out of bounds memory regions
* Demote a couple of noisy log messages


brentpicasso:
:shipit:

PR #828, opened 2016-08-21
Reduce MK2 Wifi Max Sample rate down to 10Hz

stieg:
We have 1/4 the bandwidth for wifi on this device when compared to
RCT.  This is due to hardware limitations.  As a result we need to
set stricter limits for this device to ensure it remains operational
under load.

This reduces WiFi sample baud rate to 10Hz.  After doing this our
reliability over WiFi was much improved.


brentpicasso:
:shipit:

Issue #829, opened 2016-08-22
Status API - need basic wifi init status


brentpicasso:
We have a status API for WiFi, but it is missing a clear indicator of the WiFi module initialization state - as in, are we able to communicate and perform basic initialization. Similar to what we have for IMU.

This is needed for production testing and field diagnosis.

Issue #821, opened 2016-08-18
RCT should not report Bluetooth and Logger status


brentpicasso:
Since RCT does not have these devices, status on these shouldn't be reported.

stieg:
So if I don't report these will the app crash or just assume them to be `null`?

ryandoherty:
The app uses the status dict from the getStatus response to determine which menu items and values to render, so we're good.

stieg:
:+1:

PR #832, opened 2016-08-23
Improve Status API Reporting

stieg:
Adds additional WiFi status parameters ("active" and "initialized") and removes BT and Logging reporting when these are not available on the target system.  Tested on both RCT and MK2

Issues: #829 #821

### DIFF
--- a/src/logger/loggerApi.c
+++ b/src/logger/loggerApi.c
@@ -237,14 +237,20 @@ static void get_imu_status(struct Serial *serial, const bool more)
 
 static void get_wifi_status(struct Serial* serial, const bool more)
 {
+        const LoggerConfig *lc = getWorkingLoggerConfig();
+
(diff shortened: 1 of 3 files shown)


src/logger/loggerApi.c
```diff
@@ -237,14 +237,20 @@ static void get_imu_status(struct Serial *serial, const bool more)
 
 static void get_wifi_status(struct Serial* serial, const bool more)
 {
+        const LoggerConfig *lc = getWorkingLoggerConfig();
```
brentpicasso: I guess we don't need to #ifdef  wifi status as everything supports it currently. Legit.

brentpicasso:
:shipit:

src/logger/loggerApi.c
stieg: Yep.

Issue #822, opened 2016-08-18
WiFi unusable if set to connect to non-existent client network.


stieg:
If you set your RCT to connect to a client WiFi network, it will do so well if the proper network settings are in place and the network is in range.  However failing one of the two above will put the wifi driver into a loop it will not recover from until it gets within range of the network or gets its configuration updated.  We need to be able to handle this more gracefully in firmware.

stieg:
Merged

PR #834, opened 2016-08-25
Adjust WiFi client to better handle connect/disconnect actions

stieg:
The behavior of the WiFi client on the ESP8266 was not entirely
clear and thus I ended up coding it incorrectly.  This change addresses
that so that we use the auto-connect features that are built into the
ESP8266 chip and better handle the callbacks that we get from it.

Issue #822

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -133,15 +133,19 @@ static void wifi_action_callback(const char* msg)
 	/* Look at our 5th character to determin our action */
 	switch (msg[5]) {
 	case 'C':
+		/* WIFI CONNECT */
 		action = CLIENT_ACTION_CONNECT;
 		break;
 	case 'D':
+		/* WIFI DISCONNECT */
 		action = CLIENT_ACTION_DISCONNECT;
 		break;
 	case 'G':
+		/* WIFI GOT IP */
 		action = CLIENT_ACTION_GOT_IP;
 		break;
 	default:
+		pr_warning_str_msg(LOG_PFX "Unknown Wifi Action: ", msg);
 		action = CLIENT_ACTION_UNKNOWN;
 		break;
 	}
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -123,6 +123,32 @@ static void ipd_urc_cb(char *msg)
 		state.hooks.data_received_cb(chan_id, len, data);
 }
 
+static void wifi_action_callback(const char* msg)
+{
+	/* Check for the hook */
+	if (!state.hooks.client_state_changed_cb)
+		return;
+
+	enum client_action action;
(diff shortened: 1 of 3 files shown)


src/devices/esp8266.c
```diff
@@ -123,6 +123,32 @@ static void ipd_urc_cb(char *msg)
...
+static void wifi_action_callback(const char* msg)
+{
+	/* Check for the hook */
+	if (!state.hooks.client_state_changed_cb)
+		return;
+
+	enum client_action action;
+	/* Look at our 5th character to determin our action */
+	switch (msg[5]) {
+	case 'C':
```
brentpicasso: that's clever! always  unique?

brentpicasso: Nit suggestion: would be cool to include, above each switch statement a comment with the actual message from the ESP8266 that we're capturing. for reference.

src/devices/esp8266.c
```diff
@@ -123,6 +123,32 @@ static void ipd_urc_cb(char *msg)
...
 
+static void wifi_action_callback(const char* msg)
+{
+	/* Check for the hook */
+	if (!state.hooks.client_state_changed_cb)
+		return;
+
+	enum client_action action;
+	/* Look at our 5th character to determin our action */
+	switch (msg[5]) {
```
brentpicasso: what if our message is too short, by accident?

brentpicasso:
A few comments, looks good otherwise

src/devices/esp8266.c
stieg: If it wasn't unique then I could not use it :D.  Always unique from all replies that I have seen.  Will be happy to add the comments you requested.

src/devices/esp8266.c
stieg: Then the use of this method is being done incorrectly.  I only bounds check public methods or methods that are callbacks.  In this case since this is internal I do not bounds check.

stieg:
Merging since commit is trivial.

src/devices/esp8266.c
brentpicasso: A bit pedantic but worth pointing out; the strncmp() used in the function that calls this function only checks for the string "WIFI " - it doesn't check if the string is at least 6 characters to check for the start of that next word. 

if (strncmp(msg, "WIFI ", 5) == 0) {
 +		wifi_action_callback(msg);
 +		return true;
 +	}

What would happen is it would land on the null terminator - technically safe but a more explicit check would be to check for minimum length too. 

Too bad strncmp() doesn't provide wildcards! :)

stieg: It does do 6 characters though because you are not counting the null byte.
That is implicit in that it must be there.  Hence why this is safe

PR #835, opened 2016-08-25
GPS Improvements

stieg:
Improves GPS startup time and helps reduce the GPS sync issues seen after system resets.  Tested on both MK2 and RCT.


brentpicasso:
:shipit:

Issue #584, opened 2016-05-09
Add cummulitive AT error tracking and automatic reset of WiFi device


stieg:
If our WiFi somehow gets into a bad state, we need to be able to detect this and automatically reset the WiFi device.  This way we have a recovery path instead of just getting wedged indefinitely.

stieg:
Depends on #583

stieg:
Saw a good example of this after the chip ran for a while:

```
GPS: Successfully disabled NMEA messages
GPS: provisioned
[at] Command timed out
[esp8266] send_data_cb failed with msg: Bad response value
[ESP8266 Driver] Failed to send data
[at] Command timed out
[esp8266] send_data_cb failed with msg: Bad response value
[ESP8266 Driver] Failed to send data
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[at] Unhandled msg received: ��������������������������������������������������������������������������������������������
[at] Unhandled msg received: ���������������������������������������������������������������������������������������������������������������������������������������������������
[at] Unhandled msg received: �
[wifi] Event overflow: Beacon
[at] Unhandled msg received: ���������������
[wifi] Event overflow: Beacon
```
< Repeats a lot.  I enable ascii debug>
```
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[wifi] Event overflow: Beacon
[...]

PR #838, opened 2016-08-30
Reset WiFi if too many errors occur

stieg:
Adds an error tracking mechanism to the driver so that if too many
commands fail, then we reset the wifi hardware.  Currently this value
is hardcoded to 3 failures.

Issue #584

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -48,6 +48,7 @@
 #define AT_TASK_TIMEOUT_MS	3
 #define CHECK_DONE_SLEEP_MS	3600000
 #define CLIENT_BACKOFF_MS	3000
+#define CMD_MAX_FAILURES	3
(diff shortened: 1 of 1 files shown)


src/drivers/esp8266_drv.c
```diff
@@ -201,6 +191,34 @@ static void cmd_set_check(enum check check)
...
+	 * it is 0.  If we hit the threshold of failures, trigger
+	 * a reset.
+	 */
+	if (!success)
+		++esp8266_state.cmd.failures;
+	else if (esp8266_state.cmd.failures)
+		--esp8266_state.cmd.failures;
+
+	if (esp8266_state.cmd.failures >= CMD_MAX_FAILURES) {
+		pr_info(LOG_PFX "Too many failures.  Resetting...\r\n");
```
brentpicasso: maybe warning level instead?

brentpicasso:
only one small comment; otherwise :shipit:

PR #839, opened 2016-09-01
Fix Too Many Connections Crash; Increase Heap Size

stieg:
This PR gets our heap and extra 11K in RAM by moving the .bss segment into CCM and re-organizing some of our RAM layout.  It also fixes a regression caused by the Rude AT support and fixes a crash seen when too many connections were opened to the RCT device.

### DIFF
--- a/platform/rct/FreeRTOS/Source/portable/MemMang/heap_4.c
+++ b/platform/rct/FreeRTOS/Source/portable/MemMang/heap_4.c
@@ -94,7 +94,6 @@ task.h is included from an application file. */
 #define heapADJUSTED_HEAP_SIZE	( configTOTAL_HEAP_SIZE - portBYTE_ALIGNMENT )
 
 /* Allocate the memory for the heap. */
-extern size_t _CONFIG_HEAP_SIZE;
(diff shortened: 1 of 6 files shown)


platform/rct/FreeRTOS/Source/portable/MemMang/heap_4.c
```diff
@@ -94,7 +94,8 @@ task.h is included from an application file. */
 #define heapADJUSTED_HEAP_SIZE	( configTOTAL_HEAP_SIZE - portBYTE_ALIGNMENT )
 
 /* Allocate the memory for the heap. */
-static unsigned char ucHeap[ configTOTAL_HEAP_SIZE ];
+extern size_t _CONFIG_HEAP_SIZE;
```
stieg: Crap... forgot to delete this.  Will nuke.

src/drivers/esp8266_drv.c
```diff
@@ -1461,6 +1481,7 @@ bool esp8266_drv_close(struct Serial* serial)
         }
 
 done:
+	cso->chan_id = -1;
```
brentpicasso: extract -1 to a #define to explicitly indicate what it means

brentpicasso:
small comments; resolve and :shipit:

brentpicasso:
:shipit:

PR #842, opened 2016-09-02
Stop LUA if no script; fix log bug

stieg:
This PR will stop the LUA task if the user provides no script.  This just makes sense.  This also fixes a trivial logging bug.

### DIFF
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -21,9 +21,9 @@
 
 #include "FreeRTOS.h"
 #include "GPIO.h"
-#include "led.h"
 #include "capabilities.h"
 #include "lauxlib.h"
+#include "led.h"
 #include "lua.h"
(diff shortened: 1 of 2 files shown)


src/lua/luaTask.c
```diff
@@ -241,6 +242,15 @@ static void lua_task(void *params)
 
                 /* If its a known unrecoverable, fail fast */
                 switch (rc) {
+		case LUA_NO_PERIODIC_FUNCTION:
```
brentpicasso: I'm thinking this will still allow a one time script to still execute?  just confirming.

brentpicasso:
just one clarifying question; once resolved :shipit:

src/lua/luaTask.c
stieg: Yes.  This should do that as that happens when the script is loaded, not when the periodic function is run.

PR #843, opened 2016-09-02
Improves upon panic behavior

stieg:
Does this by clearing all LEDs before panic loop starts and then also adjusts the flash count to ensure its much more readable by a human.


brentpicasso:
:shipit:

PR #845, opened 2016-09-02
Increase stack sizes for button and wifi tasks

stieg:
Both were below the 32 value for their high water mark.  This adds
enough to ensure they are above that threshold.  I added more to the
wifi task because I once saw a stack smash that is probably related
to a couple of interupts firing and that lacking sufficient space.


brentpicasso:
:shipit:

Issue #848, opened 2016-09-05
various config writing fail via wifi


brentpicasso:
When connected over WiFi, various configs fail to write. Example:

Writing WiFi config
writing OBD2 config

problem does not repro over usb. Maybe stack smash?

stieg:
Well this is hard to trace down since I don't know if the device is failing or is returning a failed to accept error code (The app does ignore these replies).  If it was a stack smash you would see a panic since the debug builds check the stack sanity every context switch.

brentpicasso:
git describe :  2.10.0_RC2-28-g9342906 

Previously I had also observed it when also writing OBDII config, but now only writing WiFi config makes the problem happen consistently.
No response is observed when writing WiFi config or OBDII config. 

 app is launched as:   ./main.py -d -- --conn_type=wifi

STR: 
connect app via wifi
read config
modify wifi config
write wifi config
observe app displaying timeout message

example output from command line
[DEBUG  ] [SocketComm  ] writing message {"setWifiCfg":{"active":true,"ap":{"active":true,"encryption":"none","password":"","ssid":"RaceCapture 35239383620","channel":11},"client":{"active":true,"password":"xxxxxxxxx","ssid":"spam333"}}}
[DEBUG  ] [RCPAPI      ] Tx: {"getStatus":null}
[DEBUG  ] [SocketComm  ] writing message {"getStatus":null}
[ERROR  ] [SocketConnection] timeout
[WARNING] [RCPAPI      ] Read message timeout waiting for setWifiCfg
[WARNING] [RCPAPI      ] POKE
[DEBUG  ] [SocketComm  ] writing message

stieg:
Oh... I wonder if this is causing the WiFi connection to break the connection because its adjusting parameters.  Do you remember what you changed exactly?  For example did you change the client settings?  This would likely cause the WiFi to become unresponsive for a few seconds while it would update the client settings (Happens right after the config is updated).

stieg:
Also... thank you for the `git describe` output.  That helps a lot :smile:

brentpicasso:
:+1: 

I have RCT connected as a client to home wifi.  I went and disabled the AP (something unrelated), wrote config and verified the issue occurred.

stieg:
[...]

PR #850, opened 2016-09-05
Sleep X millis before applying a configuraiton update to the Wifi

stieg:
This is so that when a change happens our subsystem has enough time to
reply that the change was successful before acting upon a potentially
disruptive change.  Its the best solution I can come up with for issue 848.

Issue: #848

### DIFF
--- a/src/drivers/esp8266_drv.c
+++ b/src/drivers/esp8266_drv.c
@@ -43,9 +43,10 @@
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
-
 #define AP_BACKOFF_MS		30000
 #define AT_TASK_TIMEOUT_MS	3
+/* MS to wait before checking/updating wifi configs */
+#define CFG_UPDATE_SLEEP_MS	1000
 #define CHECK_DONE_SLEEP_MS	3600000
 #define CLIENT_BACKOFF_MS	3000
 #define CMD_MAX_FAILURES	3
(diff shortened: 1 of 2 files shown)


stieg:
Ok now this will check for any changes to the config before invoking the methods to update the WiFi settings.

brentpicasso:
looks like the delay is 1 second. Think that's enough in case an entire config is written and flashed? (e.g. restore from a saved file)

stieg:
Depends I suppose.  The user is literally changing WiFi settings while configuring over wifi.  That will always be an issue.  Think of it like adjusting network settings while you are ssh'd into a machine.

brentpicasso:
1 second is a long time, and it's a blind delay, but should be sufficient for now. I think it's worth testing this scenario - loading a saved config (which touches all config elements) that also updates wifi settings and then writing it back to unit.  If it seems to update reliably then I say merge it!

stieg:
I would say that the above scenario will not work (especially since the unit is fighting the flood of samples coming back).  Perhaps 3 - 5 seconds is key.  I'll play with it and come up with a solution.

stieg:
So I just timed this.  Saving an RCT config and then re-writing that config takes up to 10 seconds when its contending with the telemetry stream!  That is way too long to wait to apply new wireless settings.  IMHO the app should warn people about adjusting the WiFi settings when using the WiFi link.  I have done what I can reasonably do on the firmware side, but I'm open to suggestions.

stieg:
Filed https://github.com/autosportlabs/RaceCapture_App/issues/1202 to track the warning in the app.

brentpicasso:
We should measure responsiveness after https://github.com/autosportlabs/RaceCapture_App/issues/1190 is resolved.  The plan is to run the telemetry rate at 1Hz when not logging a session / in dashboard mode. This will improve responsiveness somewhat when saving.

stieg:
Sure we can do that, but any additional delay added is merely a bandaid at best.  It will never solve the problem; rather it will only hide it.  So this is why I question this approach beyond the solution I have provided in this patch.

PR #851, opened 2016-09-06
Increase some of our memory footprint now that we have more to work with

stieg:
Increases the following:

* Predictive timer buffers - more accuracte predictive timing
* Wifi connection Tx buffers - So we can send larger messages
* Wifi driver stack - To ensure we don't overflow and crash.


brentpicasso:
:shipit:

Issue #847, opened 2016-09-05
Firmware crash with regular use


brentpicasso:
top of 2.10.0, debug build. 

Firmware crashes with a 2 blink panic with regular use over wifi (RCT).  Things that will cause the crash: Connect, operate for a while, disconnect, re-connect, etc will trigger the crash. Will repro with single client

stieg:
@brentpicasso Please try and use the output of `git describe` instead of saying "top of 2.10.0".  While I understand your meaning, it is quite imprecise and often doesn't tell me enough.  In this case the panic flashes changed a bit recently, so I need to know if you are using the old style or the new style.

brentpicasso:
:+1: 

git describe reports:
2.10.0_RC2-28-g9342906

stieg:
Ok.  Here is how you begin to diagnose these issues:

```
enum panic_cause {
        PANIC_CAUSE_UNKNOWN = 1,
        PANIC_CAUSE_ASSERT,
        PANIC_CAUSE_OVERFLOW,
        PANIC_CAUSE_MALLOC,
        PANIC_CAUSE_UNREACHABLE,
        PANIC_CAUSE_TASK_CREATE,
        PANIC_CAUSE_HARDWARE,
        PANIC_CAUSE_SCHEDULER,
};
```

This is in panic.h and is what defines the meaning of the flashes.  In this case it would seem you hit some form of assert.  My guess is that this is a hard fault since that is what the hardfault handler reports back upon panic.

brentpicasso:
There's no specific pattern where this reproduces; it does happen fairly quickly though with just wifi connection, disconnects, re-connects, etc.

stieg:
Finally got it to reproduce:

```
#0  0x0801c448 in cpu_device_spin (ms=1397, ms@entry=5000) at hal/cpu_stm32/cpu_device_stm32.c:95
#1  0x0800e5f8 in panic (cause=cause@entry=PANIC_CAUSE_MALLOC) at ../../src/util/panic.c:47
#2  0x0800e650 in vApplicationMallocFailedHook () at ../../src/util/panic.c:81
#3  0x0800ee66 in pvPortMalloc (xWantedSize=<optimized out>, xWantedSize@entry=513) at FreeRTOS/Source/portable/MemMang/heap_4.c:244
[...]

Issue #542, opened 2016-04-29
Move from a Serial device model to a Socket or File Handle model in our API code


stieg:
Traditionally we passed around Serial handles to do all of our reading and writing.  And life was good and simple.  Really... it was a simpler time.  But now that we are fancy and are supporting proper sockets, things have become more complex.  Namely the issue we now face is that Serial devices don't know if they are connected or not.  Frankly, they don't care.  They are simple.  But sockets are more sophisticated and do in fact care if they are disconnected.  And internally we will need to start knowing this behavior so that we can do the right thing upon disconnect.  Thus this ticket exists to move our internal infrastructure to using Sockets instead of Serial devices.  The wise idea here being that if we do this properly, we can build sockets on top of the Serial device and have the traditional Serial layer be mostly unchanged.  But with the WiFi (and now 3G) device in play we need more sophisticated tools to work in a more sophisticated world.

stieg:
Bumping this up to the next major release because we can not currently cleanly destroy serial objects without risking segfaults.  We need some form of intermediary system (aka file handles) that reference a stream, but are not necessarily always backed by a physical device.  In this way can we sanely close and reap resources without fear of causing segfaults.  See Issue #847.

PR #853, opened 2016-09-07
Deprecate use of serial_destroy; replace with re-opening instead

stieg:
Turns out there are lots of race conditions associated with destroying
Serial devices.  Namely when tasks return from a suspended state they
may try to access a serial device that no longer exists.  This obviously
not good and was causing intermittent crashes on RCT.

This patch works around the issue by never destroying serial devices.
By doing this we will no longer have the segfault issues, but instead
will incur other issues like inproper buffer sizes.  I am willing to
live with that for this release.  Issue #542 tracks a proper fix
(moving to POSIX file handles and streams) which will put an intermediate
buffer inbetween these actions, allowing us to safely destroy the backends
without risk of segfault.

Issue #847

### DIFF
--- a/src/serial/serial.c
+++ b/src/serial/serial.c
@@ -107,7 +107,8 @@ void serial_close(struct Serial* s)
 /**
  * Re-opens a closed serial object.  This exists because there are inherent
  * race conditions when using serial objects directly and closing/destroying
- * them.  We need
+ * them.  We need an intermediary step like what glibc does with file handles
+ * and streams.  That is tracked by issue #542.
  */
 void serial_reopen(struct Serial* s)
 {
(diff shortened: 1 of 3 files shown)


src/serial/serial.c
```diff
@@ -104,6 +104,17 @@ void serial_close(struct Serial* s)
 	unblock_rx_queue(s);
 }
 
+/**
+ * Re-opens a closed serial object.  This exists because there are inherent
+ * race conditions when using serial objects directly and closing/destroying
+ * them.  We need
```
brentpicasso: complete the thought here? :)

brentpicasso:
Looks sane.  one tiny comment I made and :shipit:

src/serial/serial.c
stieg: Haha... perhaps that would be a good idea.

Issue #841, opened 2016-09-02
MK2 Does Not Come Up Cleanly After Flashing Firmware


stieg:
Seems to hang somewhere in the init process after a powercycle.  Should fix this.

stieg:
This is with 2.10.0_RC2-21-g94043c2 on MK2 with both BT and Wifi enabled.

PR #857, opened 2016-09-11
Fix wifi init hang; add better init logging

stieg:
We were encountering an overflow issue in our logic that would cause
the WiFi subsystem to hang and never recover during init if it was
enabled and no wifi device was attached. This patch fixes that and
adds lots more debug to help out with future issues.

I also added a lengh parameter instead of requiring a NULL terminated
list of bauds for the probe command.  I figured this was a saner
approach less prone to failure.

Issue #841

### DIFF
--- a/include/modem/at_basic.h
+++ b/include/modem/at_basic.h
@@ -37,9 +37,9 @@ bool at_basic_ping(struct Serial* serial, const size_t tries,
 		   const tiny_millis_t delay_ms);
 
 int at_basic_probe(struct Serial* serial, const int bauds[],
-		   const size_t tries, const tiny_millis_t delay,
-		   const size_t msg_bits, const size_t parity,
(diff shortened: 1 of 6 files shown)


brentpicasso:
Most of this looks really straightforward. :+1:  However, what part fixed the overflow issue? Hard to tell.

stieg:
So the part that really fixed the overflow was the conversion from the unsigned `size_t` type to the singed `tiny_millis_t` in the at_basic.c library (not in this changeset).  I was hitting a race condition there where my stop time was in the past relative to current time.  When you do <past Time> - <current Time> you get  a negative, which in the unsigned world rolls over to a very large positive.  That was the bug.  I did that conversion in a different changeset after I hit a similar bug.  This changeset has all the logging that now proves that this issue is now resolved along with the addition of the array length parameter.

Issue #837, opened 2016-08-30
GPS on RCT gets into re-init loop after a few seconds


stieg:
Seems that the GPS code gets into a re-init loop after a short amount of time.  Here is some logging:

```
*** Start! ***
Initializing CAN0 with baud rate 500000
CAN init success!
GPS: Initializing...
IMU: init res=0
Start OBD2 task
timebase/logging/telemetry sample rate: 50/50/50
GPS: provisioning attempt
GPS: probing baud rate: 921600
[ESP8266 Driver] Checking WiFi Device...
[ESP8266 Driver] Initializing Wifi
[esp8266] Version info:
        AT version:0.50.0.0(Sep 18 2015 20:55:38)
        SDK version:1.4.0
        compile time:Sep 18 2015 21:30:56
[esp8266] Initialized
[ESP8266 Driver] Initialization successful
[ESP8266 Driver] Checking WiFi Device...
[ESP8266 Driver] Increasing baud to 921600
[ESP8266 Driver] Checking WiFi Device...
[ESP8266 Driver] Checking WiFi Device...
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Retrieving Wifi Client Info
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Client is inactive.
[ESP8266 Driver] Checking WiFi AP
[ESP8266 Driver] Retrieving Wifi AP Info
[ESP8266 Driver] AP info:
         SSID      : RaceCapture 34330313020
         Channel   : 8
         Encryption: 0
[ESP8266 Driver] Checking WiFi AP
[ESP8266 Driver] Retrieving Wifi IP info
[ESP8266 Driver] Checking Client
[ESP8266 Driver] Client is inactive.
[ESP8266 Driver] Checking WiFi AP
[ESP8266 Driver] AP configured properly
[ESP8266 Driver] Station IP: 192.168.4.1
[ESP8266 Driver] Checking Server
[ESP8266 Driver] Starting server on port: 7223
[ESP8266 Driver] Socket connected on channel 0
[ESP8266 Driver] Setup called on allocated channel 0
[ESP8266 Driver] Closing stale Serial object
[ESP8266 Driver] Connected on comm channel 0
[wifi] New external connection: Wifi Chan 0
GPS: probing baud rate: 115200
GPS: Using baud rate: 115200
GPS: module detected at: 115200
GPS: detected current update rate: 25
GPS: Configuring NMEA messages
[...]

PR #858, opened 2016-09-13
Delay 1000ms before trying to talk to GPS chip

stieg:
To allow the chip time to get itself together before talking to it.  Hopefully addresses #837


brentpicasso:
looks sane. :shipit: 

Curious, any testing feedback after the change? I know it's completely hit / miss when it happens.

stieg:
Still playing with it.  Difficult to reproduce :(

Issue #844, opened 2016-09-02
Interactive Lua causes system to hang (MK2)


brentpicasso:
STR:
open an interactive lua session in minicom
typed one character at lua command line
observe hang

LED flash pattern - 4 flashes, on the 4th flash it stays lit until the cycle repeats

top of 2.10.0 / debug build

stieg:
Dug into this with debugger.  A few things to follow up on:

* Not sure if we have much of a buffer for the commands since much of that was changed when I moved the USB command handling over to the new rx_buffer code.
* USB stack is most definitely getting smashed since this is taking the entire lua operation when it runs.  Need to find a way to have the LuaTask run this command so we can keep the USB task size small.
* Possible free of a null pointer, but this may also be side effect of stack smash.

@brentpicasso how much is this blocking you?  Not the easiest thing in the world to fix but will poke at it soonish.  I imagine it's not a blocker but mayhaps I am missing something.

brentpicasso:
thanks @stieg  =- Not blocking me right now, I can test by editing and re-running the main script.

PR #859, opened 2016-09-14
Run Lua interactive commands in Lua Task

stieg:
This semi-complex change causes us to run interactive Lua commands
in the Lua Task as opposed to the task hosting the interactive Lua
session.  This is done because we don't know the stack depth of
the task hosting the interactive session, hence we can't gaurantee
it will not overflow when invoking Lua commands.

Issue: #844

### DIFF
--- a/src/lua/luaCommands.c
+++ b/src/lua/luaCommands.c
@@ -26,6 +26,7 @@
 #include "luaScript.h"
 #include "luaTask.h"
 #include "lualib.h"
+#include "macros.h"
 #include "memory.h"
 #include "modp_numtoa.h"
 #include "printk.h"
@@ -39,14 +40,13 @@ void ExecLuaInterpreter(struct Serial *serial, unsigned int argc, char **argv)
                      "to leave");
         put_crlf(serial);
 
-        cmd_context *cmdContext = get_command_context();
-        char *luaLine = cmdContext->lineBuffer;
+        static char luaLine[256];
 
         g_interactive_mode = 1;
 
         for(;;) {
                 serial_write_s(serial, "> ");
-                interactive_read_line(serial, luaLine, cmdContext->lineBufferSize);
+                interactive_read_line(serial, luaLine, ARRAY_LEN(luaLine));
 
                 if (0 == strcmp(luaLine, "exit"))
                         break;
--- a/src/lua/luaTask.c
+++ b/src/lua/luaTask.c
@@ -65,7 +65,8 @@ struct lua_run_state {
 };
 
 enum run_status {
-	LUA_CMD_LOCK_ERROR = -1,
+	LUA_CMD_LOCK_ERROR = -2,
+	LUA_CMD_GENERIC_ERROR = -1,
 	LUA_CMD_FAILURE = 0,
 	LUA_CMD_SUCCESS = 1,
 };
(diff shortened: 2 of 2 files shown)


stieg:
Am going to do a regression test of Lua tomorrow but basic testing both with and without a script showed that the original issue reported is now fixed.  Just need to ensure that no regressions were introduced by this.

src/lua/luaCommands.c
```diff
@@ -39,14 +40,13 @@ void ExecLuaInterpreter(struct Serial *serial, unsigned int argc, char **argv)
                      "to leave");
         put_crlf(serial);
 
-        cmd_context *cmdContext = get_command_context();
-        char *luaLine = cmdContext->lineBuffer;
+        static char luaLine[256];
```
brentpicasso: meh, lots of space in the CCM, right? :)

src/lua/luaTask.c
```diff
@@ -321,42 +376,67 @@ static lua_State* setup_lua_state()
...
+	 * Notice we give the lua_signal twice here. The first time will
+	 * unblock the luaTask so that it may perform our command.  The
+	 * second call completes the interactive command.  This is done this
+	 * way so we may pop all elements/error messages as needed. Therefore
+	 * it is safe to manipulate the Lua state since the lua_interactive
+	 * method will keep the Lua lock until we give the lua_signal the
+	 * second time.
+	 *
+	 * Once this task has completed reading the results of the command
+	 * and has given the lua_signa, the lua_interactive task will
```
brentpicasso: spelling: lua_signal

src/lua/luaTask.c
```diff
@@ -227,9 +275,16 @@ static void lua_task(void *params)
         };
 
         int consecutive_failures = 0;
-        for(portTickType xLastWakeTime;;
-            vTaskDelayUntil(&xLastWakeTime, state.callback_interval)) {
-                xLastWakeTime = xTaskGetTickCount();
+	portTickType wake_tick = 0;
+        for(;;) {
```
brentpicasso: so.. how does this play into how responsive the lua command is based on the current tickRate()?

Will a user enter a command, and then wait for the current tick to fire to see it execute? e.g. default tick rate is 1Hz - it could take up to 1 second for a command to be processed?

brentpicasso:
wow that's a lot but it's good. the thread interaction can be tricky; we should both hammer on this tomorrow. 

Once main question in the changeset.

brentpicasso:
detected bug:

if onTick() is not defined, I get an error:

RaceCapture/Pro MK2 > lua                                                                                                                                  
Entering Lua Interpreter. enter 'exit' to leave                                                                                                            
> println('hello')                                                                                                                                         
Internal Error: Lua not active.                                                                                                                            
>

src/lua/luaTask.c
brentpicasso: Testing it directly it seems the command is processing immediately, even if tickRate() is 1Hz, which seems good, but it seems there's something else I'm not understanding out of reviewing the code.

brentpicasso: More data. while at 1Hz, I rapidly pasted the following into interactive mode, with a carriage return in the clipboard, so it processed immediately after the paste. 

println('asdfasdf')


The commands were happening quickly with manual pastes, but if I held down the keyboard to have it repeat, then the commands started processing at 1Hz! interesting.

src/lua/luaCommands.c
stieg: Yep.  I have a ticket out to be able to pass in the object with the buffer to handle this.. just not part of this patch and not a requirement for release.  Hence this hack.

stieg:
Its expected, but lets talk about it since I'm not sure if it's a bug or a feature.

brentpicasso:
Tested the latest. :shipit:  :D

stieg:
You are up early @brentpicasso .  Merging.

Issue #861, opened 2016-09-16
WiFi UDP beacon stops transmitting


ryandoherty:
Not 100% reproducible, but seems to occur under these conditions:

- Mk2 (tested on this only so far) 2.10.0-RC3-8-gd7b2f41
- AP mode for WiFi module
- Kindle fire connected to AP
- 1.7.5 app

- Connect to AP from tablet
- Open app, change connection type to wifi
- Watch app connect
- Go to dashboard mode
- Wait a few minutes. After ~5 minutes, the app will see a disconnect, then it will reconnect
- Kill app, clear app preferences/cache via Android/OS settings. Do not uninstall.
- Reopen app. 
- Watch app wait for RCP, even though it is connected to RCP's wifi
- Watch wifi activity LED, it will not blink at all.

ryandoherty:
RCP's wifi is still working, if you don't clear the app preferences, it can reconnect because it remembers the IP address previously used to connect to RCP. The UDP beacon is not sending afaict. 

I did a Wireshark sniff from my laptop when connected to RC's AP, no UDP beacons were being sent.

ryandoherty:
Seems like the 'trick' to reproducing this is for there to be a disconnect between RCP and the app. Happens randomly and sometimes never. Not sure what causes it or how to repro.

ryandoherty:
RCP logs (via USB) from around the time UDP beacons stopped:

```lly received.
GPS: Configuring was successful.
GPS: detected current update rate: 1
GPS: Configuring update rate to 10: win
GPS: Configuring NMEA messages
GPS: Message was successfully received.
GPS: Successfully configured NMEA
GPS: configure message type: win
GPS: configure navigation data message interval: win
GPS: Configuring Gnss Navigation Mode: win
GPS: Disabling NMEA messages
GPS: Message was successfully received.
GPS: Successfully disabled NMEA messages
GPS: provisioned
[USB] Starting telem stream on USB
[ESP8266 Driver] Socket connected on channel 1
[wifi] New external connection: Wifi Chan 1
[wifi] Starting telem stream on Wifi Chan 1
[wifi] Stopping telem stream on Wifi Chan 1
[wifi] Starting telem stream on Wifi Chan 1
[wifi] Telemetry stream rate too high.  Reducing to 10
[wifi] Stopping telem stream on Wifi Chan 1
[wifi] Starting telem stream on Wifi Chan 1
[...]

PR #862, opened 2016-09-19
Handle the <0-4>,CONNECT FAIL channel action on ESP8266

stieg:
Seems there is another vague channel action, the CONNECT FAIL action.
We didn't handle this before.  When this happens we should close
the connection.  This patch now does that.  When this fails sometimes
it can cause the WiFi chip to get into a bad state.

Issue #861


brentpicasso:
looks sane. :shipit:  and we'll bang on it some more.

Issue #863, opened 2016-09-20
WiFi AP configuration appends password to AP name and doesn't set password


ryandoherty:
STR:

- Use app 1.7.5
- Firmware 2.10.0-RC3-17-g8283fa5
- Tested on mk2
- Open app, turn on WiFi & AP mode, set AP name and password
- Write config:
``` {"setWifiCfg":{"active":true,"ap":{"active":true,"encryption":"none","password":"1234567","ssid":"RaceCapture/Pro MK2 Ryan","channel":11},"client":{"active":false,"password":"","ssid":""}}}```

- Read config: 
```{"wifiCfg":{"active":true,"client":{"active":false,"ssid":"","password":""},"ap":{"active":true,"ssid":"RaceCapture/Pro MK2 Ryan1234567","password":"1234567","channel":11,"encryption":"none"}}}```

Then with a tablet/phone/computer, join the wifi network RCP created, it will not ask for a password.

Expected behavior: AP name does not contain password and joining network requires a password.

PR #864, opened 2016-09-20
Add strntcpy, convert methods as needed

stieg:
@ryandoherty found a string buffer overflow scenario where it was possible for an external source to give a long enough string that would be copied in unterminated which could cause undesired behavior.  This PR fixes that by introducing the `strntcpy` method, which is like strncpy but always guarantees that any copied string is NULL terminated.

Issue #863


brentpicasso:
:shipit:

PR #866, opened 2016-09-21
Fixes All the Issues Ryan Found

stieg:
Because @brentpicasso and I never were able to find them.  Specifically it does the following:

* Fixes the UDP beacon issue by adding a check on the socket and restarting it if the socket gets closed
* Removes special queue sizing for sockets because of limitation on closing serial ports.
* No longer sends "0.0.0.0" as addresses as the app really doesn't like this.

I have sent builds for @ryandoherty to break.  Hopefully he will be unable to this time.


src/tasks/wifi.c
```diff
@@ -451,23 +451,31 @@ static void do_beacon()
...
+		state.beacon.serial = NULL;
+	}
+
         /*
          * If here then we have at least one IP.  Send the beacon.  Now
          * we need a channel to send the beacon on. Lets grab one if we
          * don't have one yet.
          */
         if (NULL == state.beacon.serial) {
-                state.beacon.serial =
```
brentpicasso: strictly a formatting change it seems. I had to copy and paste this to do a text diff to ensure there were no code changes :P

src/drivers/esp8266_drv.c
```diff
@@ -315,20 +315,22 @@ static struct channel* channel_setup(const unsigned int index,
...
-	 *       be applied since I am not creating a new Serial object.
-	 *       This should get fixed when issue #542 is resolved.  Sorry.
+	 * HACK:
+	 * Because I can no longer destory Serial devices due to the
+	 * chance of hard fault (issue #847), I must instead reopen
+	 * the Serial device if it has already been allocated. This means
+	 * that rx_size and tx_size have to be the default sizes for now.
+	 * This should get fixed when issue #542 is resolved.  Sorry.
 	 */
+	/* if (0 == rx_size) */
```
brentpicasso: recommend removing commented out code

src/tasks/wifi.c
```diff
@@ -437,7 +437,7 @@ static void send_beacon(struct Serial* serial, const char* ips[])
                 for(++ips; *ips && 0 == **ips; ++ips);
                 const bool more = !!*ips;
                 /* Don't add empty strings to the array */
-                if (0 != *ip)
+                if (0 != *ip && !STR_EQ(ip, IPV4_NO_IP_STR))
```
brentpicasso: we were sending 0.0.0.0 as one of the IP addresses?  I missed that part of the earlier convo

src/tasks/wifi.c
```diff
@@ -451,23 +451,31 @@ static void do_beacon()
...
 
+	/*
+	 * Check if the socket is closed.  If so, whine about it then move
+	 * on and try re-opening it.
+	 */
+	if (state.beacon.serial &&
+	    !serial_is_connected(state.beacon.serial)) {
+		pr_warning("Beacon socket closed!  WTF!\r\n");
+		state.beacon.serial = NULL;
+	}
```
brentpicasso: so basically the rug got pulled out from underneath the socket; this ensures the serial is NULL'd so it's forced to open again?

src/drivers/esp8266_drv.c
stieg: See comment above.  There is a reason I left it in there.

src/tasks/wifi.c
stieg: Yes we have been doing that for some time now and it is just something that we shouldn't do.  So I went ahead and fixed it.

src/tasks/wifi.c
stieg: Yep.

src/tasks/wifi.c
stieg: Yes.  I originally put the change here and then moved it out.  So this was some leftover from that.

stieg:
Resolves #861

Issue #874, opened 2016-10-12
Altitude go 140911008 when below 0


mumchristmas:
I live in Shanghai. Some points of the city are below the sea level. :)

stieg:
Oops... that is no good.  Thanks for reporting.

stieg:
Hey @mumchristmas Do you have a log file perhaps that you can upload that shows this?  Also what version of the firmware are you running?  TIA.

mumchristmas:
[RCT-datapoint-161012-ColtMa.xlsx](https://github.com/autosportlabs/RaceCapture-Pro_firmware/files/525198/RCT-datapoint-161012-ColtMa.xlsx)

This is what dig form the sq3 file. See the line 7097 to 7539.
My RCT is running official 2.10.0 fw.

stieg:
Huh... well this is pretty funny.  Below is the datasheet from our GPS chip vendor:

![screenshot from 2016-10-12 13-17-53](https://cloud.githubusercontent.com/assets/141519/19322372/5f5c152e-907e-11e6-987a-ae442356f45a.png)

We make use of the mean sea level altitude calculation.  Notice how according to the docs this is a UINT32.  Ha!  Wonder if this is a typo in the document or if its a design oversight.  Will reach out to our vendor.

stieg:
Looks like documents were wrong, hence the bug:

![screenshot from 2016-10-13 10-08-15](https://cloud.githubusercontent.com/assets/141519/19354622/0ba1ab60-912d-11e6-87d0-b86e6e0b2861.png)

Will update code accordingly.

PR #876, opened 2016-10-13
Convert GPS altitude to signed type

stieg:
Previously our information on this field indicated that it was an
unsigned type.  This has been shown to be incorrect as is noted
on issue #874.  This commit resolves that issue.

Issue #874


brentpicasso:
looks sane to me

brentpicasso:
:shamrock:

PR #879, opened 2016-10-17
updates to track autodetection semantics

### LINKED ISSUE #878: always report back configured trackId if track database is not present
In cases when a track database is not present (size == 0) as in RCT,  the meaning of the auto configure flag sould mean that the current track settings should be manipulated externally -it's a flag for the external app to use.   When we do have a track database, as in RCP, the flag should be used internally to manage itself.

To support this, the currently configured trackId should always be reported regardless of the auto configuration setting, *if* the track database is not present. 

Also, we should normalize configuration to always enable track auto-configuration by default.

brentpicasso:
This PR affects RCT:
* always report configured track ID if we have no track database
* normalizes default configuration.
* Allows auto configure flag to have a useful meaning in RCT, where currently it's a flag that's present but ignored.

This is to match updates to app to enhance track auto-detection, in PR: https://github.com/autosportlabs/RaceCapture_App/pull/1278

Resolves #878

### DIFF
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -651,11 +651,13 @@ static void lapstats_setup(const GpsSnapshot *gps_snapshot)
         const Track *track = NULL;
         track_status_t track_status;
         if (auto_detect_track) {
-                track_status = TRACK_STATUS_AUTO_DETECTED;
-                track = auto_configure_track(NULL, gp);
+                const Track *default_track = (get_tracks()->count == 0) ? &trackConfig->track : NULL;
+                track = auto_configure_track(default_track, gp);
+                track_status = default_track->trackId ? TRACK_STATUS_AUTO_DETECTED : TRACK_STATUS_WAITING_TO_CONFIG;
                 if (track)
                         pr_info_int_msg("track: detected track ",
                                         track->trackId);
+
         } else {
                 track_status = TRACK_STATUS_FIXED_CONFIG;
                 track = &trackConfig->track;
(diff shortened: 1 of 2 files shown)


src/lap_stats/lap_stats.c
```diff
@@ -651,11 +651,13 @@ static void lapstats_setup(const GpsSnapshot *gps_snapshot)
         const Track *track = NULL;
         track_status_t track_status;
         if (auto_detect_track) {
-                track_status = TRACK_STATUS_AUTO_DETECTED;
-                track = auto_configure_track(NULL, gp);
+                const Track *default_track = (get_tracks()->count == 0) ? &trackConfig->track : NULL;
+                track = auto_configure_track(default_track, gp);
+                track_status = default_track->trackId ? TRACK_STATUS_AUTO_DETECTED : TRACK_STATUS_WAITING_TO_CONFIG;
```
stieg: This code will result in a SEGFAULT if count is != 0 since default_track will be null.

brentpicasso: Check the path of the original code - it was designed to handle the null case when a null track is passed into set_active_track(). Here, I'm retaining that behavior for the original case. Let me know if I missed anything.

src/lap_stats/lap_stats.c
```diff
@@ -639,27 +640,38 @@ static void lapstats_setup(const GpsSnapshot *gps_snapshot)
...
                 track_status = TRACK_STATUS_AUTO_DETECTED;
-                track = auto_configure_track(NULL, gp);
-                if (track)
-                        pr_info_int_msg("track: detected track ",
+                track = auto_configure_track(&trackConfig->track, gp);
+                if (track != &trackConfig->track) {
+                        pr_info_int_msg(_LOG_PFX "Auto-detected track from db ",
                                         track->trackId);
+                }
+                else {
```
stieg: Formatting is off.

Issue #808, opened 2016-08-13
tab character in channel name creates corrupt meta data


brentpicasso:
Related app issue: https://github.com/autosportlabs/RaceCapture_App/issues/1117

STR:
* create analog channel named ABC; in the channel customization popup add a tab character after the name (press tab). Or, edit configuration file directly to inject tab (ASCII 9) into channel name.

* write configuration in app
* power cycle and re-connect
* observe many errors in meta data reported by the app.  JSON appears corrupted:

{
	"meta": [{
				"nm": "Interval",
				"ut": "ms",
				"min": 0,
				"max": 0,
				"prec": 0,
				"sr": 1
			}, {
				"nm": "Utc",
				"ut": "ms",
				"min": 0,
				"max": 0,
				"prec": 0,
				"sr": 1
			}, {
				"nm": "ABC "
				","
				ut ":"
				ratio ","
				min ":0.0,"
				max ":20.0,"
				prec ":2,"
				sr ":5},{"
				nm ":"
				Battery ","
				ut ":"
				Volts ","
				min ":0.0,"
				max ":20.0,"
				prec ":2,"
				sr ":1},{"
				nm ":"
				AccelX ","
				ut ":"
				G ","
				min ":-3.0,"
				max ":3.0,"
				prec ":2,"
				sr ":25},{"
				nm ":"
				AccelY ","
				ut ":"
				G ","
				min ":-3.0,"
				max ":3.0,"
				prec ":2,"
				sr ":25},{"
				nm ":"
				AccelZ ","
				ut ":"
				G ","
				min ":-3.0,"
				max ":3.0,"
				prec ":2,"
				sr ":25},{"
				nm ":"
				Yaw ","
				ut ":"
				Deg / Sec ","
				min ":-120.0,"
				max ":120.0,"
				prec ":2,"
				sr ":25},{"
				nm ":"
				Pitch ","
				ut ":"
				Deg / Sec ","
				min ":-120.0,"
				max ":120.0,"
				prec ":2,"
				sr ":25},{"
				nm ":"
				Roll ","
				ut ":"
				Deg / Sec ","
				min ":-120.0,"
				max ":120.0,"
				prec ":2,"
				sr ":25},{"
				nm ":"
				Latitude ","
				ut ":"
				Degrees ","
				min ":-180.0,"
				max ":180.0,"
				prec ":6,"
				sr ":10},{"
				nm ":"
				Longitude ","
				ut ":"
[...]

PR #880, opened 2016-10-20
Properly encode/decode JSON strings in firmware

stieg:
Turns out we were not properly encoding/decoding JSON strings.
Certain characters are considered illegal in a JSON stream and
thus must be encoded (escaped) so they become legal charaters.
They then must be unencoded on the receiving end so that their
meaning is restored.

This change adds in encode and decode methods, makes use of them
at the appropriate spots and removes broken code that previously
was doing this incorrectly/incompletely.

Note: We don't support unicode encoded charaters.  I just replace
those with a '?' if I see them.

Issue #808

### DIFF
--- a/src/jsmn/jsmn.c
+++ b/src/jsmn/jsmn.c
@@ -19,8 +19,10 @@
  * this code. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include "api.h"
 #include "jsmn.h"
 #include "macros.h"
+#include "serial.h"
 #include "str_util.h"
 #include <stdbool.h>
 #include <stddef.h>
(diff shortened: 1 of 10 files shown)


src/jsmn/jsmn.c
```diff
@@ -372,7 +374,7 @@ bool jsmn_exists_set_val_int(const jsmntok_t* root, const char* field,
 	if (!node)
 		return false;
 
-	int* value = val;
+	int* value = (int*) val;
```
brentpicasso: were these casts for good form? or to suppress warnings?

stieg: Both?

stieg: Suppress warnings from the C++ compiler.  And the cast is good form anyways.

stieg:
@brentpicasso So good to go?

brentpicasso:
:shipit:

Issue #881, opened 2016-10-20
SIM900 Connects But Code Thinks Its Not Connected


stieg:
Logic bug in code most like:

```
[cell] Resetting modem...

[serial (Cell)] TX: AT(\r)(\n)
[serial (Cell)] RX: AT(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)[cell] auto-baud successful

[serial (Cell)] TX: ATE0(\r)(\n)
[serial (Cell)] RX: ATE0(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)
[serial (Cell)] TX: AT+CGMI(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: SIMCOM_Ltd(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)[cell] manufacturer is SIMCOM_Ltd

[cell] Loading driver for SIM900

[serial (Cell)] TX: AT+IFC=0,0(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)
[serial (Cell)] TX: AT+CIPMUX=0(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)
[serial (Cell)] TX: AT+CIPMODE=1(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)
[serial (Cell)] TX: AT+CGSN(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: 866771029563243(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)
[serial (Cell)] TX: AT+CSQ(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: +CSQ: 0,0(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[serial (Cell)] RX: OK(\r)(\n)
[serial (Cell)] TX: AT+CNUM(\r)(\n)
[serial (Cell)] RX: (\r)(\n)
[...]

PR #882, opened 2016-10-20
Fix SIM900 connect string comparisons

stieg:
Since I had to modify our serial sentence parser to keep the \r\n
in the parsed messages, I needed to patch our comparison strings to
include the \r\n characters as well.  Seems that I missed a couple.

Issue #881


stieg:
Tested and verified locally.

PR #883, opened 2016-10-20
Only log samples if they are newer than previous ones.

stieg:
With slow SD cards you can get yourself in a situation where you log
older samples.  This happens when init of the card takes a while.
To fix this I just check the time on the sample and ensure its always
higher than the last one.  If not, skip it.

Issue #793


brentpicasso:
:shipit:

Issue #872, opened 2016-10-03
malformed status JSON when cellular is enabled


brentpicasso:
STR:

2.10.0 RCP
resetConfig
enable cellular
restart unit
connect app
observe messages in log:
```json
'{"status":{"system":{"model":"RaceCapture/Pro MK2","ver_major":2,"ver_minor":10,"ver_bugfix":0,"serial":"57003E000351343038383337","git_info":"2.10.0","release_type":"official","uptime":9337},"GPS":{"init":1,"qual":0,"lat":0.0,"lon":0.0,"sats":0,"DOP":0},"cell":{"init":1,"IMEI":"355922062181330' | Unterminated string starting at: line 1 column 280 (char 279)
```

stieg:
So this message just gets printed to the log?  Any other information around where the message is printed?

brentpicasso:
This gets printed to the app as a response to the status log query as soon as cellular is enabled and the unit is power cycled.

stieg:
An example of the bug:

```
[DEBUG  ] [RCPAPI      ] Tx: {"getStatus":null}
[DEBUG  ] [RCPAPI      ] Rx: {"status":{"system":{"model":"RaceCapture/Pro MK2","ver_major":2,"ver_minor":10,"ver_bugfix":0,"serial":"2A0039001347333231343238","git_info":"2.10.0-8-g3665ac5","release_type":"development","uptime":19091},"GPS":{"init":1,"qual":0,"lat":0.0,"lon":0.0,"sats":0,"DOP":0},"cell":{"init":0,"IMEI":"","sig_str":0,"number":"","state":"unknown"},"telemetry":{"status":0,"state":"idle","dur":0},"bt":{"init":2},"logging":{"status":0,"dur":0},"track":{"status":0,"valid":false,"trackId":0,"inLap":0,"armed":0},"imu":{"init":true},"wifi":{"active":false,"initialized":false,"ap":{"active":false},"client":{"active":false,"connected":false}}}}
[DEBUG  ] [RCPAPI      ] Tx: {"getStatus":null}
[...]

PR #884, opened 2016-10-21
Strip IMEI content when read from serial line

stieg:
Else the \r\n characters remain in there and that brakes the mind of
our app.

Issue #872


brentpicasso:
:shipit:

PR #887, opened 2016-11-06
Change the read/write CSer Lua methods back to raw character methods

stieg:
Some confusion in the docs lead to a regression on the behavior of the
read/write CSer Lua methods.  Apparently they were intended to write
RAW characters but this was not clear.  This patch reverts it back
to the old behavior so the scripts will work again.


brentpicasso:
:shipit:

PR #889, opened 2016-11-07
Fix version printing, cleanup version macros and methods

stieg:
Fixes the version printing as reported in issue #886.  Also fixes
the need to update the version info in the Makefile, which is now
derived from the git tag.  We also clean up some macros to remove
some of the less flexible code so we can create the user info as
needed.


brentpicasso:
Looks good. :shipit:

Issue #870, opened 2016-10-02
Optimize scheduling of OBDII PIDs of different sample rates


brentpicasso:
We should query OBDII PIDs in an optimized way so that faster channels are queried proportionately more often than slower channels.

e.g. RPM @ 50Hz and fuel level @ 1Hz. 

Idea:

maintain a counter for each channel that counts down on every round robin. the channels that count down sooner will 'fire' and be scheduled for querying

mumchristmas:
Excuse me, how will the RCT handle the sample rate mismatch (like the e.g.) with 2.10.0 firmware?
Should I set every channel with same rate now?

stieg:
@mumchristmas  Internally the unit stores the last sampled value and that is what is returned upon sampling.  So varying sample rates will not make too much of a difference here.  The OBD-II code does round robin queries as fast as it can, thus updating its samples as fast as it can.

meikeric:
Please take a look at PR:https://github.com/autosportlabs/RaceCapture-Pro_firmware/pull/892 and let me know your thoughts. I tried to keep it tight, fast and in the loop.

PR #892, opened 2016-11-15
Optimize scheduling of OBDII PIDs of different sample rates #870

meikeric:
Work on Issue 870: Optimize scheduling of OBDII PIDs of different sample rates
On each OBD2 loop, fetch the current milliseconds, create a bitmap of timeouts. Perform an AND on the timeout bitmask against the OBD2Config.Pid[].cfg.sampleRate

### DIFF
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -274,11 +274,22 @@ typedef struct _PWMConfig {
 
 #define OBD2_CHANNELS 20
 
+typedef enum _OBD2TimeOuts
+{
+    Hz0_2 = 1 << 0, // 5 seconds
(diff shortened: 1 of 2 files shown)


brentpicasso:
Hi - thanks for the PR! We're in the process of completely overhauling the OBDII and CAN mapping infrastructure, which will incorporate this type of optimization while fitting into the bigger requirements. We'll   consider the ideas you have here into the work we're doing.  Thanks again!

stieg:
Per @brentpicasso 's note, since we are working on new bits here I'm going to close these for now. @meikeric If you would like to contribute please contact @brentpicasso and we can discuss best ways to help out.

Issue #172, opened 2015-10-26
Discrete CAN mapping capabilities


ryandoherty:
The ability to map CAN data values to a channel, as a more user friendly mapping capability before having to use Lua scripting. Lua Scripting would be reserved for more elaborate mappings requiring custom logic.

Each Mapping has the following configuration:
- A standard ChannelConfig:
- Name
- Units
- min
- max
- logging precision
- CAN Mapping
- Source data type(unsigned, signed, IEEE754)
- Endian
- CAN channel
- CAN ID
- Bit offset within CAN message
- bit width
- Value multiplier (float)
- Value offset (float)
# 

A maximum of 100 mapping can be stored in the config (# TBD)

Upon the receipt of each CAN message, the ID is checked against a known mapping.

If a CAN ID is matched, the value is computed from the bit offset, then adjusted using the multiplier and offset Then, the channel value is set, presented in the telemetry stream.

stieg:
@brentpicasso I believe you are working on this so assigning this to you.

kosenko-max:
That would be wonderful if that will come together with some bitwise math option (with access to current values of channels) and https://en.wikipedia.org/wiki/ISO_15765-2 support. Otherwise in a majority of cases that I have in Evo X CAN it won't really work. I can't say how widespread that standard is, but it is a mandatory standard pretty much, so I assume that "a lot".

I suggest to consider that many CAN messages contain a lot of data (I have some that bring 20 channels in one multipart message), so it would be great not to do same request 20 times. 

It's also would be extremely helpful if possible to define priority of CAN channels. So that some CAN requests made only 1 out of X times to give more bandwidth for high speed sensors, but be able to collect a lot of slow moving data as well.

brentpicasso:
Related:
https://www.autosportlabs.com/next-generation-can-bus-mapping/

The new CAN mapping mechanism should handle 90% of broadcast style CAN data
seen typically from aftermarket ECUs and some OEMS (like seen on the BMW
E46), alleviating the requirement of performing the mapping only in Lua.

OBD-II style request / reply transactions will be handled separately by an
enhancement of the OBD-II subsystem, which will catch OEM specific PID
requests (mode 23 et.al.) Lua will remain as a fallback for other types of
mapping. The goal is to iterate on this feature and catch additional
schemes / standards, but always have Lua based mapping as the ultimate
fall-back.


[...]

PR #894, opened 2016-11-16
Discrete CAN mapping capabilities #172

meikeric:
Overhaul to CAN processing.

OBD2 PID responses are now out of sync with PID requests. This free's up the system from blocking the CAN and potentially missing other CAN values.  However, we can still do a blocking OBD2 request using OBD2_request_PID for things like LUA.

The CAN_Rx_Task is now the loop for processing CAN Rx messages. It first check's the message against the OBD2 PID config's and then checks it against the new CANMapConfig.CANMapChannelConfig.  One CAN message can in theory match multiple Channel's.

Changed the CAN_device_mock to use a simple FIFO queue so we can test CAN message processing through unit tests.

Created Unit Tests for: CANMap, CAN_device_mock

### DIFF
--- a/include/CAN/CAN.h
+++ b/include/CAN/CAN.h
@@ -38,6 +38,8 @@ CPP_GUARD_BEGIN
 
 #define CAN_MSG_SIZE 8
 
+#define DEFAULT_CAN_TIMEOUT 		100
+
 typedef struct _CAN_msg {
     int isExtendedAddress;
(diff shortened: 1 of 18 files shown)


stieg:
Per @brentpicasso 's note, since we are working on new bits here I'm going to close these for now.  @meikeric If you would like to contribute please contact @brentpicasso and we can discuss best ways to help out.

PR #897, opened 2016-11-20
raise OBDII stack to match RCP to prevent heap smash 

### LINKED ISSUE #895: OBDII causing watchdog on certain vehicles / ECUs
When plugged into certain ECUs, RCT appears to watchdog, and only when OBDII is enabled in the firmware. 

So far 2013 Subaru BRZ, 2015 Mazda CX5 and Alfa Romeo 4C is showing the problem. 

Issue does not happen if OBDII is disabled in firmware config.  Theory is OBDII task is receiving some invalid data that's causing the firmware to crash.

brentpicasso:
resolved #895

### DIFF
--- a/src/OBD2/OBD2_task.c
+++ b/src/OBD2/OBD2_task.c
@@ -34,7 +34,7 @@
 #include "taskUtil.h"
 #include "capabilities.h"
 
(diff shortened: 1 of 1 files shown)


brentpicasso:
Self-reviewing as this was a trivial fix and is verified to resolve issue in the field.

PR #902, opened 2017-02-02
P/brent/mk3

brentpicasso:
Hardware support for MK3. 

In addition to the new / changed hardware features, refactored the following:

* GPS code a bit so skytraq logic is centralized and not duplicated. 
* fileWriter and related write statuses based on some suggestions in code base. 
* removed old MK1 code from connectivityTask

Updates to reflect hardware changes:

Analog input #8
RPM Input #4
GPS Reset control
ESP8266 boot control pin
ESP8266 reset pin
SD card detection pin
CAN 1 termination control
CAN 2 termination control
Aux2 serial port TX
Aux2 serial port RX
WiFi status LED
Bluetooth status LED
Error Status LED
CAN status LED
Telemetry status LED
SD card status LED
GPS status LED
50Hz WiFi
Remove PWM / analog out feature
update product name


include/logger/loggerConfig.h
```diff
@@ -318,6 +318,9 @@ typedef struct _OBD2Config {
 typedef struct _CANConfig {
     unsigned char enabled;
     int baud[CONFIG_CAN_CHANNELS];
+#if CAN_SW_TERMINATION == true
```
stieg: Does this need this `#ifdef` or can we just leave the termination Boolean array there in even in platforms that don't support it?  Train of thought here is to minimize `#ifdefs` in code where reasonable.

platform/mk2/config.mk
```diff
@@ -61,7 +61,7 @@ $(HAL_SRC)/fat_sd_stm32/fatfs/option/unicode.c \
 $(HAL_SRC)/fat_sd_stm32/sdcard_device_stm32.c \
 $(HAL_SRC)/fat_sd_stm32/tm_stm32f4_fatfs.c \
 $(HAL_SRC)/fat_sd_stm32/tm_stm32f4_gpio.c \
-$(HAL_SRC)/gps_skytraq/gps_device_skytraq.c \
+$(HAL_SRC)/gps_stm32/gps_device_lld_stm32.c \
```
stieg: What does `_lld_` represent?

platform/mk3/HOW_TO_FLASH.txt
```diff
@@ -0,0 +1,34 @@
+================================
```
stieg: Just a thought... if you write this in GitHub Markdown syntax then you will have a webpage you can point people to directly that will be rendered nicely on GitHub.  YMMV.

platform/mk3/capabilities.h
```diff
@@ -0,0 +1,97 @@
...
+#define VIRTUAL_CHANNEL_SUPPORT	1
+#define WIFI_SUPPORT		1
+
+/* Wifi Specific Info */
+#define WIFI_MAX_BAUD       921600
+#define WIFI_MAX_SAMPLE_RATE    50
+#define WIFI_ENABLED_DEFAULT    true
+
+/* Rx Max Message length */
+#define RX_MAX_MSG_LEN	768
```
stieg: Should this be 1024 or higher?  Or did we reduce the WiFi msg frame down and I missed it?

platform/mk3/capabilities.h
```diff
@@ -0,0 +1,97 @@
...
+ * These values dictate how the LUA garbage collector will behave.
+ * Tweaking these is necessary in low memory environments to ensure
+ * that LUA's memory footprint does not exceed what can be spared.
+ * A value of 0 means that you want to use the default.  For more info
+ * see http://www.lua.org/manual/5.1/manual.html#2.10
+ */
+/* Pause between runs.  < 100 means don't wait */
+#define LUA_GC_PAUSE_PCT	99
+/* Runtime of GC to malloc.  Setting to 10x for agressive behavior. */
+#define LUA_GC_STEP_MULT_PCT	1000
```
stieg: Should evaluate if this is the best setting for our GC since this means the GC is running 10x the rate to allocation.  Wish we had a profiler built in

src/devices/gps_skytraq_s1216_sup500f8.c
```diff
@@ -669,6 +668,8 @@ gps_status_t GPS_device_init(uint8_t sampleRate, struct Serial *serial)
     size_t gps_init_status = GPS_STATUS_NOT_INIT;
 
     while(attempts-- && gps_init_status == GPS_STATUS_NOT_INIT) {
+        /* Perform low level init / reset of GPS device */
+        gps_device_lld_init();
```
stieg: Does this cause a power cycle?  If so be careful of the Mk2 bug RE delay needed before GPS init.

platform/mk3/capabilities.h
```diff
@@ -17,9 +17,9 @@
 #define WIFI_SUPPORT		1
 
 /* Wifi Specific Info */
-#define WIFI_MAX_BAUD		230400
-#define WIFI_MAX_SAMPLE_RATE	10
-#define WIFI_ENABLED_DEFAULT	false
+#define WIFI_MAX_BAUD       921600
```
stieg: Well... be careful here as this is how I found the limitations of the ESP8266 before.  YMMV but if you start seeing dropped characters remember your baud settings.

test/capabilities.h
```diff
@@ -106,6 +106,7 @@ CPP_GUARD_BEGIN
 #define TIMER_CHANNELS			3
 #define PWM_CHANNELS			4
 #define CAN_CHANNELS			2
+#define CAN_SW_TERMINATION      true
```
stieg: You need to add this to the other capabilities files as `false`

platform/mk2/hal/fat_sd_stm32/sdcard_device_stm32.c
```diff
@@ -26,3 +26,8 @@ void disk_init_hardware(void)
 {
 	SD_Init();
 }
+
+bool sdcard_device_card_present(void)
```
stieg: IMHO I would implement this as an enum where the values are as follows:

```c
SD_PRESENT_UNKNOWN = -1,
SD_PRESENT_FALSE = 0,
SD_PRESENT_TRUE = 1,
```

So we can tell the difference in code and logs.  Minor but worth a thought.

src/logger/fileWriter.c
```diff
@@ -280,6 +281,12 @@ static void open_log_file(struct logging_status *ls)
         pr_info(_RCP_BASE_FILE_ "Opening log file\r\n");
         ls->writing_status = WRITING_INACTIVE;
 
+        if (!sdcard_present()) {
+                ls->writing_status = SD_CARD_NOT_PRESENT;
+                pr_error(_RCP_BASE_FILE_ "SD card not present\r\n");
```
stieg: `pr_info`.  No error here.  Just not present.

src/logger/fileWriter.c
```diff
@@ -383,6 +390,9 @@ TESTABLE_STATIC int logging_sample(struct logging_status *ls,
                 if (WRITING_ACTIVE != ls->writing_status)
                         open_log_file(ls);
 
+                if (SD_CARD_NOT_PRESENT == ls->writing_status)
```
stieg: Is this the correct location here since we try top open a log file above this?  I would double check this.

platform/mk3/HOW_TO_FLASH.txt
brentpicasso: good point

platform/mk3/capabilities.h
brentpicasso: I just replicated it from MK2 and RCT. <shrug>. We can bump it later.

platform/mk3/capabilities.h
brentpicasso: So this means the GC is running super aggressive, right? That could also slow down the script.

src/devices/gps_skytraq_s1216_sup500f8.c
brentpicasso: This causes a hard reset of the device, correct.

brentpicasso: In MK3 we have a 50ms delay after releasing the reset line:  

bool gps_device_lld_reset()
{
        GPIO_ResetBits(GPS_RESET_GPIO, GPS_RESET_PIN);
        delayMs(GPS_RESET_DURATION_MS);
        GPIO_SetBits(GPS_RESET_GPIO, GPS_RESET_PIN);
        delayMs(GPS_BOOT_DURATION_MS);
        return true;
}

I think this should be ok.

test/capabilities.h
brentpicasso: Yep, they're there. buried in this PR somewhere, I'm sure.

platform/mk2/hal/fat_sd_stm32/sdcard_device_stm32.c
brentpicasso: Hm... If it proves to be an issue we can roll that in.

src/logger/fileWriter.c
brentpicasso: it felt like an error condition, so...  But I can see that it could be just a normal message. But, message will show up at a higher priority, which I think is important.

src/logger/fileWriter.c
brentpicasso: that's a good point. I should move it above.

brentpicasso: Ah, I remember. open_log_file(ls) will modify the struct, so that's why we check afterwards. So, I'll leave it as-is.

Issue #108, opened 2015-10-26
Predictive Timer Jumps At End of Lap / Freezes Other Times


ryandoherty:

stieg:
Unfortunately not enough info to work with here.  Re-open if there is info + logs to work with.

brentpicasso:
This is observed frequently enough that we need to leave it open and get some example data.

ryandoherty:
I may have logs, along with Scott. Will dig.

stieg:
Finally observed this at Vodden the Hell 2016.  Timer jumps its value at the end of a lap.  This may be due to an log buffer bounds violation when doing the timing projection.  If this is the case, it could be displaying old lap data, causing the jump that users are reporting.  Julian Cordle from Snowspeeders reported the issue to me in 2.9.1 MK2 at the track and I saw it on MK1 2.8.7.

stieg:
Looking at the code the issues likely lies in `findTwoClosestPts` and `getSplitAgainstFastLap`.  It seems that if we can't find good values to use for the projection due to an overflow of data in the buffer then we could get values that could cause us to return a previous answer (hence a Freeze effect) or values that cause a bad calculation that could lead to invalid results.  Will continue digging here...

wcorredor:
This is still happening in the 2.10.2. I observed the same behavior at Sebring on Jan 23 2016. Lap prediction seems to be correct until the very end of the lap (about 300 yards from finish line) and suddenly it goes up by about 3 seconds until one crosses the start line when the correct time shows up.

brentpicasso:
Hi, thanks for the follow up. 

Do you have any log file data you can upload that catches this in the act? If we have the raw data we might be able to pump it back through the predictive timer to see if it reproduces. 

Thanks!

wcorredor:
Yes, I do have data. In the car at the moment.

We are travelling today to Atlanta for Chumpcar this weekend. I will send it to you as soon as I have some piece of mind.

I still owe you the E30 script; still doing some adjustments and testing before publishing it. It is great but it has gotten a bit large as I have added lots of sensors and information. I have struggled a lot with the amount of memory available for the script. I have reduced the variables to the bare minimum to use the stack as much as possible, but I am running out of juice and can’t add not even one new idea.

I hope this weekend I can finish “mpg”. It has proven harder that I thought and difficult to test while racing. As I posted a TotalDistante() function (that does not reset every lap) would help to reduce code and be more precise.

Regards

[...]

PR #903, opened 2017-02-13
Accept only high quality GPS samples in predictive time code

stieg:
Adds checks that ensure we only accept high quality GPS snapshot values.  "High quality" is defined as:

* Differential GPS is available
* Our dilution of precision <= 2.0

The idea here is that low quality samples are getting into our predictive timer when racers are going under bridges or near mountains that can briefly affect the GPS quality.  By restricting what samples we accept we ensure that the samples we do have are of high enough quality to be "accurate enough".  This will translate into more accurate predictive times and hopefully eliminate the jumps some users are seeing when they get inaccurate samples in the predictive timer queue.

The potential downside to this is that there are longer periods of time where we don't record samples.  This will make the system much less accurate.  However I am willing to accept this risk because of how much the system already relies on high GPS accuracy.

Also there is a patch here that addresses newer GCC compilation issues.

Issue #108


brentpicasso:
Looks pretty straightforward. I think we should road test this before we merge. I can do it; are you set up to do it also?

stieg:
Wow totally missed the comment.  Not really setup currently but could be.  Probably would just use RCT as that is easier to muck with.

brentpicasso:
No problem; can do a drive test this weekend too.

brentpicasso:
More data; This video from Matt shows the lap time freezing then showing a very high value moments before end of lap. Apparently this was repeatable lap-to-lap.   We need to access the data to verify, but from the terrain it looks like there's no reason GPS would suddenly degrade. 

https://www.youtube.com/watch?v=3Lf4Mojoa6I

stieg:
@brentpicasso Lets solve one issue at a time.  I'm sure there are several.  Unless you see good reason not to merge this lets get it in.

brentpicasso:
agreed. :shipit:

PR #905, opened 2017-02-18
P/brent/next can obd2 api

brentpicasso:
* Implements the CAN mapping API and enhanced OBDII API.  Mapping functionality will appear next.

Also:
* Contains coincidental refactorings - removing dead code; refactoring unit tests around API
* Improves logic around automatically resetting logger config (if the config size changed, we reset config automatically)

### DIFF
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -356,8 +356,14 @@ typedef struct _CANChannelConfig {
 typedef struct _PidConfig {
     ChannelConfig cfg;
     CANMapping mapping;
+
+    /* flag for passive mode where we only listen for OBDII PID responses */
     uint8_t passive;
+
+    /* The OBDII PID to query */
     uint8_t pid;
+
+    /* The OBDII mode specified in the query */
     uint16_t mode;
 } PidConfig;
 
@@ -573,7 +579,7 @@ int getConnectivitySampleRateLimit();
 int encodeSampleRate(int sampleRate);
 int decodeSampleRate(int sampleRateCode);
 
-uint8_t filterBgStreamingMode(uint8_t mode);
+uint8_t filter_background_streaming_mode(uint8_t mode);
 
 PWMConfig * getPwmConfigChannel(int channel);
 char filterPwmOutputMode(int config);
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -337,28 +337,42 @@ enum CANMappingEndian {
 
 
 typedef struct _CANMapping {
+    /* The standard channel configuration */
+    ChannelConfig channel_cfg;
+
     /* CAN ID to match this mapping */
     uint32_t can_id;
+
     /* mask to apply to ID */
     uint32_t can_mask;
+
     /* multiplier for conversion formula */
     float multiplier;
+
     /* divider for conversion formula */
     float divider;
+
     /* adder for conversion formula */
     float adder;
+
     /* flag to indicate bit mode or byte mode for offset and length */
     uint8_t bit_mode;
+
     /* indicates the encoding of the raw data type: unsigned, signed, IEEE 754 floating point */
     uint8_t type;
+
     /* the can bus we expect the data to show up on */
     uint8_t can_channel;
+
     /* indicates endian-ness for multi-byte values */
     uint8_t big_endian;
+
     /* byte or bit offset of the data to extract within the CAN message */
     uint8_t offset;
+
     /* byte or bit length of the data to extract within the CAN message */
     uint8_t length;
+
     /* the conversion filter to apply to quickly convert units without changing mapping */
     uint8_t conversion_filter_id;
 } CANMapping;
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -326,24 +326,49 @@ typedef struct _CANConfig {
 #endif
 } CANConfig;
 
+/* define max offsets and length for CAN mappings */
+#define MAX_CAN_MAPPING_OFFSET_BYTES 7
+#define MAX_CAN_MAPPING_LENGTH_BYTES 4
+
+enum CANMappingEndian {
+    CANMappingEndian_Big = 0,
+    CANMappingEndian_Little
+};
+
+
 typedef struct _CANMapping {
+    /* CAN ID to match this mapping */
     uint32_t can_id;
+    /* mask to apply to ID */
     uint32_t can_mask;
+    /* multiplier for conversion formula */
     float multiplier;
+    /* divider for conversion formula */
     float divider;
+    /* adder for conversion formula */
     float adder;
+    /* flag to indicate bit mode or byte mode for offset and length */
     uint8_t bit_mode;
+    /* indicates the encoding of the raw data type: unsigned, signed, IEEE 754 floating point */
     uint8_t type;
+    /* the can bus we expect the data to show up on */
     uint8_t can_channel;
-    uint8_t endian;
+    /* indicates endian-ness for multi-byte values */
+    uint8_t big_endian;
+    /* byte or bit offset of the data to extract within the CAN message */
     uint8_t offset;
+    /* byte or bit length of the data to extract within the CAN message */
     uint8_t length;
+    /* the conversion filter to apply to quickly convert units without changing mapping */
     uint8_t conversion_filter_id;
 } CANMapping;
 
 typedef struct _CANMappingConfig {
+    /* all available CAN mappings */
     CANMapping can_mappings[CONFIG_CAN_MAPPINGS];
+    /* globally enables/disables CAN mappings */
     uint8_t enabled;
+    /* number of mappings set within configuration */
     uint16_t enabled_mappings;
 } CANMappingConfig;
 
@@ -473,9 +498,6 @@ typedef struct _ConnectivityConfig {
         struct wifi_cfg wifi;
 } ConnectivityConfig;
 
-#define SD_LOGGING_MODE_DISABLED					0
-#define SD_LOGGING_MODE_CSV							1
-
 /**
  * Configurations specific to our logging infrastructure.
  */
@@ -553,8 +575,7 @@ int getConnectivitySampleRateLimit();
 int encodeSampleRate(int sampleRate);
 int decodeSampleRate(int sampleRateCode);
 
-unsigned char filterBgStreamingMode(unsigned char mode);
-unsigned char filterSdLoggingMode(unsigned char mode);
+uint8_t filterBgStreamingMode(uint8_t mode);
 
 PWMConfig * getPwmConfigChannel(int channel);
 char filterPwmOutputMode(int config);
@@ -575,6 +596,9 @@ char filterGpioMode(int config);
 ImuConfig * getImuConfigChannel(int channel);
 int filterImuMode(int mode);
 int filterImuChannel(int channel);
+
+uint8_t filter_can_mapping_endian(uint8_t value);
+uint8_t filter_can_channel(uint8_t value);
 
 unsigned int getHighestSampleRate(LoggerConfig *config);
 size_t get_enabled_channel_count(LoggerConfig *loggerConfig);
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -356,8 +356,9 @@ typedef struct _CANChannelConfig {
 typedef struct _PidConfig {
     ChannelConfig cfg;
     CANMapping mapping;
-    unsigned short pid;
-    unsigned short mode;
+    uint8_t passive;
+    uint8_t pid;
+    uint16_t mode;
 } PidConfig;
 
 
@@ -605,6 +606,7 @@ bool should_sample(const int sample_rate, const int max_rate);
 int getHigherSampleRate(const int a, const int b);
 
 int flashLoggerConfig(void);
+void reset_logger_config(void);
 int flash_default_logger_config(void);
 
 void logger_config_reset_gps_config(GPSConfig *cfg);
--- a/include/logger/loggerConfig.h
+++ b/include/logger/loggerConfig.h
@@ -375,14 +375,20 @@ typedef struct _CANMapping {
     uint8_t conversion_filter_id;
 } CANMapping;
 
-typedef struct _CANMappingConfig {
+typedef struct _CANChannel {
+    /* The standard channel configuration */
+    ChannelConfig channel_cfg;
+    CANMapping mapping;
+} CANChannel;
+
+typedef struct _CANChannelConfig {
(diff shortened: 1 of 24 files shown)


include/logger/loggerConfig.h
```diff
@@ -272,59 +276,104 @@ typedef struct _PWMConfig {
...
 
 
 typedef struct _OBD2Config {
     unsigned char enabled;
     unsigned short enabledPids;
-    PidConfig pids[OBD2_CHANNELS];
+    PidConfig pids[CONFIG_OBD2_CHANNELS];
 } OBD2Config;
 
-#define DEFAULT_ENABLED_PIDS 1
```
brentpicasso: this was dead code.

include/logger/loggerConfig.h
```diff
@@ -451,9 +498,6 @@ typedef struct _ConnectivityConfig {
         struct wifi_cfg wifi;
 } ConnectivityConfig;
 
-#define SD_LOGGING_MODE_DISABLED					0
```
brentpicasso: this was dead code.

src/logger/loggerConfig.c
```diff
@@ -349,23 +363,11 @@ int decodeSampleRate(int rate_code)
     }
 }
 
-unsigned char filterBgStreamingMode(unsigned char mode)
+uint8_t filter_background_streaming_mode(uint8_t mode)
 {
     return mode == 0 ? 0 : 1;
 }
 
-unsigned char filterSdLoggingMode(unsigned char mode)
```
brentpicasso: this was dead code.

src/logger/loggerConfig.c
```diff
@@ -690,14 +697,24 @@ int flashLoggerConfig(void)
...
+        pr_warning(_LOG_PFX "size of LoggerConfig changed\r\n");
+    }
+    return changed;
+}
+
 static bool checkFlashDefaultConfig(void)
 {
         const VersionInfo sv = g_savedLoggerConfig.RcpVersionInfo;
-        bool changed = version_check_changed(&sv, "Config DB");
+        bool changed = version_check_changed(&sv) || _config_size_changed();
```
brentpicasso: This helps prevent RC from crashing when we muck with the size of the logger config. It's an extra layer of protection on top of checking the major / minor version #

test/Makefile
```diff
@@ -96,7 +96,7 @@ INCLUDES = \
 CPP = g++
 CC = gcc
 
-CPPFLAGS := -g -DRCP_TESTING -DAPI_REV=$(API) \
+CPPFLAGS := -g -DRCP_TESTING -DAPI_REV=$(API) -Werror=unused-variable\
```
brentpicasso: This helped clean up our unit tests

src/logger/loggerApi.c
```diff
@@ -1352,6 +1347,133 @@ int api_setCanConfig(struct Serial *serial, const jsmntok_t *json)
...
+        channel_max += index;
+        if (channel_max > CONFIG_CAN_MAPPINGS) {
+            return API_ERROR_PARAMETER;
+        }
+
+        for (chans_tok++; index < channel_max; index++) {
+            CANChannel *pid_cfg = can_channel_cfg->can_channels + index;
+            ChannelConfig *chCfg = &(pid_cfg->channel_cfg);
+
+            int result = _set_can_mapping(chans_tok, &(pid_cfg->mapping));
```
stieg: So if there is an error in doing this do we destroy prior channel mappings?  If so we should avoid that by keeping a local copy and only setting the config after all data has been parsed and validated.

include/logger/loggerConfig.h
```diff
@@ -326,24 +326,49 @@ typedef struct _CANConfig {
 #endif
 } CANConfig;
 
+/* define max offsets and length for CAN mappings */
+#define MAX_CAN_MAPPING_OFFSET_BYTES 7
+#define MAX_CAN_MAPPING_LENGTH_BYTES 4
+
+enum CANMappingEndian {
+    CANMappingEndian_Big = 0,
```
brentpicasso: I wasn't sure which way to default, but since the processor is little endian by default, I just chose that as the default as well. I think it's about 50/50 out there for CAN mappings in the wild.

include/logger/loggerConfig.h
```diff
@@ -272,59 +276,104 @@ typedef struct _PWMConfig {
 #define DEFAULT_PWM_CONFIG {DEFAULT_PWM_CHANNEL_CONFIG, MODE_PWM_FREQUENCY, MODE_LOGGING_PWM_DUTY, DEFAULT_PWM_DUTY_CYCLE, DEFAULT_PWM_PERIOD}
 
 
-#define OBD2_CHANNELS 20
+#define DEFAULT_CAN_BAUD_RATE 500000
+
+typedef struct _CANConfig {
+    unsigned char enabled;
```
brentpicasso: Changing to boolean. research shows that it defaults to 8 bits, which is what I originally wanted.

include/logger/loggerConfig.h
```diff
@@ -272,59 +276,104 @@ typedef struct _PWMConfig {
...
+    uint32_t can_id;
+
+    /* mask to apply to ID */
+    uint32_t can_mask;
+
+    /* multiplier for conversion formula */
+    float multiplier;
+
+    /* divider for conversion formula */
+    float divider;
```
brentpicasso: I agree in principal and that's how I had it originally.  However, it's apparently a thing in the CAN mapping world where explicit dividers are specified and CAN .dbc files have an explicit divider stated, even though it could be represented as a fractional multiplier.  So, I added it to be compatible with the quasi-standard.

include/logger/loggerConfig.h
```diff
@@ -272,59 +276,104 @@ typedef struct _PWMConfig {
...
+} CANChannelConfig;
+
 
 typedef struct _PidConfig {
     ChannelConfig cfg;
-    unsigned short pid;
+    CANMapping mapping;
+
+    /* flag for passive mode where we only listen for OBDII PID responses */
+    uint8_t passive;
```
brentpicasso: changing to boolean.

include/logger/loggerConfig.h
```diff
@@ -272,59 +276,104 @@ typedef struct _PWMConfig {
...
+
+    /* The OBDII PID to query */
+    uint8_t pid;
+
+    /* The OBDII mode specified in the query */
+    uint16_t mode;
```
brentpicasso: I was unsure, but I'm defaulting to not needing an enum currently - there are some standard modes - like 01 - 09 and then mode 0x22, but it seems anything could be specified. So I'm leaving it as  a regular number.

src/logger/loggerApi.c
```diff
@@ -66,25 +65,16 @@
 #include <stdlib.h>
 #include <string.h>
 
-/* Max number of PIDs that can be specified in the setOBD2Cfg message */
-#define MAX_OBD2_MESSAGE_PIDS 10
+/* Max number of channels that can be specified in the setOBD2Cfg message */
+#define MAX_OBD2_MESSAGE_PIDS 4
```
brentpicasso: This is a constraint on how many PIDs at a time can be sent in one call.  IIRC the app currently sends all of them in a group, but there's a limitation in the current app where you can't set more than 10 due to JSON message size. 

So, we need to send N at a time, this is the limit of how many PIDs can be sent in one "set" API call. 

However, right now the 1.9.0 app sends PIDs one by one in discrete 'set' messages.

src/logger/loggerApi.c
```diff
@@ -1352,6 +1347,133 @@ int api_setCanConfig(struct Serial *serial, const jsmntok_t *json)
     return API_SUCCESS;
 }
 
+static void _json_can_mapping(struct Serial *serial, CANMapping *mapping, int more)
```
brentpicasso: Yes, I think we can standardize on put. changing to json_put_can_mapping

we have some places where we use the term 'send' - those can be refactored later.

src/logger/loggerApi.c
```diff
@@ -1375,17 +1368,83 @@ int api_get_can_channel_config(struct Serial *serial, const jsmntok_t *json)
...
+    jsmn_exists_set_val_uchar(chan, "offset", &offset, NULL);
+
+    uint8_t length = can_mapping->length;
+    jsmn_exists_set_val_uchar(chan, "len", &length, NULL);
+
+    /* check for invalid mapping offsets and length */
+    if (offset > MAX_CAN_MAPPING_OFFSET_BYTES * (bit_mode? 8 : 1))
+            return API_ERROR_PARAMETER;
+
+    if (length > MAX_CAN_MAPPING_OFFSET_BYTES * (bit_mode? 8 : 1))
```
brentpicasso: YES. thanks for catching that. :)

src/logger/loggerApi.c
```diff
@@ -1352,6 +1347,133 @@ int api_setCanConfig(struct Serial *serial, const jsmntok_t *json)
...
+        json_float(serial, "mult", mapping->multiplier, DEFAULT_CAN_MAPPING_PRECISION, 1);
+        json_float(serial, "div", mapping->divider, DEFAULT_CAN_MAPPING_PRECISION, 1);
+        json_float(serial, "add", mapping->adder, DEFAULT_CAN_MAPPING_PRECISION, 1);
+        json_int(serial, "type", mapping->type, 1);
+        json_int(serial, "filtId", mapping->conversion_filter_id, more);
+}
+
+int api_get_can_channel_config(struct Serial *serial, const jsmntok_t *json)
+{
+    CANChannelConfig * can_channel_cfg = &(getWorkingLoggerConfig()->can_channel_cfg);
```
brentpicasso: :+1:   I also realized I could use more const as for these new function parameters, too.

platform/rct/Makefile
```diff
@@ -16,7 +16,7 @@ OBJCOPY := $(CROSS_COMPILE)objcopy
 ASL_FW_POSTPROCESS := asl_f4_fw_postprocess
 
 #specify the use of the external crystal
-BOARD_DEFINES = -DPLL_SOURCE_HSE -DUSE_I2C1
+BOARD_DEFINES = -DPLL_SOURCE_HSE -DUSE_I2C1 -fno-zero-initialized-in-bss
```
brentpicasso: I think you're misunderstanding the flag. This will help: http://stackoverflow.com/questions/8721475/if-a-global-variable-is-initialized-to-0-will-it-go-to-bss

src/jsmn/jsmn.c
```diff
@@ -366,6 +366,18 @@ const jsmntok_t * jsmn_find_get_node_value_prim(const jsmntok_t *node, const cha
         return jsmn_find_get_node_value(node, name, JSMN_PRIMITIVE);
 }
 
+bool jsmn_exists_set_val_uchar(const jsmntok_t *root, const char * field, uint8_t *val, uint8_t (*filter)(uint8_t))
+{
+    const jsmntok_t *valueNode = jsmn_find_get_node_value_prim(root, field);
+    if (valueNode) {
+        unsigned char value = atoi(valueNode->data);
+        if (filter != NULL)
+            value = filter(value);
```
brentpicasso: It's a streamlined way for the value to be filtered *if* the value is present in the JSON message.

brentpicasso: will document.

src/jsmn/jsmn.c
```diff
@@ -329,6 +329,18 @@ int jsmn_isNull(const jsmntok_t *tok)
     return strncmp("null", tok->data, 3) == 0;
 }
 
+const jsmntok_t * jsmn_find_node_type(const jsmntok_t *node, const jsmntype_t node_type)
+{
+    if (NULL == node)
+            return NULL;
+
+    for (; node->start || node->end; ++node)
```
brentpicasso: this just emulates the other functions in the jsmn library we incorporated.  It sweeps through the current group of nodes in the json structure from beginning to end.

src/jsmn/jsmn.c
```diff
@@ -366,6 +366,18 @@ const jsmntok_t * jsmn_find_get_node_value_prim(const jsmntok_t *node, const cha
         return jsmn_find_get_node_value(node, name, JSMN_PRIMITIVE);
 }
 
+bool jsmn_exists_set_val_uchar(const jsmntok_t *root, const char * field, uint8_t *val, uint8_t (*filter)(uint8_t))
```
brentpicasso: agreed. unsigned char technically != uint8_t.  Changing to jsmn_exists_set_val_uint8

src/logger/loggerConfig.c
```diff
@@ -197,6 +197,12 @@ static void resetCanConfig(CANConfig *cfg)
     }
 }
 
+uint8_t filter_can_channel(uint8_t value)
+{
+        if (value >= CONFIG_CAN_CHANNELS)
```
brentpicasso: thanks for the min() suggestion. 

I thought about the 2nd part. We have a theme throughout where we filter input values to guarantee things are in range, so I was emulating this. Part of the goal was to be able to easily import CAN mappings from a system that supports 2 CAN channels into systems that support 1.

src/logger/loggerApi.c
brentpicasso: since it's a large chunk of memory it's not feasible to create a local copy.  Might need to do a filtering technique where we rail invalid values.  Will think about this.

brentpicasso: changing to a filtering technique to rail the values so we're at least consistent with the rest of the API.

src/jsmn/jsmn.c
brentpicasso: fixing.

include/logger/loggerConfig.h
```diff
@@ -272,59 +276,104 @@ typedef struct _PWMConfig {
...
+    CANMapping mapping;
+} CANChannel;
+
+typedef struct _CANChannelConfig {
+    /* all available CAN mappings */
+    CANChannel can_channels[CONFIG_CAN_MAPPINGS];
+    /* globally enables/disables CAN mappings */
+    uint8_t enabled;
+    /* number of mappings set within configuration */
+    uint16_t enabled_mappings;
```
brentpicasso: not sure what you mean by sparse value?

platform/rct/capabilities.h
```diff
@@ -33,14 +33,18 @@
...
-#define ANALOG_CHANNELS	1
-#define IMU_CHANNELS	7
-#define	GPIO_CHANNELS	0
-#define TIMER_CHANNELS	0
-#define PWM_CHANNELS	0
-#define CAN_CHANNELS	1
+//Sensor Channels
+#define ANALOG_CHANNELS	        1
+#define IMU_CHANNELS	        7
+#define	GPIO_CHANNELS	        0
```
brentpicasso: fixed

PR #906, opened 2017-02-21
CAN bus mapping and unit tests

brentpicasso:
This builds upon the next_can_obd2_API, containing the mapping scheme + unit tests.

### DIFF
--- a/include/CAN/can_mapping.h
+++ b/include/CAN/can_mapping.h
@@ -19,7 +19,7 @@ CPP_GUARD_BEGIN
  * @param mapping the can mapping to check against
  * @return true if the CAN message matches the mapping's CAN ID and ID mask
  */
-bool canmapping_match_id(CAN_msg *can_msg, CANMapping *mapping);
+bool canmapping_match_id(const CAN_msg *can_msg, const CANMapping *mapping);
 
 /**
  * performs the full mapping against for the CAN message using the CAN mapping
@@ -28,7 +28,7 @@ bool canmapping_match_id(CAN_msg *can_msg, CANMapping *mapping);
  * @param mapping the mapping to be applied to the message
  * @return true if the mapping was successfully applied
  */
-bool canmapping_map_value(float *value, CAN_msg *can_msg, CANMapping *mapping);
+bool canmapping_map_value(float *value, const CAN_msg *can_msg, const CANMapping *mapping);
 
 /**
  * apply the mapping's formula against the specified value
@@ -36,7 +36,7 @@ bool canmapping_map_value(float *value, CAN_msg *can_msg, CANMapping *mapping);
  * @param the mapping containing the formula
  * @return the value after transformed by the formula
  */
-float canmapping_apply_formula(float value, CANMapping *mapping);
+float canmapping_apply_formula(float value, const CANMapping *mapping);
 
 /**
  * extract the raw value using the specified mapping
@@ -44,7 +44,7 @@ float canmapping_apply_formula(float value, CANMapping *mapping);
  * @param the mapping that specifies the offset and bit length within the raw data
  * @return the extracted value
  */
-float canmapping_extract_value(uint64_t raw_data, CANMapping *mapping);
+float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping);
 
 CPP_GUARD_END
 #endif /* CAN_MAPPING_H_ */
--- a/include/CAN/can_mapping.h
+++ b/include/CAN/can_mapping.h
@@ -1,0 +1,17 @@
+/*
+ * can_mapping.h
+ *
+ *  Created on: Feb 19, 2017
+ *      Author: brent
+ */
+
+#ifndef CAN_MAPPING_H_
+#define CAN_MAPPING_H_
+
+#include "loggerConfig.h"
+#include "CAN.h"
+#include <byteswap.h>
+
+float map_value(CAN_msg *can_msg, CANMapping *mapping);
+
+#endif /* CAN_MAPPING_H_ */
--- a/include/CAN/can_mapping.h
+++ b/include/CAN/can_mapping.h
@@ -13,9 +13,12 @@
 
 CPP_GUARD_BEGIN
 
+
 float map_value(CAN_msg *can_msg, CANMapping *mapping);
 
-float foo(void);
+float apply_formula(float value, CANMapping *mapping);
+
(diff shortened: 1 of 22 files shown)


include/CAN/can_mapping.h
```diff
@@ -0,0 +1,50 @@
...
+bool canmapping_match_id(CAN_msg *can_msg, CANMapping *mapping);
+
+/**
+ * performs the full mapping against for the CAN message using the CAN mapping
+ * @param value the mapped value is set in this parameter if the CAN mapping matches the specified message
+ * @param can_msg the CAN message containing the raw data
+ * @param mapping the mapping to be applied to the message
+ * @return true if the mapping was successfully applied
+ */
+bool canmapping_map_value(float *value, CAN_msg *can_msg, CANMapping *mapping);
```
brentpicasso: +1.   my achilles heel!

include/util/byteswap.h
```diff
@@ -28,18 +28,30 @@
...
+/*
+ * Byte swap unsigned 16 bits
+ */
 uint16_t swap_uint16( uint16_t val );
 
-//! Byte swap short
+/*
+ * Byte swap signed 16 bits
+ */
 int16_t swap_int16( int16_t val );
```
brentpicasso: we actually had it before, I re-organized the code and updated the documentation, which made it look like new code. :)

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
+/*
+ * Race Capture Firmware
+ *
+ * Copyright (C) 2016 Autosport Labs
```
brentpicasso: probably need to sweep through and update all copyrights

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
+ */
+#include "can_mapping.h"
+#include "byteswap.h"
+#include <byteswap.h>
+
```
brentpicasso: fixing.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+#include <stdio.h>
+
+float canmapping_extract_value(uint64_t raw_data, CANMapping *mapping)
```
brentpicasso: fixing. thank you.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+{
+	uint8_t offset = mapping->offset;
+	uint8_t length = mapping->length;
+	if (! mapping->bit_mode) {
```
brentpicasso: Customers need to express it both ways, but most of the time they're thinking in whole byte offset and lengths, and it's not reasonable to force them to think in bits. So, we need the flexibility to express it both ways. Happy to chat voice about it next opportunity.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+	            case 2:
+	                raw_value = swap_uint16(raw_value);
+	                break;
+	            case 3:
+	                raw_value = swap_uint24(raw_value);
+	                break;
+	            case 4:
+	                raw_value = swap_uint32(raw_value);
+	                break;
+	            default:
```
brentpicasso: case 1 is one byte, nothing to swap.  we only support mapping lengths of 1-4 bytes, so this covers everything.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+	                break;
+	        }
+	}
+	return (float)raw_value;
+}
+
+float canmapping_apply_formula(float value, CANMapping *mapping)
+{
+		value *= mapping->multiplier;
+		if (mapping->divider)
```
brentpicasso: I agree with you on principal. However, industry standard CAN mapping schemes provides for a somewhat redundant divider. Explained this in the last PR, you might've missed it.   Overall, CAN mapping standards includes a divider, it's an industry thing. 

I didn't have it initially, and I initially resisted including it, but after reading and seeing different CAN mapping schemes for devices out there, I included it after allo.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+				value /= mapping->divider;
+		value += mapping->adder;
+		return value;
```
brentpicasso: Yep, I separated it out b/c of the optional divider. I'm pretty confident the compiler will optimize things.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+}
+
+bool canmapping_match_id(CAN_msg *can_msg, CANMapping *mapping)
+{
+    uint32_t can_id  = can_msg->addressValue;
+    if (mapping->can_mask)
```
brentpicasso: I thought of that. using zero IMO is the more common "value not defined" that people would identify. and yes, zero isn't a useful number for masking, so that's why I chose it.

src/CAN/can_mapping.c
```diff
@@ -0,0 +1,82 @@
...
+            can_id &= mapping->can_mask;
+
+    return can_id == mapping->can_id;
+}
+
+bool canmapping_map_value(float *value, CAN_msg *can_msg, CANMapping *mapping)
+{
+        if (! canmapping_match_id(can_msg, mapping))
```
brentpicasso: fixing.

src/util/byteswap.c
```diff
@@ -23,26 +23,28 @@
...
-//! Byte swap short
 int16_t swap_int16( int16_t val )
 {
     return (val << 8) | ((val >> 8) & 0xFF);
 }
 
-//! Byte swap unsigned int
+uint32_t swap_uint24( int32_t val )
+{
+    val = ((val & 0xFF0000) >> 16) | (val & 0x00FF00) | ((val & 0x0000FF) << 16);
```
brentpicasso: I could do that

brentpicasso: I collapsed the two statements. I didn't make the parameter const since it's passing by value anyway.

brentpicasso:
Thanks for the review. Applied suggested changes and merging.

PR #910, opened 2017-03-28
add units converter for CAN and OBDII channels

brentpicasso:


include/units/units_conversion.h
```diff
@@ -0,0 +1,33 @@
...
+#ifndef UNITS_CONVERSION_H_
+#define UNITS_CONVERSION_H_
+
+/**
+ * Perform a units conversion for the specified units conversion id
+ * @param id the units conversion id
+ * @param value the value to convert
+ * @return the converted value if the id is a valid conversion id; if not, the same value is returned (no conversion)
+ **/
+float convert_units(int id, const float value);
```
stieg: Seems like this id should be an enum instead.

src/units/units_conversion.c
```diff
@@ -0,0 +1,147 @@
...
+		return value;
+}
+static float c_to_f(float value)
+{
+		return  value * 1.8 + 32;
+}
+
+static float f_to_c(float value)
+{
+		return (value - 32) / 1.8;
```
stieg: A thought here: since div is much slower than mult consider making this `(value - 32) * 0.555555556`

brentpicasso: :+1:

brentpicasso:
Thanks for the quick review. I can define the conversion IDs in enums, even though they won't be referenced anywhere in the firmware - these are IDs the app provides directly.

brentpicasso:
@stieg  - enums added for units conversion

stieg:
Nice.  Looks much improved with the use of the enum.  A more pedantic me might suggest to break each conversion category into its own .c and .h file and use the same method signature to do the conversion (or something similar), but that may also introduce more trouble than its worth.

PR #911, opened 2017-04-02
Enable types handling in can mapping

brentpicasso:
Enables type handling for CAN mapping


src/CAN/can_mapping.c
```diff
@@ -47,7 +49,30 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
...
+        		}
+        		else if (mapping->length <= 16){
+        			extracted_value = *((int16_t*)&raw_value);
+        		}
+        		else {
+        			extracted_value = *((int32_t*)&raw_value);
+        		}
+        		break;
+        	case CANMappingType_unsigned:
+        	default:
```
stieg: This seems dubious.  Since this is an enum you should not have a default case here.  You want to handle it this way because if you add a new enum in the future you want the compiler to tell you that you have forgotten a case in a case statement like this.  This is a benefit of using enums with gcc.  So get rid of the default and handle all specific types.

src/CAN/can_mapping.c
```diff
@@ -32,6 +32,8 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
         }
         uint32_t bitmask = (1UL << length) - 1;
         uint32_t raw_value = (raw_data >> offset) & bitmask;
+
+        /* normalize endian */
         if (mapping->big_endian) {
                 switch (mapping->length) {
```
stieg: I should have caught this before but I missed it.  If you use a `default` value here it should lead to an `UNREACHABLE` assertion only.  Why?  Because you don't support more than 4 bytes.  Fail fast this way instead of silently masking the error.

src/CAN/can_mapping.c
```diff
@@ -32,6 +32,8 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
         }
         uint32_t bitmask = (1UL << length) - 1;
         uint32_t raw_value = (raw_data >> offset) & bitmask;
+
+        /* normalize endian */
```
stieg: Again this always makes the assumption you are on a little endian system.  This is potentially dangerous.  Do you at least have a check to ensure that we are a little endian system and blow up if we are not?  Would be wise.

src/CAN/can_mapping.c
```diff
@@ -47,7 +49,30 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
...
                 }
         }
-        return (float)raw_value;
+
+        /* convert type */
+        float extracted_value = 0;
+        switch (mapping->type) {
+        	case CANMappingType_IEEE754:
+        		extracted_value = *((float*)&raw_value);
+        		break;
```
stieg: The `extraced_value` variable is totally unnecessary.  Remove it and just direct return.

src/CAN/can_mapping.c
```diff
@@ -47,7 +49,30 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
...
+        	case CANMappingType_signed:
+        		if (mapping->length <= 8) {
+        			extracted_value = *((int8_t*)&raw_value);
```
stieg: Umm... So I am super confused about what is happening here.  Raw value is an 8 bit float pointer?  Why are you using a `uint32_t` to represent an address that may not even be 32 bits in length?  That is very unportable.  Should probably be a `void*` since that will adjust with varying address sizes.  And even if you use that the conversion here seems super off.  You are stuffing a signed int like value into a float!?!  Something ain't stirring the Koolaid well here.

jlwall: I had made the initial request to Brent on this. Ideally you should strive to support U1 to U32's and S2 to S32's on all signals, to stay bit atomic. On OEM cans' there are items like S7's and S9's that would not completely work here.

Currently before this commit there was no int8_t support of variables and this is a step forward. Perhaps a fully agree'd upon functional requirement and then Test case would be the best place to start, to allow the code to follow

src/CAN/can_mapping.c
brentpicasso: Agree on handling all allowed cases explicitly. 

Are you suggesting a panic(PANIC_CAUSE_UNREACHABLE);  if >4 for length?

src/CAN/can_mapping.c
brentpicasso: @stieg  this uses C type punning to coerce the unsigned value into the target type. The cast to the final (normalized) float type is implicit; I agree it could be clearer - e.g.

extracted_value = (float)*((int16_t*)&raw_value);

src/CAN/can_mapping.c
brentpicasso: agree. will address.

src/CAN/can_mapping.c
brentpicasso: A runtime check for endian-ness would be useful. We can schedule this work if we ever switch to a different processor architecture.

src/CAN/can_mapping.c
brentpicasso: Thanks, @jlwall .  We haven't seen examples (yet) in the common ECU mappings we already support - thankfully they're pretty simple U8's and U16 values. From your experience, what cases will we expect to see S7's and S9's ? 

However, for BMW's E46 steering angle sensor we're going to need a new data type to handle the sign-magnitude approach they're using - it's not a conventional signed 16 bit number.

src/CAN/can_mapping.c
brentpicasso: Agreed. updated.

src/CAN/can_mapping.c
brentpicasso: applied feedback with unreachable panic.

src/CAN/can_mapping.c
brentpicasso: addressed with panic / unreachable.

brentpicasso:
applied review feedback, as well as handling a new type for handling sign-magnitude numbers as seen on bmw E46 steering angle sensors.

src/CAN/can_mapping.c
stieg: Honestly I would do it now rather than later.  You will totally forget to do it later and the bug will get lost in the noise.  Should be pretty simple to do.  Do something like:
```C
volatile int x = 0xA5;
const char* c = (void*) &x;
switch(*c) {
case 0xA5: return LITTLE_ENDIAN;
case 0x00 : return BIG_ENDIAN;
default; ASSERT(UNREACHABLE);
};
```

src/CAN/can_mapping.c
stieg: Yes.  If you should never get there, we should panic if we do.  Means something is very wrong.

include/logger/loggerConfig.h
```diff
@@ -299,6 +299,7 @@ enum CANMappingType {
     CANMappingType_unsigned = 0,
     CANMappingType_signed,
     CANMappingType_IEEE754,
+    CANMappingType_sign_magnitude,
```
stieg: What does this mean?  Perhaps a comment?

src/CAN/can_mapping.c
```diff
@@ -44,10 +49,52 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
                         raw_value = swap_uint32(raw_value);
                         break;
                     default:
+                        /* Should never be mapping anything larger than 4 bytes */
+                        panic(PANIC_CAUSE_UNREACHABLE);
                         break;
```
stieg: Unnecessary :)

src/CAN/can_mapping.c
```diff
@@ -44,10 +49,52 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
...
+
+        /* convert type */
+        switch (mapping->type) {
+            case CANMappingType_unsigned:
+                    return (float)raw_value;
+        	case CANMappingType_signed:
+                    if (length <= 8) {
+                        return (float)*((int8_t*)&raw_value);
+                    }
+                    else if (length <= 16){
```
stieg: NIT: the `else if` statements here are totally redundant because you are doing direct returns from within the if statements.

src/CAN/can_mapping.c
```diff
@@ -44,10 +49,52 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
...
+                    else {
+                        return (float)*((int32_t*)&raw_value);
+                    }
+            case CANMappingType_IEEE754:
+                    return *((float*)&raw_value);
+            case CANMappingType_sign_magnitude:
+                    /**
+                     *  sign-magnitude is used in cases where there's a sign bit
+                     *  and an absolute value indicating magnitude.
+                     *  e.g. BMW E46 steering angle sensor
```
stieg: F***ing BMWs :+1:

src/CAN/can_mapping.c
brentpicasso: Thanks. if we do it now, I'd rather put it in capabilities.h as a #define . A convention also used in freeRTOS and ChibiOS and I don't think it's a bad approach.

src/CAN/can_mapping.c
brentpicasso: Latest code now has this panic.

src/CAN/can_mapping.c
```diff
@@ -44,10 +49,52 @@ float canmapping_extract_value(uint64_t raw_data, const CANMapping *mapping)
...
+                     **/
+                    {
+                            uint32_t sign;
+                            if (length <= 8) {
+                                    sign = 1 << 7;
```
stieg: Err.... so what happens if length = 7?  Would that yield a valid answer or no?  If its not a possible case you should use a switch statement here.  If it is a possibel case consider using something like:
```c
sign = 1 << (length - 1);
```
Or if I'm totally off-base then yeah fix as needed

brentpicasso: The initial idea was to  'snap' the data size to either 8, 16 or 32 bit sign-magnitude numbers in an effort to keep things sane.  Let me check.

brentpicasso: Ok, simplified algorithm where we just calculate the sign based on the length of the mapping.  it works.  Thanks!

case CANMappingType_sign_magnitude:
                    /**
                     *  sign-magnitude is used in cases where there's a sign bit
                     *  and an absolute value indicating magnitude.
                     *  e.g. BMW E46 steering angle sensor
                     **/
                    {
                            uint32_t sign = 1 << (length - 1);
                            return raw_value < sign ? (float)raw_value : -(float)(raw_value & (sign - 1));
                    }

src/CAN/can_mapping.c
brentpicasso: agreed. fixed.

src/CAN/can_mapping.c
brentpicasso: fixed.

PR #921, opened 2017-04-26
Added hal support for timer_device_get_count and timer_device_reset_count.

ddimensia:
Added a 32 bit pulse_count to the channel state for PWM input channels.  The pulse count is updated whenever the a rising edge is detected on the hardware timer.  If the pulse_count overflows it will simply start back at 0.

This change will allow the Lua script functions getTimerCount and resetTimerCount to function for mk2 and mk3 hardware.  The change was verified on a mk2 device with a 1khz pulsed input and lua script capturing and resetting the pulse count.

### DIFF
--- a/platform/mk2/hal/timer_stm32/timer_device_stm32.c
+++ b/platform/mk2/hal/timer_stm32/timer_device_stm32.c
@@ -48,6 +48,7 @@ static struct state {
         uint16_t period;
         uint16_t duty_cycle;
         uint16_t q_period_ticks;
+        uint32_t pulse_count;
 } g_state[MK2_TIMER_CHANNELS];
 
 static struct config {
@@ -340,11 +341,18 @@ bool timer_device_init(const size_t chan, const uint32_t speed,
         return false;
 }
 
-void timer_device_reset_count(size_t chan) {}
+void timer_device_reset_count(size_t chan)
+{
+        if (chan >= MK2_TIMER_CHANNELS)
+                return;
+
+        struct state *s = &g_state[chan];
+        s->pulse_count = 0;
+}
 
 uint32_t timer_device_get_count(size_t chan)
 {
-        return 0;
+        return chan < MK2_TIMER_CHANNELS ? g_state[chan].pulse_count : 0;
 }
 
 static uint32_t ticks_to_us(const size_t chan, const uint16_t ticks)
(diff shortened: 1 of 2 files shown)


platform/mk2/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -48,6 +48,7 @@ static struct state {
         uint16_t period;
         uint16_t duty_cycle;
         uint16_t q_period_ticks;
+        uint16_t pulse_count;
```
brentpicasso: why not uint32_t?

ddimensia: No good reason, I guess I thought I was saving memory or something.  I'll make the change and update the PR.

ddimensia:
Updated to use a uint32 for the pulse counter.  This makes it extremely unlikely to overflow for normal use cases.

platform/mk2/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -340,11 +341,17 @@ bool timer_device_init(const size_t chan, const uint32_t speed,
         return false;
 }
 
-void timer_device_reset_count(size_t chan) {}
+void timer_device_reset_count(size_t chan)
```
stieg: make this const and normally we prefer something like `if (chan >= MK2_TIMER_CHANNELS) return;` as this avoids indentation and is easier to expand upon.

stieg: Also this should be a static method if it isn't exposed in the headers

platform/mk2/hal/timer_stm32/timer_device_stm32.c
```diff
@@ -403,6 +410,7 @@ static void update_device_state(const size_t chan, const uint16_t p_ticks,
         s->period = total_ticks;
         s->duty_cycle = 100 * (uint32_t) (h_ticks / total_ticks);
         s->q_period_ticks = 0;
+        s->pulse_count += 1;
```
stieg: Use ++ here as this is more efficient and will not cause a read modify write instruction series.

stieg:
Jenkins: Ok to test

ddimensia:
@stieg The timer_device_reset_count method is defined in include/timer/timer_device.h and is called through Lua so shouldn't be static.  I didn't make the arg in that function const to keep it consistent with the header.  

I addressed your other comments.

brentpicasso:
resolves #925

PR #934, opened 2017-06-30
enable auto-detection of 29 bit OBDII. resolves #890

### LINKED ISSUE #890: 29 / 11 bit OBDII support
Some vehicles, like certain honda products only support 29 bit CAN OBDII. need to support this.

brentpicasso:
Enable auto-detection of 29 or 11 bit OBDII protocol.

### DIFF
--- a/include/CAN/can_mapping.h
+++ b/include/CAN/can_mapping.h
@@ -31,7 +31,7 @@ CPP_GUARD_BEGIN
  * match the can message based on the specified CAN mapping ID and ID mask
  * @param can_msg the CAN message to test
  * @param mapping the can mapping to check against
- * @return true if the CAN message matches the mapping's CAN ID and ID mask
+ * @return true if the CAN message matches the mapping's CAN ID and ID mask, or true if mapping's CAN ID is 0
  */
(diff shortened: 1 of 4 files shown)


include/CAN/can_mapping.h
```diff
@@ -31,7 +31,7 @@ CPP_GUARD_BEGIN
  * match the can message based on the specified CAN mapping ID and ID mask
  * @param can_msg the CAN message to test
  * @param mapping the can mapping to check against
- * @return true if the CAN message matches the mapping's CAN ID and ID mask
+ * @return true if the CAN message matches the mapping's CAN ID and ID mask, or true if mapping's CAN ID is 0
```
stieg: Why 0?  What is special about that?  Does this not break behavior of the function?

brentpicasso: I changed the API so that a CAN ID of 0 means "match any CAN ID".

brentpicasso: (like a wildcard)

brentpicasso: Ok, enhanced documentation.

Issue #904, opened 2017-02-17
Switch to newlib nano toolchain


brentpicasso:
Memory savings will be dramatic, especially around small RAM / Flash sized devices:

https://github.com/32bitmicro/newlib-nano-1.0

brentpicasso:
Main project page with link to latest 4.x released is here; https://launchpad.net/gcc-arm-embedded

brentpicasso:
Evaluating gcc 4.9 with newlib nano enabled:

current 4.7 gcc (gcc-arm-none-eabi-4_7-2013q3)

brent@brent-desktop:~/git-projects/RaceCapture-Pro_firmware$ arm-none-eabi-size platform/rct/main.elf 
   text	   data	    bss	    dec	    hex	filename
 172400	   1852	   9572	 183824	  2ce10	platform/rct/main.elf



4.9 gcc (gcc-arm-none-eabi-4_9-2014q4), without newlib nano

brent@brent-desktop:~/git-projects/RaceCapture-Pro_firmware/platform/rct$ arm-none-eabi-size main.elf 
   text	   data	    bss	    dec	    hex	filename
 163216	   1812	   9316	 174344	  2a908	main.elf



4.9 gcc (gcc-arm-none-eabi-4_9-2014q4), with newlib nano enabled (--specs=nano.specs)

brent@brent-desktop:~/git-projects/RaceCapture-Pro_firmware$ arm-none-eabi-size platform/rct/main.elf 
   text	   data	    bss	    dec	    hex	filename
 151472	    840	   9316	 161628	  2775c	platform/rct/main.elf

PR #937, opened 2017-07-03
Memory savings and OBDII / CAN channels expansion

### LINKED ISSUE #936: Increase number of available OBDII channels and CAN mappings by platform
RCT: add more OBDII and CAN mapping channels
RCP MK2, MK3 and Apex: Increase OBDII channels

brentpicasso:
* switch to gcc 4.9 with newlib nano for memory savings. 
* move select things from un-initialized BSS to initialized DATA segments by providing a compiler hint by initializing the selected structs to {0}
* increase OBDII and CAN channels for RCT, increase OBDII channels for RCP. 

Note, memory savings wasn't needed for RCP OBDII channels increase, just bundled it together. 

Resolves #936 #904

### DIFF
--- a/platform/rct/Makefile
+++ b/platform/rct/Makefile
@@ -273,7 +273,7 @@ CPU_FLAGS := \
 -mthumb \
 -mfloat-abi=soft \
 
-CFLAGS := -DUSE_STDPERIPH_DRIVER -DSTM32F30X -MD -c
+CFLAGS := -DUSE_STDPERIPH_DRIVER -DSTM32F30X -MD -c 
 CFLAGS += $(CPU_FLAGS)
 CFLAGS += $(ASL_CFLAGS)
 CFLAGS += $(BOARD_DEFINES)
@@ -281,7 +281,7 @@ CFLAGS += $(addprefix -I,$(inc-y))
 
 LDFLAGS := $(CPU_FLAGS)
 LDFLAGS += -TProject/STM32_FLASH.ld
-LDFLAGS += -Wl,-cref,-u,Reset_Handler,-Map=main.map
+LDFLAGS += -Wl,-cref,-u,Reset_Handler,-Map=main.map --specs=nano.specs
 LDFLAGS += -Wl,--gc-sections,-static
 LDFLAGS += $(FREERTOS_LD_FLAGS)
 
--- a/platform/rct/Project/STM32_FLASH.ld
+++ b/platform/rct/Project/STM32_FLASH.ld
@@ -36,8 +36,8 @@ ENTRY(Reset_Handler)
 _estack = 0x20000000 + 40K;    /* end of 40K RAM */
 
 /* Generate a link error if heap and stack don't fit into RAM */
-_Min_Heap_Size = 0x200;      /* required amount of heap  */
-_Min_Stack_Size = 0x400; 	/* required amount of stack */
+_Min_Heap_Size = 0x100;      /* required amount of heap  */
+_Min_Stack_Size = 0x100; 	/* required amount of stack */
 
 /* Specify the memory areas */
 MEMORY
--- a/platform/rct/hal/imu_stm32/imu_device_stm32.c
+++ b/platform/rct/hal/imu_stm32/imu_device_stm32.c
@@ -39,7 +39,7 @@
 #define IMU_TASK_PRIORITY	(tskIDLE_PRIORITY + 2)
 #define IS_9150_ADDR            0x68
 
-static struct is9150_all_sensor_data sensor_data[2];
+static struct is9150_all_sensor_data sensor_data[2] = {0};
 static struct is9150_all_sensor_data *read_buf = &sensor_data[0];
 static struct is9150_all_sensor_data *fill_buf = &sensor_data[1];
 static enum imu_init_status init_status;
--- a/README.md
+++ b/README.md
@@ -62,8 +62,8 @@ signal gain in wooded areas).
 
 ### Toolchain Setup
 Do the following to setup the MK2 toolchain:
-* Download [the official MK2
-  toolchain](https://launchpad.net/gcc-arm-embedded/4.7/4.7-2013-q3-update/)
+* Download [the official
+  toolchain](https://launchpad.net/gcc-arm-embedded/+milestone/4.9-2014-q4-major)
 * Extract the tarball to a directory of your choice
 * Add the 'bin' directory from within the newly extracted directory to
   your system PATH.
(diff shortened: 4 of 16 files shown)


stieg:
BSS is always initialized to 0 per C compiler spec.  See https://en.wikipedia.org/wiki/.bss .  But this is just me replying to your description above... starting the review now...

platform/rct/Makefile
```diff
@@ -273,15 +273,15 @@ CPU_FLAGS := \
 -mthumb \
 -mfloat-abi=soft \
 
-CFLAGS := -DUSE_STDPERIPH_DRIVER -DSTM32F30X -MD -c
+CFLAGS := -DUSE_STDPERIPH_DRIVER -DSTM32F30X -MD -c
```
stieg: clean yer whitespace you dirty hippy.  applies throughout.

platform/rct/Project/STM32_FLASH.ld
```diff
@@ -36,8 +36,8 @@ ENTRY(Reset_Handler)
 _estack = 0x20000000 + 40K;    /* end of 40K RAM */
 
 /* Generate a link error if heap and stack don't fit into RAM */
-_Min_Heap_Size = 0x200;      /* required amount of heap  */
-_Min_Stack_Size = 0x400; 	/* required amount of stack */
+_Min_Heap_Size = 0x100;      /* required amount of heap  */
+_Min_Stack_Size = 0x100; 	/* required amount of stack */
```
stieg: These are pretty small values.  Have you load tested this?

platform/rct/hal/imu_stm32/imu_device_stm32.c
```diff
@@ -39,7 +39,7 @@
 #define IMU_TASK_PRIORITY	(tskIDLE_PRIORITY + 2)
 #define IS_9150_ADDR            0x68
 
-static struct is9150_all_sensor_data sensor_data[2];
+static struct is9150_all_sensor_data sensor_data[2] = {0};
```
stieg: Why?  What is gained here by doing this (other than migrating from .bss to .data).

stieg: Same thing applies throughout.

brentpicasso:
Actually,  it's not it's initialized to 0 that matters, it's that it's
initialized in user code that determines what section it goes into.  And on
RCT, bss and data are in different ram regions on the processor.  I'm
moving stuff to make room for a bigger logger config struct.

Brent Picasso
Autosport Labs

brentpicasso:
So if you look at STM32_FLASH.ld you'll see that BSS is in the 8K core-coupled memory region, and HEAP, DATA are in the 48K main RAM segment. These segments are not contiguous, but in separate address spaces. This is the design of the STM32F303 processor we have in RCT. 

Per GCC, BSS is for un-initialized data, and DATA is for initialized data. 

We need more room in BSS for a larger LoggerConfig struct, so we can add more CAN and OBDII channels. 

By explicitly initializing variables to a value  - {0} in this case - it forces GCC to move these variables into the DATA segment, where we can make room for a larger LoggerConfig. 

If the {0} initializer consumes more memory, it is in the flash region where the effect is negligible. RAM is not impacted.

Thanks, hope this helps clarify things!

platform/rct/Project/STM32_FLASH.ld
brentpicasso: This is the stack and "heap" (such as it is) used by main() before freeRTOS fires up.  Once the scheduler is started, we don't need it anymore.  See: http://www.freertos.org/FAQMem.html#RAMReduce

I put "heap" in quotes since it's not really used as heap with a heap manager.  In fact, these two variables are only used further down in the linker script to ensure there's enough space left over, and that's calculated as _Min_Heap_Size + _Min_Stack_Size. 

Net-net, main() gets 0x100 + 0x100 (512 bytes) for a stack upon startup, which is never used again after the scheduler starts.

platform/rct/hal/imu_stm32/imu_device_stm32.c
brentpicasso: The gain is to make room for a larger LoggerConfig object in BSS (described in the top level comment).

stieg: I get the premise of the logic but I disagree with the implementation.  Just because you specify initialized data here does not in fact always mean that the compiler will keep this data in the .data section.  IIRC there is a provision in the C standard that says if the code is initialized to all zeros its at the compilers discretion as to whether the code goes to .bss or to .data.  If you need more .bss space then why not adjust memory maps to allocate some more space (I believe you can have more than one .bss section)?  My concern here is that you are making the firmware unnecessarily larger and making it less portable by doing this instead of adjusting memory allocations where necessary.

brentpicasso: Ok -  if GCC will automatically juggle memory to fit in all available dis-contiguous BSS segments then I will give that a shot.  thanks!

brentpicasso: So, research tells me that ld does not have the ability to distribute objects between multiple bss sections. Apparently some commercial linkers can do this, but not GNU's ld. 

The links below are my research - first 3 links have the strongest discussions around this; bottom links to documentation does not show any information about combining multiple regions. 

At the moment, I think that the {0} initializer hint is the most portable solution to achieving the goals we want.  When we pull the predictive timer out of RaceCapture/Track's firmware and move it into the app, we'll recover a big chunk of RAM and might be able to unwind most if not all of this. 
-----------------------------------------------
https://community.arm.com/iot/embedded/f/discussions/6266/linker-instruction-to-split-bss-section-into-two-different-ram-banks

https://mcuoneclipse.com/2013/04/14/text-data-and-bss-code-and-data-size-explained/#comment-72754

https://e2e.ti.com/support/development_tools/compiler/f/343/t/104507
https://community.nxp.com/thread/426299


https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/Using_ld_the_GNU_Linker/sections.html#OUTPUT-SECTION-REGION

ftp://ftp.gnu.org/old-gnu/Manuals/ld-2.9.1/html_mono/ld.html#SEC19

platform/rct/Project/STM32_FLASH.ld
stieg: Ahh... sorry misinterpreted the variable names.  You are right and this is sane.  I wish we could recover all of that space.  Not sure how we would go about doing that but there is probably a way...

platform/rct/hal/imu_stm32/imu_device_stm32.c
stieg: Bummer.  Just did some extra digging and came to the same conclusion.  Oh well.  I probably would have used some form of label to get it into the data section as opposed to initializing it (just to keep firmware size small) but I don't suppose it much matters in the end.  Good sleuthing for a proper solution that doesn't exist.

brentpicasso: Thanks too for pushing for the best possible solution! Yeah, the {0} seemed like the least offensive option. I thought of a memory segment attribute, like we have on the loggerConfig flash memory block, but that would make it even less portable.

PR #938, opened 2017-07-03
Enable 32 bit PIDs, and refactor duplicate ChannelConfig (!!)

### LINKED ISSUE #924: support 32 bit PID requests
Need to support 32 bit PID requests as implemented by some manufacturers.

### LINKED ISSUE #936: Increase number of available OBDII channels and CAN mappings by platform

brentpicasso:
This enables 32 bit PIDs and also includes a refactoring for some major memory savings. 

Turns out for each OBDII channel we had an extra ChannelConfig defined in the PidConfig, but not used. 

Refactor to use the ChannelConfig that's in the CANMapping object. 

Resolves #924 and helps memory situation to resolve #936

### DIFF
--- a/src/OBD2/OBD2.c
+++ b/src/OBD2/OBD2.c
@@ -180,7 +180,7 @@ bool OBD2_init_current_values(OBD2Config *obd2_config)
         size_t max_sample_rate = 0;
         for (size_t i = 0; i < obd2_channel_count; i++) {
                 max_sample_rate = MAX(max_sample_rate,
-                                      decodeSampleRate(obd2_config->pids[i].cfg.sampleRate));
+                                      decodeSampleRate(obd2_config->pids[i].mapping.channel_cfg.sampleRate));
         }
         obd2_state.max_sample_rate = max_sample_rate;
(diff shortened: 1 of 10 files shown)


src/OBD2/OBD2.c
```diff
@@ -136,23 +136,33 @@ bool OBD2_is_state_stale(void)
...
+                if (pid > 0xFFFF) {
+                        msg.data[0] = 5;
+                        msg.data[2] = (pid >> 24);
+                        msg.data[3] = (pid >> 16) & 0xFF;
+                        msg.data[4] = (pid >> 8) & 0xFF;
+                        msg.data[5] = pid & 0xFF;
+                }
+                else {
+                        msg.data[0] = 3;
+                        msg.data[2] = (pid >> 8) & 0xFF;
```
stieg: Is the masking necessary if its a 16 bit value?

brentpicasso: nope! the if statement above takes care of that. I guess I was just extra cautious since it's a uint32_t.  Will fix.

PR #943, opened 2017-10-03
WiFi camera control, WIP

brentpicasso:
This is part 1 of wifi camera control - updating the wifi driver to accommodate calls to the camera device. 

Part 2 will involve the configuration and making the camera control more modular to handle different brands / models of cameras. 

TODO sections highlight future work that will need to be handled.

### DIFF
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -1026,8 +1026,10 @@ static bool send_data_cb(struct at_rsp *rsp, void *up)
 	}
         case AT_RSP_STATUS_SEND_OK:
                 /* Then we have successfully sent the message */
-		status = true;
+                status = true;
                 goto fini;
+        case AT_RSP_STATUS_BUSY:
+                pr_warning(LOG_PFX "Got busy response!");
         default:
                 /* Then bad things happened */
                 cmd_failure("send_data_cb", "Bad response value");
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -109,11 +109,11 @@ static void ipd_urc_cb(struct at_info *ati, char *msg)
 	 * comes from the command messages.
 	 */
 	if (serial_len < len) {
-	    pr_info(LOG_PFX "Received less data than expected: ");
-	    pr_info_int(serial_len);
-	    pr_info("/");
-	    pr_info_int(len);
-	    pr_info_str_msg("; data: ", data);
+	    pr_debug(LOG_PFX "Received less data than expected: ");
+	    pr_debug_int(serial_len);
+	    pr_debug("/");
+	    pr_debug_int(len);
+	    pr_debug_str_msg("; data: ", data);
 	}
 
 	data[len] = 0;
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -366,8 +366,6 @@ static bool init_soft_reset_cb(struct at_rsp *rsp, void *up)
 
 	/* Reset Serial to default values first */
 	struct Serial* serial = state.ati->sb->serial;
-	serial_clear(serial);
-	esp8266_set_default_serial_params(serial);
 
 	/* Wait for the ready notice from the modem and ping test. */
 	esp8266_wait_for_ready(serial);
@@ -399,6 +397,15 @@ static bool init_soft_reset_cb(struct at_rsp *rsp, void *up)
  */
 static bool init_soft_reset()
 {
+    /* Initialize the esp8266 hardware */
+
+        wifi_device_init();
+        struct Serial* serial = state.ati->sb->serial;
+        serial_clear(serial);
+        esp8266_set_default_serial_params(serial);
+
+
+
         const char cmd[] = "AT+RST";
         return NULL != at_put_cmd(state.ati, cmd, _TIMEOUT_LONG_MS,
                                   init_soft_reset_cb, NULL);
@@ -1073,7 +1080,7 @@ bool esp8266_send_data(const unsigned int chan_id, struct Serial *serial,
         char cmd[32];
         snprintf(cmd, ARRAY_LEN(cmd),"AT+CIPSEND=%d,%d", chan_id, (int) len);
 
-        return NULL != at_put_cmd(state.ati, cmd, _TIMEOUT_LONG_MS,
+        return NULL != at_put_cmd(state.ati, cmd, _TIMEOUT_SUPER_MS,
                                   send_data_cb, ti);
 }
 
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -109,22 +109,13 @@ static void ipd_urc_cb(struct at_info *ati, char *msg)
 	 * comes from the command messages.
 	 */
 	if (serial_len < len) {
-
-		/* There's more data coming */
-		while (serial_len < len) {
-		        /* reset pointer so we have a continuous string */
-		        ati->sb->curr_len--;
-		        char *msg = serial_buffer_rx(ati->sb, 1);
-                serial_len = strlen(data);
-		        if (strlen(msg) == 0)
-		                break;
-		}
-        pr_debug_int_msg(LOG_PFX "data received len: ", serial_len);
-        pr_debug("'");
-        pr_debug(data);
-        pr_debug("'\r\n");
-
+	    pr_info(LOG_PFX "Received less data than expected: ");
+	    pr_info_int(serial_len);
+	    pr_info("/");
+	    pr_info_int(len);
+	    pr_info_str_msg("; data: ", data);
 	}
+
 	data[len] = 0;
 
 
--- a/src/devices/esp8266.c
+++ b/src/devices/esp8266.c
@@ -79,7 +79,7 @@ static void cmd_failure(const char *cmd_name, const char *msg)
 /**
  * This is the callback invoked when an IPD URC is seen.
  */
-static void ipd_urc_cb(char *msg)
+static void ipd_urc_cb(struct at_info *ati, char *msg)
 {
 	if (!state.hooks.data_received_cb)
 		return;
@@ -97,7 +97,7 @@ static void ipd_urc_cb(char *msg)
         const int chan_id = atoi(toks[1]);
         const size_t len = atoi(toks[2]);
         char *data = toks[3];
-	const size_t serial_len = strlen(data);
+        size_t serial_len = strlen(data);
 
 
 	/*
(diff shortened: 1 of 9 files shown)


src/devices/esp8266.c
```diff
@@ -109,11 +109,21 @@ static void ipd_urc_cb(char *msg)
 	 * comes from the command messages.
 	 */
 	if (serial_len < len) {
-		pr_error(LOG_PFX "IPD Length Mismatch.  Dropping. \r\n");
-		pr_info_int_msg(LOG_PFX "Length Expected: ", (int) len);
-		pr_info_int_msg(LOG_PFX "Length Actual: ", (int) serial_len);
-		pr_info_str_msg(LOG_PFX "Data: ", data);
-		return;
+
+		/* There's more data coming */
```
brentpicasso: Rationale: The camera's webserver sends a multi-line response. If there's less data than expected on the initial response, we need to siphon the rest of the data until complete.

src/drivers/esp8266_drv.c
```diff
@@ -736,6 +741,18 @@ static void set_fast_baud()
 	cmd_set_check(CHECK_WIFI_DEVICE);
 }
 
+static void hard_reset_wifi(void)
```
brentpicasso: This fixed a bug where, if we do a hard reset, we don't set our baud rate back to factory default of 115200.  This fixes this.

src/drivers/esp8266_drv.c
```diff
@@ -1408,8 +1426,11 @@ static void connect_cb(const bool status, const bool already_connected)
 				"connected: ", cso->chan_id);
                 channel_close(ch);
         } else if (!status) {
-		pr_info_int_msg(LOG_PFX "Failed to connect channel: ",
+                pr_info_int_msg(LOG_PFX "Failed to connect channel: ",
 				cso->chan_id);
+                esp8266_state.cmd.failures++;
```
brentpicasso: Here, we were not counting connect failures as errors. 

Sometimes our esp8266 does a hard reset, which resets the baud rate, causing future connections to always fail. this will catch it and eventually cause a hard reset.

src/tasks/wifi.c
```diff
@@ -85,8 +85,12 @@ static struct {
         struct {
                 struct Serial* serial;
         } beacon;
+        struct {
+            struct Serial* serial;
+        } camera_control;
```
brentpicasso: Adding a separate struct for tracking the camera control connection.

src/tasks/wifi.c
```diff
@@ -526,23 +630,30 @@ static void _task(void *params)
                                    portMAX_DELAY))
                         continue;
 
+                if (!esp8266_drv_is_initialized())
```
brentpicasso: If we do a hard reset while the system is currently running, existing timers will flood the esp8266 with requests (beacon, camera control, etc). This blocks commands from happening until we're up and running.

brentpicasso:
One thing that is outstanding; the HTTP request that comes back from the gopro is multi-line, which gets interpreted as multiple messages. The state machine chokes on it, but seems let it pass somewhat harmlessly. 

Ideally we'd be able to pull the data into a complete string. What I've found was that the ESP8266 doesn't seem to return the expected number of bytes on multi-line returns. 

I tried some code to pull in the remaining data on the fly, but sometimes it would pull in the "OK" response to another command, throwing the state machine off. 

Example output of HTTP response:

)] TX: AT+CIPSEND=0,109(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: OK(\r)(\n)
[serial (WiFi)] TX: {"beacon":{"name":"RCT","port":7223,"serial":"360026000157335239383620","ip":["10.5.5.109","192.168.4.1"]}}(\r)(\n)
[serial (WiFi)] RX: > (\r)(\n)
[serial (WiFi)] RX: Recv 109 bytes(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: SEND OK(\r)(\n)
[serial (WiFi)] TX: AT+CIPSEND=1,44(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: OK(\r)(\n)
[serial (WiFi)] TX: GET /bacpac/SH?t=12345678&p=%01 HTTP/1.0(\r)(\n)
[serial (WiFi)] TX: (\r)(\n)
[serial (WiFi)] RX: > (\r)(\n)
[serial (WiFi)] RX: Recv 44 bytes(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: SEND OK(\r)(\n)
[serial (WiFi)] TX: AT+CIPSEND=2,512(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: +IPD,1,220:HTTP/1.0 200 OK(\r)(\n)[esp8266] Received less data than expected: 17/220; data: HTTP/1.0 200 OK


[serial (WiFi)] RX: Allow: GET (\r)(\n)
[serial (WiFi)] RX: Accept-Ranges: bytes(\r)(\n)
[serial (WiFi)] RX: Cache-Control: no-cache(\r)(\n)
[serial (WiFi)] RX: Cache-Control: no-store(\r)(\n)
[serial (WiFi)] RX: Connection: Keep-Alive(\r)(\n)
[serial (WiFi)] RX: Server: GoPro Web Server v1.0(\r)(\n)
[serial (WiFi)] RX: Content-Type: application/octet-stream(\r)(\n)
[serial (WiFi)] RX: Content-Length: 2(\r)(\n)
[serial (WiFi)] RX: (\r)(\n)
[serial (WiFi)] RX: (\n)
[serial (WiFi)] RX: OK(\r)(\n)[at] BUG: Received more messages than can handle. Dropping.

[serial (WiFi)] TX: 9,0.01,1.02,0,1,0,507]}}(\r)(\n)
[serial (WiFi)] TX: {"status":{"system":{"model":"RaceCapture","ver_major":2,"ver_minor":12,"ver_bugfix":0,"serial":"360026000157335239)
[serial (WiFi)] TX: {"s":{"t":161000,"d":[161004
[serial (WiFi)] RX: > (\r)(\n)[at] BUG: Received more messages than can handle. Dropping.

[serial (WiFi)] RX: Recv 512 bytes(\r)(\n)[at] BUG: Received more messages than can handle. Dropping.

src/drivers/esp8266_drv.c
stieg: Oh fun.  Stupid ESP8266.

stieg:
Oh hell... the issue seems to be that the ESP8266 isn't escaping \r\n so its getting interpreted as an EOM I bet.

src/drivers/esp8266_drv.c
```diff
@@ -199,6 +199,15 @@ static void cmd_started()
         esp8266_state.cmd.in_progress = true;
 }
 
+static void check_excessive_failures(void)
+{
+        if (esp8266_state.cmd.failures >= CMD_MAX_FAILURES) {
```
stieg: Personal preference here... I always use rapid returns to avoid excess indentation. Also better to use Yoda conditionals to prevent accidental assignments. So instead I would do this:
```
if (CMD_MAX_FAILURES > esp8266_state.cmd.failures ) return;
```

src/drivers/esp8266_drv.c
```diff
@@ -400,7 +405,7 @@ static void socket_disconnect_handler(const size_t chan_id)
 	/* Check if the close op is in progress.  If so, take no action */
 	struct channel_sync_op *cso = &esp8266_state.comm.close_op;
 	if (cso->chan_id == chan_id)
-		return;
+	        return;
```
stieg: Is your editor using spaces or tabs?  I forget which one we use but best to check that its consistent less we mess up the whole damn structure

src/tasks/wifi.c
```diff
@@ -419,6 +437,13 @@ static void process_new_connection(struct Serial *s)
 	esp8266_drv_close(s);
 }
 
+static bool is_valid_ipv4(const char *address)
+{
+        return NULL != address &&
```
stieg: Couldn't this be simplified to `return address && *address && !(STR_EQ(address, IPV4_NO_IP_STR));`

src/tasks/wifi.c
```diff
@@ -491,10 +518,104 @@ static void do_beacon()
...
+    start = !start;
+}
+*/
+
+/* Camera control for Hero 2,3
+ * Format is: GET /bacpac/SH?t=<wifi-password>&p=<shutter-param>
+ */
+static void send_camera_control(struct Serial* serial)
+{
+    const char preamble[] = "GET /bacpac/SH?t=";
```
stieg: Make static so its not allocated every time.  Same with others that don't change.

stieg: Err... not allocated.  I mean not populated every invocation.  Also will save on stack space.

src/tasks/wifi.c
```diff
@@ -491,10 +518,104 @@ static void do_beacon()
...
+    const char epilogue[] = " HTTP/1.0\r\n\r\n";
+    const char shutter_param[] = "&p=";
+    const int command_size = sizeof(preamble) + WIFI_PASSWD_MAX_LEN + sizeof(shutter_param) + 2 + sizeof(epilogue) + 5;
+    char command[command_size];
+
+    strncpy(command, preamble, sizeof(preamble));
```
stieg: Damn I wish `strnprintf` worked.

src/tasks/wifi.c
```diff
@@ -491,10 +518,104 @@ static void do_beacon()
...
+
+    serial_write_s(serial, command);
+    start = !start;
 }
 
+static void do_camera_control()
+{
+        const struct esp8266_ipv4_info* client_ipv4 = get_client_ipv4_info();
+
+        if (! is_valid_ipv4(client_ipv4->address))
```
stieg: Check indentation.  Good use of rapid return.  Put bang next to method call.

src/drivers/esp8266_drv.c
brentpicasso: thanks, applying this change.

src/drivers/esp8266_drv.c
brentpicasso: spaces. I think there's a mix of indentation originally, could use a bulk reformat.

src/tasks/wifi.c
brentpicasso: looks like we're already using snprintf elsewhere in the wifi driver, so I cleaned it up :+1:

src/tasks/wifi.c
brentpicasso: done

PR #946, opened 2017-10-08
Get current sample value by name in Lua

### LINKED ISSUE #939: Expose current direct CAN mapping value in Lua
Similar to readOBD2() in lua, we need a function that exposes the current value for a direct CAN mapping. 

There's no convenient unique ID like we have for readOBD2(), since you can extract multiple values from the same CAN message. 

Some ideas:

* We can do it based on CAN mapping index, but that's pretty fragile and will break the script if the ordering of the CAN mapping changes. 
* We can key it off of the combination of CAN ID / offset / length since those are typically unique, but that would be a bit cumbersome. 
* We can query based on the actual channel name, which would make it work for any channel. This is likely the best option, but it would require an efficient / hashed lookup if requesting by name.

brentpicasso:
Resolves #939 

Extend the current get_channel_value to get the current value for any channel, if parameter is passed as a string.

### DIFF
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -52,11 +52,17 @@
 int g_loggingShouldRun;
 int g_configChanged;
 int g_telemetryBackgroundStreaming;
+struct sample * current_sample = NULL;
 
 xSemaphoreHandle onTick;
 
 /* This should be 0'd out accroding to C standards */
 static struct sample g_sample_buffer[LOGGER_MESSAGE_BUFFER_SIZE] = {0};
+
+struct sample * get_current_sample(void)
+{
+		return current_sample;
+}
 
 static LoggerMessage getLogStartMessage()
 {
@@ -270,6 +276,8 @@ void loggerTaskEx(void *params)
 
                 ++bufferIndex;
                 bufferIndex %= buffer_size;
+
+                current_sample = sample;
         }
 
         panic(PANIC_CAUSE_UNREACHABLE);
--- a/src/logger/sampleRecord.c
+++ b/src/logger/sampleRecord.c
@@ -27,9 +27,11 @@
 #include "mem_mang.h"
 #include "sampleRecord.h"
 #include "taskUtil.h"
-
+#include "macros.h"
 #include <stdbool.h>
+#include "printk.h"
 
+#define LOG_PFX "[sampleRecord] "
 size_t init_sample_buffer(struct sample *s, const size_t count)
 {
         if (s->channel_samples)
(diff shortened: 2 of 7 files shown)


test/sampleRecord_test.cpp
```diff
@@ -441,3 +441,28 @@ void SampleRecordTest::testLoggerMessageAlwaysHasTime() {
 
         CPPUNIT_ASSERT_EQUAL(true, tick < 1000);
 }
+
+void SampleRecordTest::test_get_sample_value_by_name()
+{
+		lc->ADCConfigs[7].scalingMode = SCALING_MODE_RAW;
```
stieg: What is up with your indentation?  Something is not right with your editor.

src/logger/loggerTaskEx.c
```diff
@@ -52,12 +52,18 @@
 int g_loggingShouldRun;
 int g_configChanged;
 int g_telemetryBackgroundStreaming;
+struct sample * current_sample = NULL;
```
stieg: So why are you going back to this exactly?  IIRC I removed this because of race conditions this would cause at high sample rates.

src/lua/luaLoggerBinding.c
```diff
@@ -896,19 +896,23 @@ static int lua_get_virtual_channel(lua_State *ls)
          * Numbers can be passed in as 123 or "123" in lua.  So handle
          * that case first
          */
-        VirtualChannel *vc;
         if (lua_isnumber(ls, 1)) {
+        		VirtualChannel *vc;
                 vc = get_virtual_channel(lua_tointeger(ls, 1));
+                if (vc) {
```
stieg: Same here with indentation.  Its messed up.  Please fix so we can keep this codebase clean

src/logger/sampleRecord.c
```diff
@@ -54,6 +56,43 @@ void free_sample_buffer(struct sample *s)
         s->channel_samples = NULL;
 }
 
+bool get_sample_value_by_name(struct sample *s, const char * name, double *value)
+{
+    if (!s || !value) return false;
```
stieg: What about name here?  If that is null should it do anything?

test/sampleRecord_test.cpp
brentpicasso: I updated my IDE, and I think it went back to some defaults.  I'll sweep through the entire code with a formatter after this PR

src/logger/loggerTaskEx.c
brentpicasso: I thought this was the best way for Lua to reach in and get the current sample. 

Since Lua is reaching in and not the logger task making callbacks, this was the best way I could think of.   

Suggestions on alternatives? Make a copy of it in memory and setting the current sample to that instead?

src/logger/sampleRecord.c
brentpicasso: I should catch that. I think STR_EQ would choke on a null string.

brentpicasso: fixed.

src/logger/loggerTaskEx.c
brentpicasso: Per meeting, this is an acceptable solution with minimal risk.

PR #947, opened 2017-10-10
P/brent/gopro logic

### LINKED ISSUE #399: Add GoPro Camera Support to firmware
As part of the KS deliverable, need to support GoPro camera control over WiFi.  Depends on #165

brentpicasso:
* Add logic to connect triggering to GoPro wifi triggering
* Changed auto-logger to trigger on any channel
* Camera control can trigger on any channel

* Refactor common code for triggering into auto_control.c 

Resolves #399 , #757


platform/mk2/capabilities.h
```diff
@@ -15,6 +15,7 @@
 #define USB_SERIAL_SUPPORT	1
 #define VIRTUAL_CHANNEL_SUPPORT	1
 #define WIFI_SUPPORT		1
+#define CAMERA_CONTROL      0
```
stieg: MK2 gets no camera control?

src/logger/auto_logger.c
```diff
@@ -50,147 +45,74 @@
...
+        switch(res){
+                case AUTO_CONTROL_TRIGGERED:
+                        startLogging();
+                        pr_info(LOG_PFX "Auto-starting logging\r\n");
+                        break;
+                case AUTO_CONTROL_UNTRIGGERED:
+                        stopLogging();
+                        pr_info(LOG_PFX "Auto-stopping logging\r\n");
+                        break;
+                default:
```
stieg: Unnecessary because if there is no match it will just skip the switch statement.

platform/mk2/capabilities.h
brentpicasso: No. I anticipate in the future we may need custom firmware on the ESP8266, and MK2 has no way to flash custom firmware to wifi, since the boot and reset pins are not wired.

src/logger/auto_logger.c
brentpicasso: Right. I think I did it to make the compiler happy, but maybe it was to make the IDE warnings happy. I'll remove it if it does not fail the build.  Thanks.

PR #955, opened 2017-10-30
query sensors at actual rate if we're logging

### LINKED ISSUE #954: Sensors logging at background sample rate even if we're logging
Looks like we're stuck at 50Hz for sensor channels even if we're logging data and sensor channels are > 50Hz. Need to bump up sample rate to actual logging rate if we're actually logging.

brentpicasso:
…ng; otherwise do the background rate. resolves #954

### DIFF
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -210,7 +210,12 @@ void loggerTaskEx(void *params)
 
                 const bool is_logging = logging_is_active();
 
-                if ((is_logging && currentTicks % loggingSampleRate == 0) || (currentTicks % BACKGROUND_SAMPLE_RATE == 0))
+                /**
+                 * Ensure we refresh the internal sensors at either the
+                 * logging rate or at least at background sample rate
+                 */
+                if ((is_logging && currentTicks % loggingSampleRate == 0) ||
+                                (currentTicks % BACKGROUND_SAMPLE_RATE == 0))
                         doBackgroundSampling();
 
                 if (g_loggingShouldRun && !is_logging) {
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -208,10 +208,11 @@ void loggerTaskEx(void *params)
                 /* Only reset the watchdog when we are configured and ready to rock */
                 watchdog_reset();
 
-                if (currentTicks % BACKGROUND_SAMPLE_RATE == 0)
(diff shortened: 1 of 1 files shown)


RodADavison:
Would this be closer to the intended function?:
`if ((is_logging && currentTicks % loggingSampleRate == 0) || (!is_logging && currentTicks % BACKGROUND_SAMPLE_RATE == 0))`
rather than:
`if ((is_logging && currentTicks % loggingSampleRate == 0) || (currentTicks % BACKGROUND_SAMPLE_RATE == 0))`

src/logger/loggerTaskEx.c
```diff
@@ -208,10 +208,11 @@ void loggerTaskEx(void *params)
                 /* Only reset the watchdog when we are configured and ready to rock */
                 watchdog_reset();
 
-                if (currentTicks % BACKGROUND_SAMPLE_RATE == 0)
+                const bool is_logging = logging_is_active();
+
+                if ((is_logging && currentTicks % loggingSampleRate == 0) || (currentTicks % BACKGROUND_SAMPLE_RATE == 0))
```
stieg: Wrap b/c line width.

stieg:
@RodADavison Your code assumes that the logging rate is always a multiple of the background sample rate.  What if the background sample rate is faster than the logging sample rate?  Or what if they are not a multiple of one another.  Then it breaks down.

RodADavison:
My suggestion was to fix the problems stieg mentions. I noticed the same issues.

src/logger/loggerTaskEx.c
brentpicasso: ok.

brentpicasso:
This logic line controls whether we refresh the ADCs and IMU for their current values - it doesn't actually create a sample record. 

@stieg  I think you're suggesting putting it below this code:

                /* Check if we need to actually populate the buffer. */
                const int sampledRate = populate_sample_buffer(sample,
                                                               currentTicks);
                if (sampledRate == SAMPLE_DISABLED)
                        continue;

If we do that, then we create the sample with old data from the ADC and IMU; having it above prepares the system with fresh sensor values. 

Also, this code change only augments the existing logic to ensure we refresh the ADCs and IMU at the logging rate or the background sample rate, ensuring we at least refresh the ADCs/IMU at the background sample rate.

brentpicasso:
Timeout on PR. merging; low impact change.

Issue #959, opened 2017-11-17
Bit mode little endian does not decode values properly


brentpicasso:
There is a bug with little endian bit mode CAN mapping when the bit length is not an even multiple of 8 bits (not on byte boundaries). 

Big endian bit mapping works properly - when extracting from a CAN stream, a message is extracted as such - (example 12 bit value, x= don't care bits)

xxxxHHHH LLLLLLLL

With little endian bit mode mapping we need to do the endian conversion with byte level granularity, with the left over bits tacked on to the end. 

When data comes in over CAN bus it looks like this:

LLLLLLLL HHHHxxxx

Currently the bit mode endian conversion swaps the bytes so it looks like this:

HHHHxxxx LLLLLLLL

Which is incorrect.

PR #960, opened 2017-11-17
Properly decode little-endian encoded bit mode CAN streams

brentpicasso:
Fully described in: #959

### DIFF
--- a/src/util/byteswap.c
+++ b/src/util/byteswap.c
@@ -59,34 +59,36 @@ uint64_t swap_uint64(uint64_t val)
              (val & 0x00FF000000000000UL) >> 40 | (val & 0xFF00000000000000UL) >> 56;
 }
 
-uint64_t to_little_endian_bitmode(uint64_t val, size_t bit_length)
+uint64_t decode_little_endian_bitmode(uint64_t val, size_t bit_length)
 /*
- * Convert a uint64 to little endian with variable bit length, using byte level granularity for endian-ness
+ * Convert a variable bit length little endian encoded stream, accounting for byte-level
+ * endian granularity, also accounting for remaining bits
+ *
  * example conversion:
- * incoming 12 bit big endian: xxxxHHHH LLLLLLLL
- * converts to: LLLLLLLL HHHHxxxx
+ * incoming 12 bit stream:  LLLLLLLL HHHHxxxx
+ * converts to: xxxxHHHH LLLLLLLL
  *
  * @param val incoming value
  * @param bit_length number of bits to swap
  * @return byte swapped value
  */
 {
-    if (bit_length <= 8) return val;
-    if (bit_length <= 16){
-            size_t partial_bit_length = bit_length - 8;
-            uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
-            return ((val & partial_bit_mask) << 8) + ((val >> partial_bit_length) & 0xFF);
-    }
-    if (bit_length <= 24) {
-            size_t partial_bit_length = bit_length - 16;
-            uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
-            return ((val & partial_bit_mask) << 16) + swap_uint16(val >> partial_bit_length);
-    }
-    if (bit_length <= 32) {
-            size_t partial_bit_length = bit_length - 24;
-            uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
-            return ((val & partial_bit_mask) << 24) + swap_uint24(val >> partial_bit_length);
-    }
-    panic(PANIC_CAUSE_UNREACHABLE);
-    return 0; /* to make compiler happy */
+        if (bit_length <= 8) return val;
+        if (bit_length <= 16){
+                size_t partial_bit_length = bit_length - 8;
+                uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
+                return ((val & partial_bit_mask) << 8) + ((val >> partial_bit_length) & 0xFF);
+        }
+        if (bit_length <= 24) {
+                size_t partial_bit_length = bit_length - 16;
+                uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
+                return ((val & partial_bit_mask) << 16) + swap_uint16(val >> partial_bit_length);
+        }
+        if (bit_length <= 32) {
+                size_t partial_bit_length = bit_length - 24;
+                uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
+                return ((val & partial_bit_mask) << 24) + swap_uint24(val >> partial_bit_length);
+        }
+        panic(PANIC_CAUSE_UNREACHABLE);
+        return 0; /* to make compiler happy */
 }
--- a/src/util/byteswap.c
+++ b/src/util/byteswap.c
@@ -73,20 +73,20 @@ uint64_t decode_little_endian_bitmode(uint64_t val, size_t bit_length)
  * @return byte swapped value
  */
 {
-        if (bit_length <= 8) return val;
+        if (bit_length <= 8) return val & ((1 << bit_length) - 1);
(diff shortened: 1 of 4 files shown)


src/util/byteswap.c
```diff
@@ -59,13 +59,36 @@ uint64_t swap_uint64(uint64_t val)
...
-    if (bit_length <= 8) return val;
-    if (bit_length <= 16) return swap_uint16(val);
-    if (bit_length <= 24) return swap_uint24(val);
-    if (bit_length <= 32) return swap_uint32(val);
-    if (bit_length <= 64) return swap_uint64(val);
-    panic(PANIC_CAUSE_UNREACHABLE);
-    return 0; /* to make compiler happy */
+        if (bit_length <= 8) return val;
+        if (bit_length <= 16){
+                size_t partial_bit_length = bit_length - 8;
```
stieg: All these variables you declare in this scope should be `const`

src/util/byteswap.c
```diff
@@ -59,13 +59,36 @@ uint64_t swap_uint64(uint64_t val)
...
 {
-    if (bit_length <= 8) return val;
-    if (bit_length <= 16) return swap_uint16(val);
-    if (bit_length <= 24) return swap_uint24(val);
-    if (bit_length <= 32) return swap_uint32(val);
-    if (bit_length <= 64) return swap_uint64(val);
-    panic(PANIC_CAUSE_UNREACHABLE);
-    return 0; /* to make compiler happy */
+        if (bit_length <= 8) return val;
+        if (bit_length <= 16){
```
stieg: I think this can be simpler.  Try this:
```c
const size_t mask = (1 << bit_length) -1;
 if (bit_length <= 8) return val & mask;
 if (bit_length <= 16) return swap_uint16(val) & mask;
 if (bit_length <= 24) return swap_uint24(val) & mask;
 if (bit_length <= 32) return swap_uint32(val) & mask;
 panic(PANIC_CAUSE_UNREACHABLE);
return 0; /* to make compiler happy */
```

src/util/byteswap.c
```diff
@@ -59,13 +59,36 @@ uint64_t swap_uint64(uint64_t val)
...
+ */
 {
-    if (bit_length <= 8) return val;
-    if (bit_length <= 16) return swap_uint16(val);
-    if (bit_length <= 24) return swap_uint24(val);
-    if (bit_length <= 32) return swap_uint32(val);
-    if (bit_length <= 64) return swap_uint64(val);
-    panic(PANIC_CAUSE_UNREACHABLE);
-    return 0; /* to make compiler happy */
+        if (bit_length <= 8) return val;
```
stieg: No mask here?

src/util/byteswap.c
stieg: If we were smart we could merge 4 methods into 1 by merging all the `swap_uintXX` methods and parameterizing it.

test/can_obd2/can_mapping_test.cpp
```diff
@@ -91,8 +92,22 @@ void CANMappingTest::extract_test_bit_mode(void)
                                         test_value = 0x5555555555555555 & (((uint64_t)1 << length) - 1);
                                 }
 
+                                uint64_t encoded_value = test_value;
+                                if (endian == 0) {
+                                        /* perform byte oriented endian flip that accounts for variable bit length */
+                                        if (length > 8 && length <=16) {
```
stieg: You need a test for < 8 bytes.

test/can_obd2/can_mapping_test.cpp
```diff
@@ -91,8 +92,22 @@ void CANMappingTest::extract_test_bit_mode(void)
                                         test_value = 0x5555555555555555 & (((uint64_t)1 << length) - 1);
                                 }
 
+                                uint64_t encoded_value = test_value;
+                                if (endian == 0) {
```
stieg: Does this even matter if the bit-pattern since the bit-pattern is the same all the way through (little endian vs big endian doen't matter when the pattern is the same for each byte).

test/can_obd2/can_mapping_test.cpp
stieg: LOL... no you don't.

src/util/byteswap.c
brentpicasso: I don't think this will work per the requirements. See the code comments for the function. 

We need to move the high bits to the right so we get this effect:

incoming 12 bit stream:  LLLLLLLL HHHHxxxx
converts to:   xxxxHHHH LLLLLLLL

If we simply swap as you describe, we'll have this:

incoming 12 bit stream:  LLLLLLLL HHHHxxxx
converts to: HHHHxxxx LLLLLLLL

src/util/byteswap.c
```diff
@@ -59,13 +59,36 @@ uint64_t swap_uint64(uint64_t val)
...
+        if (bit_length <= 16){
+                size_t partial_bit_length = bit_length - 8;
+                uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
+                return ((val & partial_bit_mask) << 8) + ((val >> partial_bit_length) & 0xFF);
+        }
+        if (bit_length <= 24) {
+                size_t partial_bit_length = bit_length - 16;
```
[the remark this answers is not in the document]
brentpicasso: doh. right.

src/util/byteswap.c
```diff
@@ -59,13 +59,36 @@ uint64_t swap_uint64(uint64_t val)
...
+                uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
+                return ((val & partial_bit_mask) << 16) + swap_uint16(val >> partial_bit_length);
+        }
+        if (bit_length <= 32) {
+                size_t partial_bit_length = bit_length - 24;
+                uint8_t partial_bit_mask = (1 << partial_bit_length) - 1;
+                return ((val & partial_bit_mask) << 24) + swap_uint24(val >> partial_bit_length);
```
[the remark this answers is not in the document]
brentpicasso: I think it works as intended per the code comments in the function.  What are you seeing wrong?

test/can_obd2/can_mapping_test.cpp
brentpicasso: yeah, we have to craft the bit pattern so it looks like this for little endian.  (xxx means ignore bits)

LLLLLLLL HHHHxxxx

stieg: I am aware of that.  What i mean is that the pattern 0x5555 looks the same in both little and big endian.  Look at it on a hex editor.  Point is that you shouldn't use a pattern like that because it can mask bugs.  Instead you should use a pattern like 0xAA55.  That pattern flips the bits per byte (meaning each byte is the negative of the other).  This will more likely expose problems.

brentpicasso: Got it.  Ok, I changed the bit pattern to 0xAA55AA55AA55...

src/util/byteswap.c
stieg: That is strange if that is how they are doing it because that is not how endianness works.  Endianness describes _byte_ ordering; _bit_ ordering never changes between machines (least significant bit is always on the right).  So accessing 12 bits of memory in little endian it would look like this: L7 L6 L5 L4 L3 L2 L1 L0   xxxxH3 H2 H1 H0.  If they are doing it that way that probably needs to be put in a different method since not everyone will do it that way @brentpicasso .  We can discuss if you need.

https://en.wikipedia.org/wiki/Endianness
https://www.webopedia.com/TERM/B/big_endian.html

brentpicasso: Right. it's been a bit of a brain bender to figure out, but since these bits are transmitted in a stream, the 12 bits are packed together without the leading blank bits  (x x x x H3 H2 H1 H0)

Can chat more in slack.

brentpicasso:
Changes accepted based on off-line conversation.

PR #965, opened 2017-12-11
add sub ID support to CAN mapping. resolves #964

### LINKED ISSUE #964: Indexed CAN mapping
Corresponding App issue: https://github.com/autosportlabs/RaceCapture_App/issues/1514

Some systems, like the AIM Solo DL's GPS data is provided on CAN ID 40, and uses the first byte as a sub-ID (index).

In order to map data using this scheme, we need to specify the sub ID that appears in the first byte of the CAN message.

Need additional items in the API for each CAN mapping selection:

    Enable sub-ID indexing
    Specify the index to match

This is needed for PodiumConnect SoloDL compatibility mode.

brentpicasso:
For certain CAN integrations, such as SmartyCAM GPS data, we need to match on the CAN ID as well as a sub ID in the first data byte. This extends the API and CAN mapping capabilities to meet this requirement. 

Also, refactored out a redundant ChannelConfig block which was wasting 30 bytes per CAN mapping.


test/loggerConfig_test.cpp
```diff
@@ -147,7 +147,7 @@ void LoggerConfigTest::testLoggerInitObd2Config() {
    CPPUNIT_ASSERT(c->enabled == 0);
 
    for (int i = 0; i < OBD2_CHANNELS; ++i) {
-      CPPUNIT_ASSERT(strlen(c->pids[i].mapping.channel_cfg.label));
+      CPPUNIT_ASSERT(strlen(c->pids[i].mapping.channel_cfg.label)== 0);
```
stieg: What changed here exactly?

include/logger/loggerConfig.h
```diff
@@ -344,11 +344,13 @@ typedef struct _CANMapping {
 
     /* the conversion filter to apply to quickly convert units without changing mapping */
     uint8_t conversion_filter_id;
+
+    /* sub ID index. If defined (>=0) then this uses the first byte to mach on a sub address */
+    int8_t sub_id;
```
stieg: So is this some standard or is this some device's special way of doing things.  If the latter the better place would be in the driver that handles that device.

stieg: Also any reason for this to be a signed value?  Probably should be unsigned no?

brentpicasso: It's a quasi standard, so it should be part of the core capability. 

I'm using a singed value so I can use < 0 to indicate the feature is disabled, to prevent storing another flag.  I don't expect more than 127 sub channel IDs, so it should work well for an 8 bit signed.

test/loggerConfig_test.cpp
brentpicasso: There was some old OBDII initialization code that pre-populated the OBDII channel names with values; this was removed separately so the channel name defaulted to empty string. This just acknowledges that change.

PR #970, opened 2017-12-22
keep distance channel even if gps hardware is disabled resolves #969

### LINKED ISSUE #969: Disabling hardware GPS also disables Distance channel
Distance is a calculated channel that shouldn't be disabled if GPS hardware is disabled. 

Should only disable GPS channels that actually come from the GPS device; need to keep the Distance channel since it's calculated from known values.

brentpicasso:


brentpicasso:
Straightforward changes, self-reviewing

PR #971, opened 2017-12-22
enable button if it's the only GPIO in the system. resolves #968

### LINKED ISSUE #968: disabling GPIO also disables button action
Need to handle the case if the device has no GPIO but has the button. If we disable GPIO on capabilities.h, then the button is also disabled.

brentpicasso:


brentpicasso:
self reviewing.

PR #974, opened 2017-12-23
handle cases when cellular module falls into a bad state. 

### LINKED ISSUE #973: Handle cell module bad state
It appears the SARA U2 can crash or fall into a bad state, where responses to messages show up as garbage characters, instead of OK or ERROR, or expected data.  It's possible that the cell module is rebooting into a default baud rate. 

Often this happens while we're waiting to get on the network, which can cause a long wait until the network registration times out.  We should detect these bad responses and reset the module accordingly.

brentpicasso:
Resolves #973. 

It appears the SARA U2 can crash or fall into a bad state, where responses to messages show up as garbage characters, instead of OK or ERROR, or expected data. It's possible that the cell module is rebooting into a default baud rate.

Often this happens while we're waiting to get on the network, which can cause a long wait until the network registration times out. We should detect these bad responses and reset the module accordingly.

 This seems to happen especially after AT+CNUM fails / stops responding - so also made this more robust; also it seems that the phone number is not always available on the SIM card.  So, set the 'Not available' message if no number is available on the SIM.

### DIFF
--- a/src/devices/sara_u280.c
+++ b/src/devices/sara_u280.c
@@ -40,10 +40,12 @@
 #define GPRS_ATTACH_ATTEMPTS	5
 #define GPRS_ATTACH_BACKOFF_MS	1000
 #define NET_REG_ATTEMPTS	50
+#define SARA_U280_ERROR_THRESHOLD 6
 #define NET_REG_BACKOFF_MS	1000
 #define STOP_DM_RX_EVENTS	10
 #define STOP_DM_RX_TIMEOUT_MS	1000
 #define SARA_U280_SUBSCRIBER_NUMBER_RETRIES 3
+#define FAILED_READ_BACKOFF_MS 100
 
 static bool sara_u280_get_subscriber_number(struct serial_buffer *sb,
                                             struct cellular_info *ci)
@@ -54,9 +56,11 @@ static bool sara_u280_get_subscriber_number(struct serial_buffer *sb,
         bool status = false;
         for (size_t i = 0; i < SARA_U280_SUBSCRIBER_NUMBER_RETRIES && !status; ++i)
                 status = gsm_get_subscriber_number(sb, ci);
+                if (!status)
+                        delayMs(FAILED_READ_BACKOFF_MS);
 
         if (!status)
-                pr_warning("[cell] Failed to read phone number\r\n");
+                pr_warning("[cell] Phone number not available\r\n");
         return status;
 }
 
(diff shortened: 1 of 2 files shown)


src/devices/sara_u280.c
```diff
@@ -40,10 +40,12 @@
 #define GPRS_ATTACH_ATTEMPTS	5
 #define GPRS_ATTACH_BACKOFF_MS	1000
 #define NET_REG_ATTEMPTS	50
+#define SARA_U280_ERROR_THRESHOLD 6
 #define NET_REG_BACKOFF_MS	1000
 #define STOP_DM_RX_EVENTS	10
 #define STOP_DM_RX_TIMEOUT_MS	1000
 #define SARA_U280_SUBSCRIBER_NUMBER_RETRIES 3
+#define FAILED_READ_BACKOFF_MS 100
```
stieg: That is quite some time.  Why this number?

src/devices/sara_u280.c
```diff
@@ -54,9 +56,11 @@ static bool sara_u280_get_subscriber_number(struct serial_buffer *sb,
         bool status = false;
         for (size_t i = 0; i < SARA_U280_SUBSCRIBER_NUMBER_RETRIES && !status; ++i)
                 status = gsm_get_subscriber_number(sb, ci);
+                if (!status)
+                        delayMs(FAILED_READ_BACKOFF_MS);
 
         if (!status)
-                pr_warning("[cell] Failed to read phone number\r\n");
+                pr_warning("[cell] Phone number not available\r\n");
```
stieg: So what do the docs say about this? Is there no way to differentiate between it not being available vs a failure?

src/gsm/gsm.c
```diff
@@ -79,24 +79,24 @@ bool gsm_get_subscriber_number(struct serial_buffer *sb,
...
+                goto read_fail;
 
         *num_end = '\0';
         strntcpy(ci->number, num_start, sizeof(ci->number));
         return true;
 
-parse_fail:
-        pr_warning("[gsm] Failed to parse phone number\r\n");
+read_fail:
+        strcpy(ci->number, "Not available");
```
stieg: Is the number long enough to support this string?  Might be smart to use `strcpyn(ci->number, sizeof(ci->number), "Not available")`

src/devices/sara_u280.c
brentpicasso: Eh, you're right - it seems arbitrary. There is no evidence we even need this delay so I will remove it.

src/devices/sara_u280.c
brentpicasso: Internet research says that the module may not return a phone number when asked, and this is a valid case. i.e. it can return just "OK" with no data.  This change acknowledges this possibility.

src/gsm/gsm.c
brentpicasso: It is, but I do like this approach. thanks.

src/devices/sara_u280.c
brentpicasso: Ok, removed delay.

src/gsm/gsm.c
brentpicasso: Ok, changed to strncpy (strlcpy was not available in our toolchain. weird, right?)

PR #981, opened 2018-02-09
Don't generate a lua error if a channel by name/ID is not available

### LINKED ISSUE #978: Possible race condition with getChannel()
It seems that attempting to access channels will periodically fail.  This script reproduces it. 

Could be a race condition where the channel does not exist yet when it is fetched. 

Also, this should not be a fatal runtime error, it should return nil if the channel does not exist. 

example:

setTickRate(10)
mpgID = addChannel("MPG", 10, 2, 0, 100, "mpg")
function doMPG()
  local miles = getChannel("WheelSpeed")
  local gallons = getChannel("FuelConsump")
  local mpg = miles / gallons
  setChannel(mpgID, mpg)
end

function onTick() 
	doMPG()
  end

brentpicasso:
Also, make units label available to caller (internal code) Resolves #978, and normalize units to match our canonical standard.


src/lua/luaLoggerBinding.c
```diff
@@ -904,22 +904,23 @@ static int lua_get_virtual_channel(lua_State *ls)
...
-        		}
-		}
-
-		return luaL_error(ls, "Channel not found!");
+                struct sample * s = get_current_sample();
+				            double value;
+				            char * units;
+				            if (s && get_sample_value_by_name(s, lua_tostring(ls, 1), &value, &units)){
+						                  lua_pushnumber(ls, value);
+						                  return 1;
```
stieg: What  is up with the indentation here?

brentpicasso: IDK, it's correct in my editor with spaces :-/

brentpicasso:
Thanks for the review!

PR #1000, opened 2018-07-09
extend dynamic range of filter. set min alpha to .0001

### LINKED ISSUE #999: Extend dynamic range of smoothing
Expand dynamic range of smoothing to account for scenarios that demand it, such as accounting for fuel tank slosh.

brentpicasso:
resolves #999. Improves dynamic range of filtering so we can account for scenarios like fuel tank slosh. 

When filtering is set to maximum (alpha = 0.001) time to sweep from min to max scale is about 200 seconds.

### DIFF
--- a/src/filter/filter.c
+++ b/src/filter/filter.c
@@ -21,24 +21,24 @@
 
 
 #include "filter.h"
+#include "stdutil.h"
 
 //Implements a fast Exponential Moving Average filter
 #define MIN_ALPHA 0.0001
-void init_filter(Filter *filter, float alpha)
+void init_filter(Filter *filter, const float alpha)
 {
-        filter->max_samples = 1 / (alpha != 0 ? alpha : MIN_ALPHA);
+        filter->max_samples = (int32_t)(1 / MAX(alpha, MIN_ALPHA));
         filter->current_value = 0;
         filter->total = 0;
         filter->count = 0;
 }
 
-int32_t update_filter(Filter *filter, int32_t value)
+int32_t update_filter(Filter *filter, const int32_t value)
 {
         filter->total += value;
         if (filter->count >= filter->max_samples) {
                 filter->total -= filter->current_value;
-        }
-        else {
+        } else {
                 filter->count++;
         }
         filter->current_value = filter->total / filter->count;
--- a/src/filter/filter.c
+++ b/src/filter/filter.c
@@ -22,28 +22,25 @@
 
 #include "filter.h"
 
-//Implements a fast Exponential Moving Average IIR filter
-
-//This macros defines an alpha value between 0 and 1
-#define DSP_EMA_I32_ALPHA(x) ( (unsigned short)(x * 65535) )
-
-static int dsp_ema_i32(int in, int average, unsigned short alpha)
-{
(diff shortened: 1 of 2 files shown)


src/filter/filter.c
```diff
@@ -22,28 +22,25 @@
...
-    return (int)((tmp0 + 32768) / 65536); //scale back to 32-bit (with rounding)
-}
-
+//Implements a fast Exponential Moving Average filter
+#define MIN_ALPHA 0.0001
 void init_filter(Filter *filter, float alpha)
 {
-    filter->alpha = alpha;
-    filter->current_value = 0;
+        filter->max_samples = 1 / (alpha != 0 ? alpha : MIN_ALPHA);
```
stieg: This needs to change as it allows for an alpha value below the MIN_ALPHA value
```c
filter->max_samples = 1 / max(alpha, MIN_ALPHA);
```

stieg: Also... this is a float being assigned to an int type ja?  I assume there is an implicit cast in there but not sure when it happens.  Double check it works as you believe it should.

src/filter/filter.c
```diff
@@ -22,28 +22,25 @@
...
+        filter->current_value = 0;
+        filter->total = 0;
+        filter->count = 0;
 }
 
-int update_filter(Filter *filter, int value)
+int32_t update_filter(Filter *filter, int32_t value)
```
stieg: Should there be be `const` modifiers on the parameters?

src/filter/filter.c
```diff
@@ -22,28 +22,25 @@
...
 {
-    int current_value = filter->current_value;
-    current_value = dsp_ema_i32(value, current_value, DSP_EMA_I32_ALPHA(filter->alpha));
-    filter->current_value = current_value;
-    return current_value;
+        filter->total += value;
+        if (filter->count >= filter->max_samples) {
+                filter->total -= filter->current_value;
+        }
```
stieg:
```c
if (something) {
...
} else {
...
}
```

src/filter/filter.c
brentpicasso: Ok, addressing both points.

src/filter/filter.c
brentpicasso: yep, added to the 2nd param, but not the first, as we're making a change to the member of the struct.

src/filter/filter.c
brentpicasso: fixed

brentpicasso:
Thanks for the review. Changes per feedback have been applied.

src/filter/filter.c
```diff
@@ -21,24 +21,24 @@
...
 #include "filter.h"
+#include "stdutil.h"
 
 //Implements a fast Exponential Moving Average filter
 #define MIN_ALPHA 0.0001
-void init_filter(Filter *filter, float alpha)
+void init_filter(Filter *filter, const float alpha)
 {
-        filter->max_samples = 1 / (alpha != 0 ? alpha : MIN_ALPHA);
+        filter->max_samples = (int32_t)(1 / MAX(alpha, MIN_ALPHA));
```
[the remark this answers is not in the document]
brentpicasso: Thanks, applied your suggestions.

PR #1006, opened 2019-02-08
P/brent/shiftx3 support

### LINKED ISSUE #1007: Add support for ShiftX (2/3) lua functions and button press notifications
* Create helper functions for ShiftX2/3 to reduce memory consumption and complexity of user scripts
* Broadcast button events to connected clients

brentpicasso:
Resolves #1007

### DIFF
--- a/include/CAN/CAN_dispatcher.h
+++ b/include/CAN/CAN_dispatcher.h
@@ -26,7 +26,7 @@
 
 CPP_GUARD_BEGIN
 
-void can_dispatch_message(CAN_msg *msg);
+void can_dispatch_message(const CAN_msg *msg);
 
 CPP_GUARD_END
 
--- a/include/CAN/CAN_dispatcher.h
+++ b/include/CAN/CAN_dispatcher.h
@@ -1,0 +1,33 @@
+/*
+ * Race Capture Firmware
+ *
+ * Copyright (C) 2016 Autosport Labs
+ *
+ * This file is part of the Race Capture firmware suite
+ *
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _CAN_DISPATCHER_H_
+#define _CAN_DISPATCHER_H_
+
+#include "CAN.h"
+
+CPP_GUARD_BEGIN
+
+void can_dispatch_message(CAN_msg *msg);
+
+CPP_GUARD_END
+
+#endif /* _CAN_DISPATCHER_H_ */
--- a/include/drivers/shiftx_drv.h
+++ b/include/drivers/shiftx_drv.h
@@ -1,0 +1,53 @@
+/*
+ * Race Capture Firmware
+ *
+ * Copyright (C) 2016 Autosport Labs
+ *
+ * This file is part of the Race Capture firmware suite
+ *
+ * This is free software: you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This software is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ *
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef _SHIFTX_DRV_H_
+#define _SHIFTX_DRV_H_
+
+#include "CAN.h"
+
+CPP_GUARD_BEGIN
+
+struct shiftx_configuration {
+        uint8_t orientation_inverted;
+        uint8_t brightness;
+        uint32_t can_bus;
+        uint32_t base_address;
+        uint8_t auto_brightness_scaling;
+};
+
+struct shiftx_configuration * shiftx_get_config(void);
+bool shiftx_update_config(void);
+
+bool shiftx_set_discrete_led(uint8_t led_index, uint8_t leds_to_set, uint8_t red, uint8_t green, uint8_t blue, uint8_t flash);
+bool shiftx_set_display(uint8_t digit_index, uint8_t ascii);
+
+bool shiftx_config_linear_graph(uint8_t rendering_style, uint8_t linear_style, uint16_t low_range, uint16_t high_range);
+bool shiftx_set_linear_threshold(uint8_t threshold_id, uint8_t segment_length, uint16_t threshold, uint8_t red, uint8_t green, uint8_t blue, uint8_t flash);
+bool shiftx_update_linear_graph(uint16_t value);
+
+bool shiftx_set_alert_threshold(uint8_t alert_id, uint8_t threshold_id, uint16_t threshold, uint8_t red, uint8_t green, uint8_t blue, uint8_t flash);
+bool shiftx_set_alert(uint8_t alert_id, uint8_t red, uint8_t green, uint8_t blue, uint8_t flash);
+bool shiftx_update_alert(uint8_t alert_id, uint16_t value);
+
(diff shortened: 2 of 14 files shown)


src/lua/luaLoggerBinding.c
```diff
@@ -977,6 +979,257 @@ static int lua_update_gps(lua_State *L)
...
+                lua_pushinteger(L, button_id);
+                lua_pushinteger(L, state);
+                return 2;
+        }
+        else {
+                return 0;
+        }
+}
+
+static int lua_tx_button(lua_State *L)
```
ddimensia: Should the second argument be a boolean and not an integer?  Either that or, assuming you're thinking of more than a bimodal state, should this be named tx_switch to be more generic?  Also, is the button id some opaque id that gets mapped on the app side or is it intended to map the the ascii keyboard event?  Either way it would be nice if a table could be pushed into the Lua state that maps logical button names to their ID value.

brentpicasso: Agreed, we can make it a boolean. Good catch!

brentpicasso: w/r/t using text labels for buttons, I don't think it's worth the complexity to map names at this point, given the embedded environment. We'll have a map of button codes that will correlate with CAN messages. I think it would be acceptable to expect users to remember 0=left button, 1=right button.

ddimensia: I think that's fine, just needs to be addressed with documentation with a list of ids and their behavior.

include/CAN/CAN_dispatcher.h
```diff
@@ -0,0 +1,33 @@
+/*
+ * Race Capture Firmware
+ *
+ * Copyright (C) 2016 Autosport Labs
```
stieg: LoL date.

include/drivers/shiftx_drv.h
```diff
@@ -0,0 +1,147 @@
...
+
+#ifndef _SHIFTX_DRV_H_
+#define _SHIFTX_DRV_H_
+
+#include "CAN.h"
+
+CPP_GUARD_BEGIN
+
+struct shiftx_configuration {
+        uint8_t orientation_inverted;
```
stieg: Alignment of these fields is sub optimal.  move the Uint32_t fields up and group the 3 uint8_t types together for better compaction

include/drivers/shiftx_drv.h
```diff
@@ -0,0 +1,147 @@
...
+
+/**
+ * Retreive a pointer to the current runtime configuration
+ * @return pointer to struct of the shiftx_configuration
+ */
+struct shiftx_configuration * shiftx_get_config(void);
+
+
+/**
+ * Send the configuration update message to the connected ShiftX device, using the current runtime configurationi
```
stieg: Wrapping helps.  IIRC our standard is 80 or 100.  Just do something so lines aren't insane

include/drivers/shiftx_drv.h
```diff
@@ -0,0 +1,147 @@
...
+ * Set the display value
+ * @param digit_index index of digit to set
+ * @param ascii ascii value to set (48 = '0', 49 = '1')
+ * @return true if CAN message was successfully broadcasted
+ */
+bool shiftx_set_display(uint8_t digit_index, uint8_t ascii);
+
+/**
+ * Configures the linear graph on the ShiftX device
+ * @param rendering_style 0=left->right; 1=center; 2=right->left
```
stieg: This should be an enum type instead of a uint8_t

include/drivers/shiftx_drv.h
```diff
@@ -0,0 +1,147 @@
...
+ * @param linear_style 0=smooth/interpolated; 1=stepped
+ * @param low_range low value range of linear display
+ * @param high_range high value range of linear display (ignored if rendering_style = stepped)
+ * @return true if CAN message was successfully broadcasted
+ */
+bool shiftx_config_linear_graph(uint8_t rendering_style, uint8_t linear_style, uint16_t low_range, uint16_t high_range);
+
+/**
+ * Configures a threshold for the linear graph on the ShiftX device
+ * @param threshold_id id of threshold (0-4)
```
stieg: This doesn't tell me much.  Improve please.

include/drivers/shiftx_drv.h
```diff
@@ -0,0 +1,147 @@
...
+/**
+ * Directly set an alert indicator
+ * @param alert_id id of alert (0-> number of alert indicators on device)
+ * @param red red color value
+ * @param green green color value
+ * @param blue blue color value
+ * @param flash flash rate, in Hz
+ * @return true if CAN message was successfully broadcasted
+ */
+bool shiftx_set_alert(uint8_t alert_id, uint8_t red, uint8_t green, uint8_t blue, uint8_t flash);
```
stieg: I thought, consider passing a struct with the RGB and flash values instead of declaring parameters for each one.  Will make it easier to change many methods if you need to add anything in the future and plus it's a bit cleaner.

src/CAN/CAN_dispatcher.c
```diff
@@ -0,0 +1,30 @@
...
+ * See the GNU General Public License for more details. You should
+ * have received a copy of the GNU General Public License along with
+ * this code. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "CAN_dispatcher.h"
+#include "CAN.h"
+#include "shiftx_drv.h"
+
+void can_dispatch_message(CAN_msg *msg)
```
stieg: Don't forget your `const` prefixes unless the values are going to change

src/drivers/shiftx_drv.c
```diff
@@ -0,0 +1,223 @@
...
+
+#define NOTIFICATION_BUTTON_STATE_OFFSET 60
+
+static struct shiftx_configuration shiftx_config = {0, 0, 1, 0xE3600, 51};
+
+static struct {
+        bool received;
+        uint8_t id;
+        uint8_t state;
+} button_state = {0};
```
stieg: The initialization is unnecessary as it is a static struct and thus will be initialized to zero as part of BSS.  Unless you want this in data section for some reason.

src/lua/luaLoggerBinding.c
```diff
@@ -977,6 +979,257 @@ static int lua_update_gps(lua_State *L)
         return 0;
 }
 
+static int lua_sx_update_linear_graph(lua_State *L)
+{
+        lua_validate_args_count(L, 0, 1);
+        if (lua_gettop(L) > 0) {
+                lua_validate_arg_number(L, 1);
+                uint16_t value = lua_tointeger(L, 1);
```
stieg: Any check for overflow here?

stieg: Probably not a issue due to implicit casting.

src/lua/luaLoggerBinding.c
cherongroup-gh: I like the idea of more than boolean for button state.  I use, quite often, multiple states such as { button_not_pressed, button_pressed, button_released, button_long_pressed, button_long_released }.  This allows the lower level code to differentiate short and long presses, as well as looking for change-of-state.  Say, for example, your code wants to look when a button is released, not just that it is not pressed.  Not-pressed is something more for start up whereas button released would indicate it was pressed then released.

ddimensia: I was thinking more about rotary encoders when making the initial comment, like the idea of having an encoder to select a specific screen.  I'd have to look at what something like the Ecumaster CAN switch panel supports to know if the long/short press timing is calculated on the button panel.  I don't think ShiftX devices support that and since they're not field upgradeable it would have to be emulated in Lua.

ddimensia: Doing a cursory look at the Ecumaster docs indicates that they only support binary state of the button, this would mean that if long/short press was to be implemented it would probably need to be on the RCP/Lua side.  The main use I could see for that is allowing a button to dual purpose as something like left/ok, right/cancel.

brentpicasso: We can also keep it 0/1 for now and expand it later to support more gestures (long press, double press, or whatever we come up with)

include/drivers/shiftx_drv.h
brentpicasso: done. Good catch.

include/drivers/shiftx_drv.h
brentpicasso: ok

include/drivers/shiftx_drv.h
brentpicasso: done. Thanks.

src/CAN/CAN_dispatcher.c
brentpicasso: fixed. thanks.

src/lua/luaLoggerBinding.c
brentpicasso: Yeah, it won't be crashy if they send bad data, ShiftX3 just won't work as expected.

include/drivers/shiftx_drv.h
brentpicasso: Thanks for the suggestion. Will implement.

PR #1010, opened 2019-02-22
make broadcasting of button events configurable

brentpicasso:
Allow enabling/disabling of button events for ShiftX3 via ShiftX3 config

### DIFF
--- a/src/drivers/shiftx_drv.c
+++ b/src/drivers/shiftx_drv.c
@@ -45,7 +45,7 @@
 
 #define NOTIFICATION_BUTTON_STATE_OFFSET 60
 
-static struct shiftx_configuration shiftx_config = {1, 0xE3600, 0, 0, 51};
+static struct shiftx_configuration shiftx_config = {1, 0xE3600, 0, 0, 51, true};
 
 static struct {
         bool received;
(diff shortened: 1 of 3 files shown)


src/drivers/shiftx_drv.c
```diff
@@ -67,8 +67,9 @@ void shiftx_handle_can_rx_msg(const CAN_msg *msg)
                 shiftx_update_config();
         }
 
-        if (msg->addressValue == shiftx_config.base_address + NOTIFICATION_BUTTON_STATE_OFFSET) {
-                pr_info_int_msg(_LOG_PFX "Received button event for base address: ", msg->addressValue);
+        if (msg->addressValue == shiftx_config.base_address + NOTIFICATION_BUTTON_STATE_OFFSET
+                        && shiftx_config.button_events_enabled) {
```
ddimensia: Shouldn't this just omit generating the API event down below (lines 85-93)?  It would still be good to use the 'shiftx_rx_button_press' method to be able to query if a button was pressed on the ShiftX.

brentpicasso: Agreed. will change it.

brentpicasso: Ok, fixed.

PR #1014, opened 2019-02-25
enable static OBDII PID delay to slow PID querying 

### LINKED ISSUE #1013: Add configurable static delay for OBDII queries
We suspect some ECUs that may not handle CAN OBDII properly, (especially ECUs that implemented CAN OBDII before the 2008 mandate and may have bugs). 

An issue is that ECUs may have problems handling high speed, back-to-back OBDII PID queries. 

Need a way to slow down the PID requests to aid in troubleshooting these platforms. 

Example: Mazda RX8, 2004 with manual transmission. In this case, when OBDII is enabled, upper rev limit seems affected.

brentpicasso:
...for troubleshooting / diagnostic purposes when dealing with ECUs that cannot handle RaceCapture's OBDII query rate. 

 Resolves #1013

### DIFF
--- a/include/CAN/OBD2.h
+++ b/include/CAN/OBD2.h
@@ -84,6 +84,13 @@ void update_obd2_channels(CAN_msg *msg, OBD2Config *cfg);
  */
 bool OBD2_get_value_for_pid(uint16_t pid, float *value);
 
+/**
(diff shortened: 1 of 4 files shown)


brentpicasso:
self reviewing as it is a small change.

PR #1016, opened 2019-02-26
Address track database update error under certain conditions

### LINKED ISSUE #1015: Cannot update tracks under certain conditions
If tracks cannot be updated under certain conditions where memory consumption might be high (Lua script loaded) the track updating routine may fail due to not allocating buffer memory.  This will cause the update process to fail.

brentpicasso:
Allow tracks to be updated under high memory usage conditions by reduccing track database size to 50. Resolves #1015


brentpicasso:
Self reviewing as it is a small change.

Issue #1028, opened 2019-05-07
API for resetting lap stats


brentpicasso:
Add API for resetting lap stats:

* JSON streaming API
* Lua scripting

Issue #945, opened 2017-10-04
Reset lap timing from Lua and API


brentpicasso:
Need a way to reset lap timing, including predictive timing from Lua and via API, so app can reset as well.

PR #1030, opened 2019-05-09
P/brent/session time

### LINKED ISSUE #1029: SessionTime channel
A channel to track the overall time for the current session. Would reset when lapstats would reset.

SessionTime

brentpicasso:
Creates a "SessionTime" channel which tracks the length of a session.  Continues until it is reset by the user via RaceCapture (API) or Lua. 

Resolves #1029  #945


brentpicasso:
The call to lapstats has a boolean to indicate if it should reset session time. Only if it is true will it reset session time.  

Calls from Lua or API will reset session time, everywhere else it will not.

brentpicasso:
resolves #1028

Issue #1032, opened 2019-05-10
Don't power cycle cellular on every connection retry


brentpicasso:
If the telemetry connection drops, try to re-connect without going through a full power cycle. This will help re-acquire the connection faster. 

We should try a few times to re-connect without power cycling; after a certain threshold - say 5 - we should fall back to power cycling.

PR #1034, opened 2019-05-11
Implement support for SARA-R4 module.

### LINKED ISSUE #1033: SARA-R4 support
Implement SARA-R4 support.

brentpicasso:
Implement support for SARA-R4 module, and optimize re-connect logic. 

Resolves #1033  #1032.


src/devices/cellular.c
```diff
@@ -419,9 +441,11 @@ int cellular_init_connection(DeviceConfig *config, millis_t * connected_at)
         /* This is sane since DeviceConfig is typedef'd as a serial_buffer */
         struct serial_buffer *sb = (struct serial_buffer*) config;
 
-        const int init_status = cellular_init_modem(sb);
-        if (DEVICE_INIT_SUCCESS != init_status)
-                return DEVICE_INIT_FAIL;
+        if (hard_init) {
+                const int init_status = cellular_init_modem(sb);
```
stieg: So we only initialize now if hard init is set?  We otherwise don't init?  Seems off.

src/gsm/gsm.c
```diff
@@ -38,7 +38,7 @@ bool gsm_ping_modem(struct serial_buffer *sb)
 
         serial_buffer_reset(sb);
         serial_buffer_append(sb, "AT");
-        const size_t count = cellular_exec_cmd(sb, READ_TIMEOUT, msgs,
+        const size_t count = cellular_exec_cmd(sb, READ_TIMEOUT * 2, msgs,
```
stieg: Why not just increase READ_TIMEOUT and not have unexpected behavior?

src/devices/cellular.c
brentpicasso: We pass in a flag for hard initialization - this is controlled by the outside routine (connectivity task). 

If the hard_init is false, we bypass the power cycle routine, which will speed up cellular acquisition. 

The gist is, if we lose connection, we give the system a chance to quickly re-connect. 

The logic in the connectivity task will do a 'fast init' (i.e. hard_init = false) up to X amount of times before it falls back to power cycling.

src/gsm/gsm.c
brentpicasso: I think that's  good idea. fixed.

PR #1040, opened 2019-07-03
Fixed some warnings caused by pedantic compiler options

RodADavison:
fixed redundant const declarations and 
empy while loops with semi-colons.


brentpicasso:
Hi, thanks for the PR.  Please apply this against branch r/2.17.2

The way we manage our branches is that master is latest released stable, and the upcoming release is in a named branch r/major.minor.bugfix (currently 2.17.2). 

Thanks!

PR #1047, opened 2019-08-28
P/rod/w double promote

RodADavison:
I have added the -Wdouble-promote warning to the makefiles.  I also rolled the "dependency checking" fix into this PR.  Please take a careful look a the change in predictive_timer_2.c at line 176.  The comment there makes me a little concerned that it might have depend on doing double math there.


brentpicasso:
Hi, could you point this against r/2.17.4 ? should be a quick change here in the PR.  Thanks!

brentpicasso:
NVM, I was able to do it. :)

src/util/modp_numtoa.c
```diff
@@ -141,6 +141,13 @@ void modp_ftoa(float value, char* str, int prec)
                 prec = 9;
         }
 
+	/* check for nan */
+	if ( value != value )
```
brentpicasso: So, this is going to blow up things that expect an actual number.  I believe we use this function to populate values in JSON.  With this change, we'll have something like this, which is invalid JSON:

{
	"x": nan
}

Does the original behavior output zero? Any ideas what it could output that's closer to a real value?

src/util/modp_numtoa.c
```diff
@@ -233,6 +240,13 @@ void modp_dtoa(double value, char* str, int prec)
...
+	{
+                strcpy(str,"nan");
```
brentpicasso: Ditto from previous comment

include/imu/imu.h
```diff
@@ -37,7 +39,7 @@ int imu_soft_init(LoggerConfig *loggerConfig);
 
 void imu_calibrate_zero();
 
-int imu_read(enum imu_channel channel);
+int imu_read_raw(enum imu_channel channel);
```
brentpicasso: It looks like you have some unrelated code mixed in with these changes - here and the GPS data.  Please submit these as part of separate feature PR's. Thanks!

src/imu/imu.c
```diff
@@ -51,10 +51,17 @@ static void init_filters(LoggerConfig *loggerConfig)
 void imu_sample_all()
 {
         for (size_t i = 0; i < CONFIG_IMU_CHANNELS; i++) {
-                update_filter(&g_imu_filter[i], imu_read(i));
+                update_filter(&g_imu_filter[i], imu_read_raw(i));
         }
 }
 
+float imu_read_cal_value(enum imu_channel channel )
```
brentpicasso: Looks like an unrelated change, please submit as part of a different PR when that feature is ready

src/util/modp_numtoa.c
RodADavison: Yes.  The original behaviour was to print "0.0".  I see the problem here.  Python throws an exception, whereas C happily runs code that divides by zero.  Technically this is not needed for the -Wdouble_promote.  It would be nice to see bad values when debugging, so probably safer to omit this although it would be handy to catch cases where we divide by 0 in firmware.  #ifdef DEBUG?

src/imu/imu.c
RodADavison: I knew you'd notice.  I struggle so much with git to pull things apart after the fact, I was hoping I could slide this in since it touches dependant files and is pretty innocuous, rather than tease it out.   If you really want me to relocate it I will.

brentpicasso: Please do. It seems pedantic but it really helps down the line to keep the concepts separate. Thanks!

src/util/modp_numtoa.c
brentpicasso: Retaining the original behavior and maybe doing an assert if in debug mode could be a good approach.  Check to see if we're already doing asserts elsewhere in the firmware, I think we are...

RodADavison: I don't see any assert usage.   Alternatively, it looks like the pr_..._float macros are the only ones that use printk_float_msg.  I could catch my bad floats there instead.

brentpicasso:
Getting a compile error after a make clean; make

../../src/lua/luaLoggerBinding.c: In function 'lua_set_analog_out':
../../src/lua/luaLoggerBinding.c:206:60: error: implicit conversion from 'float' to 'double' to match other operand of binary expression [-Werror=double-promotion]
         const float dutyCycle = (float) lua_tonumber(L, 2) /
                                                            ^

PR #1051, opened 2019-09-06
Update metadata while telemetry buffering

### LINKED ISSUE #1050: Metadata not updated in stream if channel structure changes
If channels are added / changed after telemetry buffering starts, metadata must be updated in the stream. 

Repro steps:

* enable telemetry buffering (SD card present)
* Create a virtual channel in lua
* When connected via telemetry, observe failure from Podium about invalid data length. (podium receives unexpected amounts of data)

brentpicasso:
Trigger an update to metadata during buffering if virtual channels are added; address race condition only present with the new telemetry buffering. 

Resolves #1050

### DIFF
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -51,7 +51,7 @@
 #define BACKGROUND_SAMPLE_RATE	SAMPLE_50Hz
 
 int g_loggingShouldRun;
-int g_config_changed;
+bool g_config_changed;
 int g_telemetryBackgroundStreaming;
 struct sample * current_sample = NULL;
 
--- a/src/logger/loggerTaskEx.c
+++ b/src/logger/loggerTaskEx.c
@@ -51,7 +51,7 @@
 #define BACKGROUND_SAMPLE_RATE	SAMPLE_50Hz
 
 int g_loggingShouldRun;
-int g_configChanged;
+int g_config_changed;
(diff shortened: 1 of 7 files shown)


src/logger/loggerTaskEx.c
```diff
@@ -51,7 +51,7 @@
 #define BACKGROUND_SAMPLE_RATE	SAMPLE_50Hz
 
 int g_loggingShouldRun;
-int g_configChanged;
+int g_config_changed;
 int g_telemetryBackgroundStreaming;
```
RodADavison: Did you intend to change int to bool as you changed assignments from 0/1 to false/true.  I can't actually tell what size gcc is using for a bool, but it is a bit more consistent.  (Looking stdbool.h it defines false/true to 0/1) so its functionally pretty much the same.

brentpicasso: General refactoring of old code.  The variable was always intended as bool. :+1:

RodADavison: I really meant - it looks like we are using "bool" elsewhere rather than "int".  I'm hoping that "bool" is of type "unsigned char" rather than "int".  I'd encourage changing the type for this variable too.

brentpicasso: Ah, got it.  Yes, I did want to change the type too, good catch. Fixed now.

PR #1061, opened 2019-10-15
Added compiler options that improve removal of unused code.

RodADavison:
reduces code size (of mk3) by about 50k, if I'm understanding the compiler output correctly.


brentpicasso:
Great! Always good to reduce code size. Thanks!

Issue #1059, opened 2019-10-15
Choose CAT M1 for Sara R4 and enable extended error logging messages


brentpicasso:

brentpicasso:
Also choose SIM card provisioning mode

PR #1064, opened 2019-10-15
Cellular provisioning improvements

brentpicasso:
Enable SIM selction mode, force LTE CAT M1, enable enhanced logging


brentpicasso:
Resolves #1059

PR #1078, opened 2020-06-16
p/rod/avoidNullGetChannelByName fixes #1076

### LINKED ISSUE #1076: Avoid nil on getChannelByName
Right now getChannelByName() gets the current channel value from a sample in-flight. If the in-flight sample does not have that channel set, the value is returned as nil. 

https://github.com/autosportlabs/RaceCapture-Pro_firmware/blob/master/src/lua/luaLoggerBinding.c#L943

It would be more usable if the last set value of the sample can be used. Currently we have a guard against the sample being set - explore if this is necessary (IOW does the sample have the old, previously set value still?)

https://github.com/autosportlabs/RaceCapture-Pro_firmware/blob/master/src/logger/sampleRecord.c#L67

RodADavison:
I haven't done any testing yet, but for initial feedback.
The main drawback to these changes is that I could imagine some times when a sample becomes very stale.  Eg. between sessions with RCP left on.  Or with potential script or config changes. Previously, the samples would be invalidated every tick, now after one update they remain valid (but stale) indefinitely.

### DIFF
--- a/include/logger/sampleRecord.h
+++ b/include/logger/sampleRecord.h
@@ -94,7 +94,6 @@ typedef struct _LoggerMessage {
         bool needs_meta;
 } LoggerMessage;
 
-double get_channel_value( ChannelSample* sample );
 /**
  * Initializes the struct sample channel_sample buffer for use.  May be called
  * again to re-initialize the space.
--- a/include/logger/sampleRecord.h
+++ b/include/logger/sampleRecord.h
@@ -121,6 +121,7 @@ void free_sample_buffer(struct sample *s);
  * @return true if the sample was found and set
  */
 bool get_sample_value_by_name(const struct sample *s, const char * name, double *value, char ** units);
+bool get_channel_value_by_name(const char * name, double *value, char ** units);
 
 /**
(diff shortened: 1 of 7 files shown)


include/logger/sampleRecord.h
```diff
@@ -121,6 +121,7 @@ void free_sample_buffer(struct sample *s);
  * @return true if the sample was found and set
  */
 bool get_sample_value_by_name(const struct sample *s, const char * name, double *value, char ** units);
+bool get_channel_value_by_name(const char * name, double *value, char ** units);
```
brentpicasso, line 124: I would refactor everywhere throughout the firmware and switch to just using get_channel_value_by_name. The original function is no longer needed. 

This is great overall!!

RodADavison: I did try this on my first pass through, but felt like I was tugging on a thread that might unravel if I went any further.  See: logger_sample_cb_t and camera_control_sample_cb.  These callbacks require the sample parameter.  I didn't actually check to see if these were used with other functions, but I assumed that if they were needed in the first place they would take more than one value.

PR #1079, opened 2020-06-19
implement reverse telemetry / set virtual channel resolves #1077

### LINKED ISSUE #1077: Allow setting value of virtual channel via API
This would enable a number of features, such as Podium pushing 'reverse telemetry' to RaceCapture in order to set channel values such as timing and scoring data. 

Currently we have allocated 100 virtual channels, and only Lua has access to creating these channels. 

API call format:

{"setVChan":{"nm":lapsToBehind","val":9}}

If the channel does not exist, it will be auto-created with defaults. 

TODO: see if we could have a lookup table for defaults that match known channel values.

brentpicasso:
Implement feature to set / create a virtual channel via the API. This will facilitate allowing Podium to push "reverse telemetry' to RaceCapture, for purposes of providing real-time timing and scoring data, plus other information as needed.

### DIFF
--- a/include/logger/loggerApi.h
+++ b/include/logger/loggerApi.h
@@ -68,6 +68,14 @@ CPP_GUARD_BEGIN
 	API_METHOD("setWifiCfg", api_set_wifi_cfg)			\
 	API_METHOD("sysReset", api_systemReset)				\
 
+
+#if VIRTUAL_CHANNEL_SUPPORT == 1
+#define VIRTUAL_CHANNEL_METHODS                 \
+    API_METHOD("setVChan", api_set_virtual_channel_value)
+#else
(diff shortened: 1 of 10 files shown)


include/logger/loggerApi.h
```diff
@@ -248,6 +257,10 @@ int api_get_camera_control_cfg(struct Serial *serial, const jsmntok_t *json);
 int api_set_camera_control_cfg(struct Serial *serial, const jsmntok_t *json);
 #endif
 
+#if VIRTUAL_CHANNEL_SUPPORT
```
RodADavison, line 260: ifdef? or VIRTUAL_CHANNEL_SUPPORT==?

brentpicasso: Thanks for catching that. fixed

Issue #1022, opened 2019-04-18
When logging at 25hz, 10hz are drop to 5hz


kokarez:
When logging at 25hz, 10hz are drop to 5hz. This is due to tick count of 25hz not overlapping with the 10hz, thus half of the ticks will be drop by the should_sample which force all sample to be align on 25hz.

brentpicasso:
Thanks for the note. Can you attach an example config that reproduces this problem?

xavierd:
I've also just seen this. The data logged for configured 10Hz sensors is actually only logged at 5Hz, including the GPS data :(. See the attached configuration.
[conf.txt](https://github.com/autosportlabs/RaceCapture-Pro_firmware/files/3829042/conf.txt)

xavierd:
And this is a log generated with the config: [rc_2.log](https://github.com/autosportlabs/RaceCapture-Pro_firmware/files/3829046/rc_2.log). If you take a look at the OilPressure for instance, you can see that a value is generated every 0.2s (ie 5Hz), while being configured for 10Hz.

brentpicasso:
You can work-around this situation by adding a channel that logs at 50Hz or higher, or changing the Accel/Gyro channel to log at 10Hz or 50Hz.  We will work to get this resolved in an upcoming release.

PR #1080, opened 2020-06-22
added special case to catch 10Hz pulse on 25Hz sampleRate: resolves Issue #1022

RodADavison:
This catches the missing 10Hz samples.  The only downside is that it generates 10Hz output when running at 25Hz, even if there are no 10Hz sources.  However, since the GPS is usually 10Hz,  few will notice.

### DIFF
--- a/src/logger/loggerConfig.c
+++ b/src/logger/loggerConfig.c
@@ -844,7 +844,14 @@ LoggerConfig * getWorkingLoggerConfig()
 
 bool should_sample(const int sample_rate, const int max_rate)
 {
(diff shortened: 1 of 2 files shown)


brentpicasso:
Good, safe fix - thank you. However, I'm sad the main algorithm can't handle this specific case. We can look at refactoring it later. 

Thanks!

brentpicasso:
Resolves #1022

RodADavison:
The notion that using the highest update frequency works for all lower
frequencies is flawed.  Perhaps a bitstring instead with each allowable
frequency encoded ... this makes the per tick logic more complicated to
determine sampling.

PR #1082, opened 2020-07-04
Also reset wifi to factory defaults when flashing default config

### LINKED ISSUE #1081: Need to also perform factory restore on wifi module when doing main configuration reset
Sometimes, the wifi module gets into a bad state, and an AT+RESTORE can recover it. 

Solution: Issue an AT+RESTORE command to the wifi module when we perform a factory reset. 

After successfully performing this operation, the RC log should show the the Wifi driver changing the AP from the factory default ESP... to the default RaceCapture SSID.

brentpicasso:
Resolves #1081


RodADavison:
It seems odd to me that the esp8266 is not being set/reset via the wifi interface.   I would not expect to espX calls except as part of the wifi implementation.

brentpicasso:
When the unit normally starts up, we do a standard reset, not a 'factory restore'.  I considered doing a AT+RESTORE on every power up, but since that would reset flash settings on the ESP device over and over, i thought it would be excessive.  

This way, the factory restore only happens when the rest of the RC factory restore occurs.

PR #1093, opened 2020-09-13
Reset config when removing virtual channels. Resolves #1092

### LINKED ISSUE #1092: Removal of virtual channels leaves sample record out of sync
Repro Steps

* Factory default
* Add the following script

```
lrwId = addChannel("WheelLR", 50, 3, -40, 40, "mm")
rrwId = addChannel("WheelRR", 50, 3, -40, 40, "mm")
lfwId = addChannel("WheelLF", 50, 3, -40, 40, "mm")
rfwId = addChannel("WheelRF", 50, 3, -40, 40, "mm")
```

*write config
* go to dashboard
* got to setup / scripting
* remove script
* write config
* go to dashboard
* Observe repeated metadata request

```
DEBUG  ] [RCPAPI      ] Rx: {"logfile":""}
[DEBUG  ] [ApiDispatcher] processing message: logfile
[DEBUG  ] [ApiDispatcher] processing message: s
[ERROR  ] [SampleMetaException] Unexpected data packet count 33; channel meta expects between 1 and 29 channels
[INFO   ] [DataBusPump ] Sample Meta is stale, requesting meta
[DEBUG  ] [RCPAPI      ] sending meta
[DEBUG  ] [RCPAPI      ] Tx: {"getMeta":null}
```

Resolved by power cycling RaceCapture device

RodADavison:
Added reset_logger and configChanged calls when virtual channels are removed.
protected a bit more code with the vchan_mutex.

### DIFF
--- a/src/virtual_channel/virtual_channel.c
+++ b/src/virtual_channel/virtual_channel.c
@@ -58,10 +58,14 @@ int find_virtual_channel(const char * channel_name)
 
 int create_virtual_channel(const ChannelConfig chCfg)
 {
+        xSemaphoreTake(vchan_mutex, portMAX_DELAY);
         /* If the channel exists, return it and be done */
         const int id = find_virtual_channel(chCfg.label);
         if (id != INVALID_VIRTUAL_CHANNEL)
+	{
+		xSemaphoreGive(vchan_mutex);
                 return id;
+	}
 
         /*
          * Here we actually try to create a new channel.  But only if we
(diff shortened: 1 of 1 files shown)


src/virtual_channel/virtual_channel.c
```diff
@@ -105,7 +108,11 @@ size_t get_virtual_channel_count(void)
 
 void reset_virtual_channels(void)
 {
+        xSemaphoreTake(vchan_mutex, portMAX_DELAY);
         g_virtualChannelCount = 0;
+        xSemaphoreGive(vchan_mutex);
+	reset_logger_config();
```
brentpicasso, line 114: I don't think we want to reset the logger config.  The virtual channels are reset when the script re-loads, but when the script reloads we don't want to wipe the rest of the logger config. 

I think triggering the configChanged() is good, however.

RodADavison: I'll make that change and retest tomorrow.

brentpicasso:
Thanks.  To confirm, with these latest changes do the repro steps no longer cause the problem? If so, I can merge.

RodADavison:
Correct!

PR #1095, opened 2020-09-13
switch lua virtual channels to base 1.  Resolves #1094

### LINKED ISSUE #1094: in lua, user defined channels should be 1 based
this script should detect and produce an error:

a=addChannel("a", 50 )
setChannel(b,10)

instead it happily sets channel "a" to 10.  
Since we can't change lua's init to zero behaviour, we should avoid 0 as a valid channel Id.

RodADavison:
I'm optimistically submitting this, hoping you'll reconsider.  It took me much less time to write this code than it did to debug one script for one user.  It's not fair to give users tools that don't give them a chance to find their own errors.

### DIFF
--- a/src/lua/luaLoggerBinding.c
+++ b/src/lua/luaLoggerBinding.c
@@ -922,7 +922,7 @@ static int lua_add_virt_channel(lua_State *L)
                 return luaL_error(L, "Unable to create channel. "
                                   "Maximum channels reached.");
 
-        lua_pushinteger(L, chan_id+1);
+        lua_pushinteger(L, chan_id + 1);
         return 1;
 }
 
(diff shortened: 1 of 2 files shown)


src/lua/luaLoggerBinding.c
```diff
@@ -951,15 +952,18 @@ static int lua_get_virtual_channel(lua_State *ls)
                 }
         }
         /* Return nil, channel not found */
```
brentpicasso, line 954: Code comment does not match code behavior.

brentpicasso (changes requested):
Looks good, only one small change request.

brentpicasso:
Looks good overall. When committing code, please run it through bin/style.h first.  I'll do that now on the 2.18.2 code.

brentpicasso:
Ok, r/2.18.2 has been updated with bin/style.sh. 

You can merge r/2.18.2 and also apply bin/style.sh to your branch.  Thanks!

brentpicasso (approved):

PR #1101, opened 2020-10-22
Add hex Format string to lua string.format

eacmen:
This allows developers to use the hex format string %x or %X in lua scripts to print hexadecimal PIDs versus float or integer values. 

`make test` passed.


lib/lua/src/lstrlib.c
```diff
@@ -854,6 +854,12 @@ static int str_format (lua_State *L)
                 modp_itoa10((LUA_INTFRM_T)luaL_checknumber(L, arg),buff);
                 break;
             }
+            case 'x':
+            case 'X': {
+                addintlen(form);
+                modp_itoaX((LUA_INTFRM_T)luaL_checknumber(L, arg),buff,16);
```
RodADavison, line 860: should this be: modp_itoa16((LUA_INTFRM_T)luaL_checknumber(L, arg),buff); ?

RodADavison, line 860: It doesn't look like the new function uitoa16 is used/needed?!

eacmen: You are correct. I made this change a long time ago and forgot that the lua code already had the function I needed. Thanks for finding that!

eacmen:
closing this PR and basing it off of `r/2.18.2` insteads.

PR #1109, opened 2021-12-26
fix edge case where firmware may crash if GPS data does not change

brentpicasso:
We automatically interpolate GPS data if the sample rate is < 10Hz. This was originally done due to PodiumConnect receiving 1Hz GPS data from some AIM systems, and sometimes the start/finish point would be missed. 

There's an edge case if the latitude data does not change from point-to-point where a zero division error can occur. 

This fix just assigns the next longitude value during the interpolation if the latitude does not change.

### DIFF
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -851,7 +851,7 @@ void lapstats_processUpdate(GpsSnapshot *gps_snapshot)
 
         for (size_t i = 0; i < interval_count; i++) {
                 /* Linearly interpolate longitude from latitude */
-                float interp_lon = lon1 + (interp_lat - lat1) * ((lon2 - lon1) / (lat2 - lat1));
+                float interp_lon = lat2 == lat1 ? lon2 : lon1 + (interp_lat - lat1) * ((lon2 - lon1) / (lat2 - lat1));
 
                 /* Update current GPS snapshot with interpolated values */
                 gps_snapshot->sample.point.latitude = interp_lat;
--- a/src/lap_stats/lap_stats.c
+++ b/src/lap_stats/lap_stats.c
@@ -793,7 +793,7 @@ void lapstats_processUpdate(GpsSnapshot *gps_snapshot)
 
         /**
          * ============================================
-         * Set up interpolation for GPS Point
+         * Set up interpolation for GPS Points
          * ============================================
          */
         float lat1 = gps_snapshot->previousPoint.latitude;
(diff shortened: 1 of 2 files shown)


ddimensia (approved):
I guess I don't understand why we're linearly interpolating over latitude and not time but this will definitely fix the div/zero error.

brentpicasso:
@ddimensia thanks for reviewing. Actually, the code linearly interpolates 3 things: time, speed, and lat/lon

RodADavison:
It looks there is still a spot for div by zero in the line above as well.

brentpicasso:
@RodADavison  which line above? If you mean the red line, then that's the line that was removed in the diff.

RodADavison:
Yes ... I appear to have misinterpreted the diff display.

RodADavison:
I like the new change.
2 other suggestions:
Line 788 - becomes if (interval_count >= 0 ) // avoids possible div by zero further down.
Lines 808 to 821 become:
float lat_interval = (lat2-lat1)/(float) interval_count;
float lon_interval = (lon2-lon1)/(float) interval_count;
float interp_lat = lat1;
float interp_lon = lon1;

RodADavison:
Likely unnecessary to mention: Code breaks at the North/South Pole Loops or a track that lies across the -180/180 line.

brentpicasso:
@RodADavison  thanks. Line 786 guards against divide by zero.  I simplified the interval algorithm as you suggested.

RodADavison (approved):
Looks good!

src/lap_stats/lap_stats.c
```diff
@@ -793,40 +793,33 @@ void lapstats_processUpdate(GpsSnapshot *gps_snapshot)
...
-        /**
-         * Evenly split up difference in latitiude into intervals
-         * longitude is linearly interpolated based on changing latitude
-         */
-        float lat_interval = fabsf(lat2 - lat1) / (float)interval_count;
-        if (lat1 > lat2)
-                /* Account for reverse direction */
-                lat_interval = -lat_interval;
-
+        float lat_interval = (lat2 - lat1) / (float)interval_count;
```
ddimensia, line 803: Should we check to make sure `delta_since_last` or `interval_count` isn't zero to prevent div zero errors?

ddimensia (changes requested):

src/lap_stats/lap_stats.c
brentpicasso: The MAX in line 786 guards against 0 interval_count.  

Will look at delta_since_last

ddimensia (approved):

PR #1115, opened 2022-06-22
Fix OFFSET minimum calculation when in bit mode

eacmen:
When in bit mode you should be able to select bits beyond 56 (7*8). This change fixes the calculation of what the max should be in bit mode which is 63.


brentpicasso (approved):
Thanks - merging into little_endian_fix

PR #1116, opened 2022-06-28
Little endian fix

brentpicasso:


serdartoren (commented):
I understand that you have made changes for non- 8, 16 and 32 lengths.
Everything looks right for me :)
Now i will test it in my local repo

PR #1118, opened 2022-08-08
P/brent/logging status

brentpicasso:
This allows the current firmware to broadcast logging start/stop messages when the Automatic Control mechanism is triggered. 

The point is to allow the app to start/stop logging synchronized with the Automatic Control feature - which can also control cameras.  This will allow us to make video + data files that are more precisely synchronized.

https://user-images.githubusercontent.com/3267533/183531817-0f7720e5-1337-48ab-94f5-aa658fdcd994.mp4

### DIFF
--- a/src/logger/api_event.c
+++ b/src/logger/api_event.c
@@ -91,6 +91,8 @@ bool api_event_destroy_callback(const int handle)
 
 void process_api_event(const struct api_event * api_event, struct Serial * serial)
 {
+        /* Dispatch API messages to connected devices */
+
         /* Only try to send if our Serial device is connected */
         if (!serial_is_connected(serial))
                 return;
@@ -117,6 +119,11 @@ void process_api_event(const struct api_event * api_event, struct Serial * seria
                 api_send_button_state(serial, &api_event->data.butt_state);
                 put_crlf(serial);
                 break;
+        case ApiEventType_LoggerStatus:
+                api_send_logging_status(serial);
+                put_crlf(serial);
+                break;
+
         default:
                 pr_warning_int_msg(LOG_PFX "Unknown ApiEvent type ", api_event->type);
                 break;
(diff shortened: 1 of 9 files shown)


src/logger/api_event.c
```diff
@@ -117,6 +119,11 @@ void process_api_event(const struct api_event * api_event, struct Serial * seria
                 api_send_button_state(serial, &api_event->data.butt_state);
                 put_crlf(serial);
                 break;
+        case ApiEventType_LoggerStatus:
```
brentpicasso, line 122: Every time we want to define a new event that gets broadcasted to connected devices (including telemetry to podium) we define a new event here

src/logger/loggerApi.c
```diff
@@ -275,18 +275,26 @@ static void get_bt_status(struct Serial* serial, const bool more)
 #if BLUETOOTH_SUPPORT
         json_objStartString(serial, "bt");
         json_int(serial, "init", (int)bt_get_status(), 0);
-        json_objEnd(serial, 1);
+        json_objEnd(serial, more);
```
brentpicasso, line 278: unrelated bug fix.

src/logger/loggerApi.c
```diff
@@ -275,18 +275,26 @@ static void get_bt_status(struct Serial* serial, const bool more)
...
 
 static void get_logging_status(struct Serial* serial, const bool more)
 {
-#if SDCARD_SUPPORT
         json_objStartString(serial, "logging");
         json_int(serial, "status", (int)logging_get_status(), 1);
         json_int(serial, "dur", logging_active_time(), 0);
-        json_objEnd(serial, 1);
-#endif
+        json_objEnd(serial, more);
```
brentpicasso, line 287: 1 -> more - unrelated bugfix.

src/logger/loggerApi.c
```diff
@@ -275,18 +275,26 @@ static void get_bt_status(struct Serial* serial, const bool more)
...
+}
+
+int api_send_logging_status(struct Serial* serial)
```
brentpicasso, line 290: This sends just the logging status in JSON form to the connected clients.

We have another API method that sends multiple statuses

serdartoren (commented):
Looks like a nice feature. 
As i understand similar job was done for SD card only before.

RodADavison (approved):
Looks good to me.

